PR #224, opened 2015-01-03
fix setequal bug

kevinushey:


eddelbuettel:
Looks good to me -- thanks to you and @thirdwing for the quick fix!

PR #225, opened 2015-01-11
Debug: memory leak of throwed object

wush978:
Dear all,

This pull request is a prototype of solving [my previous question about memory leak of non caught object](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2014-December/008319.html).

It contains a set of unit tests (`inst/unitTests/runit.exception.R`) which tests whether the destructor of throwed object is called or not.

I solve it via evaluating the stop expression after closing the scope of `try`/`catch`. The exception is converted to a stop object in R in `catch`. The stop object will be raised after exiting the scope of `catch`, which will make the compiler correctly call the destructor of the throwed object.

It passes the unit tests on Ubuntu (gcc-4.9.1), windows (Rtools with gcc 4.6.3) and OS X (clang-6.0).

Note: 
Sorry that my editor remove the 4 spaces in `inst/include/Rcpp/exceptions.h`. Please let me know if it violates your coding style or not.


jjallaire:
Some comments on the implementation:

1. I don't think you need the whole Rcpp::set_stop / is_stop business as you can handle the whole thing with local variables (just define them above try in the first macro).

2. The same issue applies to the catch(InterruptedException) and catch(..) codepaths so should be handled there as well (you probably need a distinct variable for interrupted but could use the same stack variable for the two exception cases)

wush978:
Thanks for your comment. 

In my opinion, this bug ( or issue ) is caused by the difference between C and C++ compiler. The R interpreter is not compiled by c++ compiler, so it does not call the destructor after leaving the scope. So we should not use any local variable because raising R error leaves the scope without calling the destructor of the local variable. Instead, a global smart pointer should be used and the destructor will be called before raising the next R error object or the unloading of the Rcpp.

Here is an minimal example to show this concept: <https://gist.github.com/wush978/0713e6aa5cc70f1c75b9>. The `testLocalVariable` use the local variable `expr` and `condition` which are never deleted and cause memory leak. Instead, the `testGlobalVariable` use the global variable and the destructor is correctly called even at the `dyn.unload`.

Therefore, I cannot use a local variable as you suggested.

And thanks for your reminding of the other codepaths. I'll update the code to handle them if there is no futher issues.

eddelbuettel:
Wush,   I had a few things to raise so I sent you email.

wush978:
Thanks for your suggestion and I'll close this PR first.

Issue #227, opened 2015-01-14
naming function arguments either "major" or "minor" prevents compilation


jackwasey:
```
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
int timesTwo(int major) {
   return major * 2;
}
```
I'm far from an expert, but I think the following generated code appears to the compiler as a function call to `major`.
_RcppExports.cpp_
```
Rcpp::traits::input_parameter< int >::type major(majorSEXP )
```
I don't see anything anywhere about `major` or `minor` being reserved keywords in C++, the STL or Rcpp.

eddelbuettel:
Confirmed:
```{.r}
cppFunction("int foo(int maj) { return(2*maj); }")
```
works where 
```{.r}
cppFunction("int foo(int major) { return(2*major); }")
```
fails -- and I also fail to see where the identifier comes from.

romainfrancois:
Works fine for me on OS X with clang: 

```
> require(Rcpp)
Le chargement a nécessité le package : Rcpp
> cppFunction("int foo(int major) { return(2*major); }")
> foo(2L)
[1] 4
```

romainfrancois:
I get this on ubuntu: 

```
> Rcpp:::areMacrosDefined( c( "major", "minor" ) )
major minor 
 TRUE  TRUE 
```

eddelbuettel:
It fails for me with clang++ and g++:

```{.r}
R> cppFunction("int foo(int major) { return(2*major); }")
file22fc2782fa4f.cpp: In function ‘SEXPREC* sourceCpp_42708_foo(SEXP)’:                                                                                                                                                                    
file22fc2782fa4f.cpp:17:28: error: ‘major’ was not declared in this scope
         int __result = foo(major);
                            ^
make: *** [file22fc2782fa4f.o] Error 1
ccache g++ -I/usr/share/R/include -DNDEBUG    -I"/usr/local/lib/R/site-library/Rcpp/include"    -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -g  -O3 -Wall -pipe -Wno-unused -pedantic  -c \
file22fc2782fa4f.cpp -o file22fc2782fa4f.o
/usr/lib/R/etc/Makeconf:142: recipe for target 'file22fc2782fa4f.o' failed
Error in sourceCpp(code = code, env = env, rebuild = rebuild, showOutput = showOutput,  :
[...]

PR #228, opened 2015-01-15
work around macro leakage

kevinushey:
This PR works around the (really ridiculous) macro leaking of `<sys/sysmacros.h>` from GCC. See https://github.com/RcppCore/Rcpp/issues/227 for motivation, and see https://bugzilla.redhat.com/show_bug.cgi?id=130601 for how it's probably not going to be fixed on GCC's side.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-01-15  Kevin Ushey  <kevinushey@gmail.com>
+
+        * inst/include/RcppCommon.h: remove leaked macros
+        'major', 'minor' and 'makedev' from sys/sysmacros.h
+
 2015-01-08  Dirk Eddelbuettel  <edd@debian.org>
 
         * inst/examples/OpenMP/GNUmakefile: Renamed from Makefile because it
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,14 +1,16 @@
 2015-01-15  Kevin Ushey  <kevinushey@gmail.com>
 
-        * inst/include/RcppCommon.h: remove leaked macros
-        'major', 'minor' and 'makedev' from sys/sysmacros.h
+        * inst/include/Rcpp/platform/sysmacros.h: remove leaked macros
+        'major', 'minor', 'makedev' from <sys/sysmacros.h>
+        * inst/include/RcppCommon.h: ditto
+        * inst/unitTests/cpp/Vector.cpp: simple test
 
 2015-01-08  Dirk Eddelbuettel  <edd@debian.org>
 
         * inst/examples/OpenMP/GNUmakefile: Renamed from Makefile because it
         contains GNU make extentions (which we happen to like)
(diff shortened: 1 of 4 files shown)


eddelbuettel:
I am a little fuzzy as well and torn between _"wtf do we care? users can define major"_ and _"just silently `undef` and be done with_.  Did we need to bark?

jjallaire:
Yeah, for that reason I don't like the idea of printing a warning. I also
agree that if this is an R-core problem we shouldn't do anything to paper
over it.

On Thu, Jan 15, 2015 at 7:23 AM, Dirk Eddelbuettel <notifications@github.com

kevinushey:
Thanks, you're right -- the `#warning` is overly severe, especially since we have `#pragma`s that we can use to pull any pre-defined macros over.

@jjallaire, R-Core argued that this isn't their problem since any C/C++ program that includes `<sys/types.h>` will get the `<sys/sysmacros.h>` header, which will 'leak' those macros. I would beg to differ, but ... we can still be good citizens and avoid this headache in a safe way, I think.

It looks like this feature has been available in `gcc` since 4.4 -- although it existed in an undocumented form. http://gcc.1065356.n5.nabble.com/Bug-target-35054-New-No-documentation-of-pragma-push-macro-quot-macro-name-quot-td331629.html.

And, yes, I'm being overly cautious about the `#undef` but I want to avoid this if possible:

```
#define major "My own macro!"
#include <Rcpp.h> // sorry about your macro!
```

jjallaire:
>
> And, yes, I'm being overly cautious about the #undef but I want to avoid
> this if possible:
>
> #define major "My own macro!"
> #include <Rcpp.h> // sorry about your macro!
>
>
But every library that defines macros (or un-namespaced symbols of any
kind) has this issue. I don't think it's incumbent on us to proactively
resolve any of these conflicts.

kevinushey:
Fair enough, but typically maintainers that do define macros go to some pain to ensure that conflicts are unlikely (upper-case naming; prefix with library name; so on). And I would expect that _system_ headers should go to some length to avoid macro conflicts (user headers / libraries are more 'free' to do as they will)

`gcc` actually _does_ this for the macros, but provides `major` and `minor` as shortcuts:

```
/* Access the functions with their traditional names.  */
#define major(dev) gnu_dev_major (dev)
#define minor(dev) gnu_dev_minor (dev)
#define makedev(maj, min) gnu_dev_makedev (maj, min)
```

So, the main reason why I think it's worth working around:

1. It affects everyone using `gcc`, or a Linux platform where `clang` is using `gcc`'s headers,
2. It always affects `Rcpp` due to `R.h` including `<sys/types.h>` (and hence `<sys/sysmacros.h>`)

Ultimately ... it's not a big issue, I just find it somehow really frustrating that a system header could do something like this.

kevinushey:
Thanks!

PR #244, opened 2015-01-31
Use R_ToplevelExec when evaluating R code 

jjallaire:
By using R_ToplevelExec we catch user interrupts. Without R_ToplevelExec interrupts cause a longjmp which bypasses all C++ destructors on the stack. 

For example, without this change the destructor of MyClass is NOT called if the user issues an interrupt during the execution of the function:

```cpp
#include <Rcpp.h>
using namespace Rcpp;

class MyClass {
public:
    ~MyClass() {
        Rcout << "Instance Destroyed" << std::endl;
    }
};

// [[Rcpp::export]]
void callFunc(Function f, int val) {
   MyClass object;
   f(val);
}

/*** R
callFunc(Sys.sleep, 5)
*/
```

With the change the destructor is properly called even in the face of interrupt.

### DIFF
--- a/inst/include/Rcpp/api/meat/Rcpp_eval.h
+++ b/inst/include/Rcpp/api/meat/Rcpp_eval.h
@@ -70,7 +70,9 @@ namespace Rcpp{
         Shield<SEXP> res(::Rf_eval( call, RCPP ) );
 
         // collect warnings
-        Shield<SEXP> warnings(::Rf_eval(Rf_lang1(collectWarningsSym), RCPP));
+        Shield<SEXP> warningCall(Rf_lang1(collectWarningsSym));
+        Shield<SEXP> warnings(::Rf_eval(warningCall, RCPP));
+        
         evalCall->warnings = Rcpp::as<std::vector<std::string> >(warnings);
 
         // check for error
--- a/inst/include/Rcpp/api/meat/Rcpp_eval.h
+++ b/inst/include/Rcpp/api/meat/Rcpp_eval.h
@@ -23,38 +23,57 @@
 namespace Rcpp{
 
     struct EvalCall {
-        SEXP expr_;
+        SEXP expr;
         SEXP env;
         SEXP result;
+        std::vector<std::string> warnings;
         std::string error_message;
     };
 
     inline void Rcpp_eval(void* data) {
 
         EvalCall* evalCall = (EvalCall*)data;
-        SEXP expr_ = evalCall->expr_;
         SEXP env = evalCall->env;
 
-        Shield<SEXP> expr( expr_) ;
+        Shield<SEXP> expr(evalCall->expr) ;
 
         reset_current_error() ;
 
         Environment RCPP = Environment::Rcpp_namespace();
+        SEXP withCallingHandlersSym    = ::Rf_install("withCallingHandlers");
         SEXP tryCatchSym               = ::Rf_install("tryCatch");
         SEXP evalqSym                  = ::Rf_install("evalq");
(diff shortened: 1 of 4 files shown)


kevinushey:
You can use:

    (add-hook 'before-save-hook 'delete-trailing-whitespace)

but I actually have it turned off in my `.emacs` since it's disruptive in projects that do have lots of trailing whitespace. Maybe we can put it in the `.dir-locals.el` file?

kevinushey:
@jjallaire, I get a funky error when running a 'stress test' example with this branch:

```R
#include <Rcpp.h>
using namespace Rcpp;

class MyClass {
public:
  ~MyClass() {
    Rcout << "Instance Destroyed" << std::endl;
  }
};

// [[Rcpp::export]]
void callFunc(Function f, int val) {
  MyClass object;
  f(val);
}

/*** R
f <- function(g) {
  return (function(x) {
    tryCatch(withCallingHandlers(
      g(x))
    )
  })
}

m <- f(Sys.sleep)
gctorture(TRUE)
callFunc(Sys.sleep, 5)
gctorture(FALSE)
*/
```

I got this printed on the console:

```R
> f <- function(g) {
+   return (function(x) {
+     tryCatch(withCallingHandlers(
+       g(x))
+     )
+   })
+ }

> m <- f(Sys.sleep)

> gctorture(TRUE)

> callFunc(Sys.sleep, 5)
Error in list(c(128194L, 133L), "/Users/kevin/Library/R/3.1/library/Rcpp/R/Rcpp.rdb", TRUE, function (n)  : 
  unused arguments ("/Users/kevin/Library/R/3.1/library/Rcpp/R/Rcpp.rdb", TRUE, function (n) 
{
    if (existsInFrame(n, envenv)) getFromFrame(n, envenv) else {
        e <- mkenv()
        set(n, e, envenv)
        key <- getFromFrame(n, env)
        data <- lazyLoadDBfetch(key, datafile, compressed, envhook)
        if (is.null(data$enclos)) parent.env(e) <- emptyenv() else parent.env(e) <- data$enclos
        vars <- names(data$bindings)
        for (i in seq_along(vars)) set(vars[i], data$bindings[[i]], e)
        if (!is.null(data$attributes)) attributes(e) <- data$attributes
        if (!is.null(data$isS4) && data$isS4) .Internal(setS4Object(e, TRUE, TRUE))
        if (!is.null(data$locked) && data$locked) .Internal(lockEnvironment(e, FALSE))
        e
    }
})
Instance Destroyed
```

I think there might be a `PROTECT` missing somewhere -- let me see.

inst/include/Rcpp/api/meat/Rcpp_eval.h
```diff
@@ -18,38 +18,95 @@
...
+            tryCatchCall,
+            warningRecorderSym
+        ) ) ;
+        SET_TAG( CDDR(call), warningSym ) ;
+
+        // execute the call
         Shield<SEXP> res(::Rf_eval( call, RCPP ) );
 
+        // collect warnings
+        Shield<SEXP> warnings(::Rf_eval(Rf_lang1(collectWarningsSym), RCPP));
```
kevinushey: Doing an `eval` on an `Rf_lang` symbol is dangerous -- the language object used for the evaluation can actually be cleaned up and then hence be unavailable for the call, which breaks any functions that rely on `match.call()` or other introspection on the active call.

It's best to create the language object, protect it, evaluate it, then unprotect it after it's not needed.

kevinushey:
With the protection change, the stress test runs successfully for me.

jjallaire:
@romainfrancois let us know if you see any issues still lurking here (@kevinushey found and fixed one already via gctorture).

PR #245, opened 2015-02-02
Add Rcpp::warning function as wrapper for Rf_warning

jjallaire:
This compliments the existing Rcpp::stop function and will encourage the use of the C-level warning API rather than calling base::warning via R. Also added support for tinyfmt which works the same way as it does for Rcpp::stop.

@eddelbuettel It's possible that given how simple a name "warning" is that there is a chance of symbol clashes leading to compilation errors in some packages. C++ namespaces should take care of this but it's not beyond the pale of imagination that there's a subtle corner case we run afowl of. Probably worth a revdep check to be more certain, we can back this out if it presents intractable problems.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-02-02  JJ Allaire  <jj@rstudio.org>
+
+        * inst/include/Rcpp/exceptions.h: Add Rcpp::warning function as
+        wrapper for Rf_warning
+
 2015-01-25  Kevin Ushey  <kevinushey@gmail.com>
 
         * inst/include/Rcpp/utils/tinyformat.h: define an error handler for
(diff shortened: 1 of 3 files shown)


eddelbuettel:
Yes, thanks, I'll definitely run another revdep check -- given how I only caught 3/4 of your last commit. 

We also got two 'problematic' packages on CRAN updated with the patches we sent them.

eddelbuettel:
(Ooops. Off by one in the commit message above. Meant #244, not #245)

jjallaire:
@eddelbuettel, did we run a revdep check covering this branch? (feature/warning-function). I see the reference to the overnight check but I have a feeling it was for a different branch given the close/reopen sequence. If we have run one then I'll merge this (but I'll wait if we haven't as I'm guessing there is some non-trivial chance that introducing a function named "warning" will cause a problem somewhere).

PR #247, opened 2015-02-03
Include pkg_types.h file in exports if it's present in inst/include or src

jjallaire:
This provides a mechanism for users to inject types into RcppExports.cpp (e.g. for as/wrap, typedefs, etc.). The mechanism works by convention: if file(s) named pkgname_types.h[pp] are in the src or inst/include directory they are automatically included in RcppExports.cpp


eddelbuettel:
Cooking with gas :)

PR #248, opened 2015-02-03
Simplify generated attributes code for RNGScope

jjallaire:
Use RObject and it's destructor rather than SEXP protect/unprotect. The code we were generating was correct however hard to read and reason about for users not familiar with R/Rcpp internals. Previously we held the result in a raw SEXP which we unprotected after leaving a special block created for the RNGScope:

```cpp
RcppExport SEXP rcpp_hello_world() {
BEGIN_RCPP
    SEXP __sexp_result;
    {
        Rcpp::RNGScope __rngScope;
        List __result = rcpp_hello_world();
        PROTECT(__sexp_result = Rcpp::wrap(__result));
    }
    UNPROTECT(1);
    return __sexp_result;
END_RCPP
}
```

Now, we simply construct an RObject *prior to* the RNGScope on the stack and then return that object (this ensures that any GC triggered by PutRNGState won't collect the result SEXP before it can be returned):

```cpp
RcppExport SEXP rcpp_hello_world() {
BEGIN_RCPP
    Rcpp::RObject __result;
    Rcpp::RNGScope __rngScope;
    __result = Rcpp::wrap(rcpp_hello_world());
    return __result;
END_RCPP
}
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,12 @@
 2015-02-03  JJ Allaire  <jj@rstudio.org>
 
+        * src/attributes.cpp: Simplify generated attributes code for
+        RNGScope (use RObject and it's destructor rather than SEXP
+        protect/unprotect).
+        * vignettes/Rcpp-package.Rnw: Update docs on generated code.
+
+2015-02-03  JJ Allaire  <jj@rstudio.org>
+
         * inst/include/Rcpp/exceptions.h: Add Rcpp::warning function as
         wrapper for Rf_warning
 
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -20,6 +20,9 @@
     \itemize{
       \item Include pkg_types.h file in RcppExports.cpp if it's present in
       inst/include or src.
+      \item Simplify generated attributes code for \code{RNGScope} (use
+      \code{RObject} and it's destructor rather than \code{SEXP}
+      protect/unprotect).
     }
   }
 }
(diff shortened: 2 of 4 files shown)


jjallaire:
FYI rebuilt the entire Rcpp Gallery with this change and all is well (probably also worth a revdep check). I think this is a straightforward change however I could use a sanity check to make sure I'm not missing something (as the code being modified was put in there originally to avoid a crash under extreme GC conditions and I'd certainly hate to regress back to that behavior). @eddelbuettel @kevinushey @romainfrancois

kevinushey:
FWIW, we can run unit tests with `gctorture(TRUE)` on with:

    Rcpp:::test(gctorture = TRUE)

so if you feel like heating up your laptop a bit, try `R --vanilla --slave -e "Rcpp:::test(gctorture = TRUE)"`. (I'm running that now and will report if anything is amiss)

kevinushey:
@eddelbuettel, running a plain `Rcpp:::test()` fails now because we no longer `require(RUnit)`; instead we just `requireNamespace()` it and don't qualify our calls to e.g. `checkEquals`. Should we revert back to just loading `RUnit` there?

eddelbuettel:
@kevinushey : that would make a Suggests: a Depends.  Can't we just switch to requireNamespace("RUnit") and do other cleanups as needed?

jjallaire:
Ran all of the Rcpp tests with gctorture = TRUE and all looks good.

Issue #250, opened 2015-02-04
Rcpp sugar exact value by name seems so slowly


guokai8:
Hi, I just test how to exact value  from a NumericVector by names, the result is so terrible.Below is the code, and the test result;
```
######name_index2.cpp#####
#include <Rcpp.h>
using namespace Rcpp;
//[[Rcpp::export]]
NumericVector sugar_index(StringVector& xin,NumericVector& yin){
  NumericVector res;
  res=yin[xin];
  return res;
}

>sourcecpp("name_index2.cpp")
>x<-rnorm(500)
>names(x)<-paste("a",1:500,sep="_");
>y<-sample(names(x),200)
>library(rbenchmark)
>benchmark(sugar_index(y,x),x[y])
                test         replications elapsed relative user.self sys.self user.child
1 sugar_index(y, x)          100   0.047     23.5     0.047    0.001          0
2              x[y]             100   0.002      1.0     0.003    0.000           0
```
So why the Rcpp sugar seems so slowly!
Is there other method could be used in the Rcpp instead ?

guokai8:
why the first line "#include <Rcpp.h> " could not show completely

romainfrancois:
Embed you c++ code between  "```cpp" and "```"

guokai8:
I have tried , but it failed. The point is the "sugar_index" show almost 24 slowly than the base function. Did I miss something or make mistake?

romainfrancois:
See this for github flavored markdown; https://help.github.com/articles/github-flavored-markdown/

As for the issue you report, it all happens from this line: https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Subsetter.h#L140

in this methods: 

```
    int find(const RHS_t& names, const char* str) {
        for (int i=0; i < lhs_n; ++i) {
            if (strcmp( CHAR( STRING_ELT( names, i) ), str) == 0) return i;
        }
        stop("no name found");
        return -1;
    }
```

The `find` uses string comparisons where direct pointer comparisons of the `CHARSXP` would be much faster. cc @kevinushey

guokai8:
Thanks,I will try it.Could I use another method instead ?

eddelbuettel:
There are a few things to try.  First, your measured times are very small and prone to error and overhead.  Repeat maybe 1000 or 5000 times (see `help(benchmark)`).  Second, consider writing alternate little C++ function to try different comparisons.  

[...]

PR #252, opened 2015-02-06
faster strsxp subsetting (cc @romainfrancois, #250)

kevinushey:
This compares string elements by pointers rather than contents, providing improved sugar subsetting.

@romainfrancois, if you're willing can you confirm I haven't bundled anything here? (Testing on the examples of #250 shows it's much faster, but still not on par with base R's `[`)


eddelbuettel:
I'll merge it and give it and the other recent changes a whirl over the revdeps.

PR #253, opened 2015-02-12
Add rng parameter to Rcpp::export to prevent inclusion of RNGScope in generated code

jjallaire:


eddelbuettel:
I like it !  What was the motivation today?  Leakage?  Saving a cycle here or there?  Drew's post?

One suggestion: a _positive_ term ("rng") for a _negative_ action ("do _not_ include") is weird.  Would it make sense to call it `norng` instead of `rng`?

kevinushey:
It seems like we default to `rng = TRUE` but someone could write:

    // [[Rcpp::export(rng = FALSE)]]

if so desired. Is that correct @jjallaire?

jjallaire:
I knew this had been asked for a few times and it's hung over my head that we didn't support it. I figured as long as I am mucking around in the attributes code I might as well just do it :-)

Kevin is correct, the default is rng = true so the only meaningful value is rng = false (which doesn't have the inversion of meaning problem you described)

jjallaire:
Do you think I should also support TRUE and FALSE (I currently only support true and false lowercase)

kevinushey:
I would say yes (I almost reflexively wrote `FALSE` just because I was thinking in `R` mode; also R provides `TRUE` and `FALSE` defines anyway so users might be accustomed to using that already.)

jjallaire:
Okay, I'll update to support TRUE and FALSE as well.

PR #254, opened 2015-02-12
Allow use of attributes without Rcpp:: prefix when preceded by 'using namespace Rcpp'

jjallaire:
Note that this PR is branched off the previous rng = false one so should only be taken after that one is.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,7 +2,11 @@
 
         * DESCRIPTION: bump version
         * src/attributes.cpp: Add rng parameter to Rcpp::export to 
-        prevent inclusion of RNGScope in generated code.
+        prevent inclusion of RNGScope in generated code; Allow use
+        of attributes without Rcpp:: prefix when preceded by 
+        'using namespace Rcpp'
+        * inst/unitTests/cpp/na.cpp Use un-namespaced version of
+        export attribute for the NA unit tests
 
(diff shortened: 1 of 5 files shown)


jjallaire:
I'm actually not sure about this PR. In reviewing the C++11 specification on attributes it's my understanding that attribute namespaces have nothing to do with C++ namespaces and are just a convention to prevent various attribute schemes from stomping on each other. 

This PR actually attempts to detect "using namespace Rcpp" and use that to determine if non-namespaced attributes are supported. This seems incorrect from the standpoint of emulating "real" attributes.

Since the comment-based attributes are our own scheme we could certainly feel free to not require the Rcpp:: prefix, but if we do I think it should be unconditional and have nothing to do with "using namespace Rcpp". 

What does everyone think about not requiring the "Rcpp::" prefix? I could see it both ways and will go with whatever the consensus is.

eddelbuettel:
I am not a huge fan.  The cost of prefixing via `Rcpp::` is very small, and I find it has a clear benefit of adding clarity.

jjallaire:
Okay, I think I agree. I'll close this for now.

eddelbuettel:
Thank you!  I wasn't going to fight it -- majority vote has its merits.  And we can revisit.

PR #255, opened 2015-02-14
enable includes of local files in sourceCpp

jjallaire:
This PR makes it possible to do local includes (e.g. `#include "foo.hpp"`) within a sourceCpp translation unit. This was formerly not enabled because sourceCpp caches the results of builds and only performed a rebuild when the main C++ file was modified. This meant that changes in files referenced via a local include wouldn't force a rebuild and therefore would give surprising (and incorrect) results.

We now scan recursively for all local includes within the translation unit and make their file modification times part of the cache key (so changes in these files will always result in a rebuild).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-02-14  JJ Allaire  <jj@rstudio.org>
+
+        * src/attributes.cpp: Allow includes of local files
+        (e.g. #include "foo.hpp") in sourceCpp
+
 2015-02-13  JJ Allaire  <jj@rstudio.org>
 
         * src/attributes.cpp: Allow 'R' to come immediately after '***'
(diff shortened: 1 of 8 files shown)


eddelbuettel:
You're on a roll. This is super useful too.

As I'm traveling I can't give it level of detail it warrants.  Ok if we let it sit for a few days?

jjallaire:
No hurry whatsoever, we can review and merge whenever we have the time to
do it properly.

On Feb 14, 2015, at 8:18 PM, Dirk Eddelbuettel <notifications@github.com>
wrote:

You're on a roll. This is super useful too.

As I'm traveling I can't give it level of detail it warrants. Ok if we let
it sit for a few days?

—
<https://github.com/RcppCore/Rcpp/pull/255#issuecomment-74402180>.

PR #256, opened 2015-02-14
Enable preview of vignettes in RStudio

jjallaire:
A few changes to make the 'Compile PDF' button work for Rcpp vignettes in RStudio:

- Set Sweave as the default Rnw processor in RStudio project file
- Magic comment to specify highlight package as the Sweave driver (note this is only recognized by development versions of RStudio >= v0.99.259)
- Turn off Sweave concordance (enabling it produced errors, not sure why)
- Add artifacts of PDF preview to .gitignore for vignettes directory

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2015-02-14  JJ Allaire  <jj@rstudio.org>
+
+        * Rcpp.Rproj: Specify Sweave as Rnw handler for RStudio
+        * vignettes/*.Rnw: Add driver magic comment and turn off 
+        Sweave concordance.
+        * vignettes/.gitignore: Ignore artifacts of PDF preview
+
 2015-02-13  Dirk Eddelbuettel  <edd@debian.org>
 
         * .travis.yml (install): Switch to using ppa:edd/misc to install all the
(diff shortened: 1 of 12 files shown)


eddelbuettel:
Very nice!  That was a minor wart, and I have contemplated no longer using highlight---but as it is so entrenched in our vignettes...

Cheers from Vegas and 70+ degrees. A nice break from the Chicago weather, and I guess I should not mention weather to you anyway ... ;-)

jjallaire:
I believe I can get rid of the need to include concordance = FALSE with one
more tweak to RStudio. Let me see if I can sort that out before we merge
the PR.

Don't get me started on the weather, we are in the midst of yet another
blizzard right now :-\


On Feb 14, 2015, at 8:16 PM, Dirk Eddelbuettel <notifications@github.com>
wrote:

Very nice! That was a minor wart, and I have contemplated no longer using
highlight---but as it is so entrenched in our vignettes...

Cheers from Vegas and 70+ degrees. A nice break from the Chicago weather,
and I guess I should not mention weather to you anyway ... ;-)

—
<https://github.com/RcppCore/Rcpp/pull/256#issuecomment-74402154>.

jjallaire:
Sure enough I was able to eliminate the need for the concordance directive. Updated the PR to reflect this. Note that this will all work correctly as of the next daily build of RStudio (v0.99.259).

PR #258, opened 2015-02-17
update attributes vignette with docs on new features

jjallaire:
Add documentation on features added since the original attributes vignette:

- Code sharing via local includes (note this isn't yet merged to master, see https://github.com/RcppCore/Rcpp/pull/255)
- Injecting types into RcppExports.cpp
- rng = false option

Add documentation on some features that are not attributes per-se but which are worth covering in an introduction to writing well behaved and integrated C++ function:

- Using Imports and NAMESPACE
- Rcpp::warning function
- Rcpp::checkUserInterrupt function

### DIFF
--- a/vignettes/.gitignore
+++ b/vignettes/.gitignore
@@ -1,4 +1,5 @@
 Rcpp-*.bbl
+Rcpp-*.blg
 Rcpp-*.log
 Rcpp-*.tex
 Rcpp-*.pdf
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Lovely (and embarrassingly overdue in some parts) update to the vignette.

I'll look more closely at the other one this eve but expect no issues.

jjallaire:
Embarrassingly overdue indeed! Glad we finally have all of that documented.

On Tue, Feb 17, 2015 at 8:52 AM, Dirk Eddelbuettel <notifications@github.com

PR #259, opened 2015-02-19
Allow new_env() to create an environment with a specified parent

lionel-:
Hello,

This is a small change to have a more useful way of creating environments with Rcpp.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-02-19  Lionel Henry  <lionel.hry@gmail.com>
+
+        * inst/include/Rcpp/Environment.h Allow new_env() to create an
+        environment with a specified parent
+
 2015-02-19  JJ Allaire  <jj@rstudio.org>
 
         * vignettes/Rcpp-attributes.Rnw: Add note on using inline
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Thanks for the contribution.  Do you have a usage example?   And/or test case we can add to the unit tests?

lionel-:
For example you can capture the calling environment to your cpp function from its R wrapper with `parent.frame()`. Then you create a new environment (call it `my_env`) that has the calling env as parent. You can put several things in there, knowing that it won't have side effects, and then evaluate some R code with `Rf_eval()` in this environment. Variable lookup will first take place in `my_env`, then in the calling env, then its parent and so on.

I could add a test that checks that the parent is the correct one if you want.

lionel-:
Here you go.

lionel-:
wait a minute, I made a mistake in there

why did the test not fail? :/

lionel-:
ok, I should have used `checkIdentical()`

should be good now

eddelbuettel:
Looks good.  Odd indeed that `checkEquals()` failed there.

lionel-:
Thanks for your quick review!

PR #260, opened 2015-02-20
Function constructors that look up functions in an environment or in a namespace

lionel-:
Before we needed to have
```{cpp}
    Environment ns = Environment::namespace_env("Rcpp") ;
    Function fun = ns["sourceRcpp"];
```

With this we can just do
```{cpp}
    Function fun("sourceRcpp", "Rcpp");
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-02-20  Lionel Henry  <lionel.hry@gmail.com>
+
+        * inst/include/Rcpp/Function.h New Function constructors that will
+        perform function-lookup in an environment or in a namespace.
+
 2015-02-19  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION: Bump Version: and Date:
--- a/inst/include/Rcpp/Function.h
+++ b/inst/include/Rcpp/Function.h
@@ -49,14 +49,29 @@ namespace Rcpp{
         }
 
         /**
-         * Finds a function, searching from the global environment
+         * Finds a function. By default, searches from the global environment
          *
          * @param name name of the function
+         * @param env an environment where to search the function
+         * @param ns name of the namespace in which to search the function
          */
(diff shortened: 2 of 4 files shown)


jjallaire:
I believe that this change will break the ABI (because it adds overloads to the constructor). We can be 100% sure by running something like the ABI compliance checker (https://github.com/lvc/abi-compliance-checker/), which we should probably add to our test suite.

In any case, we don't have a policy that says we will *never* break the ABI, but we would like to do it all at once (collecting many ABI breaking changes) so as to minimize/contain the disruption this causes (a lot of mysterious crashing when users update Rcpp but not the packages that depend on it).

If this does in fact break the ABI we should do one of two things:

1) Submit another change that doesn't break it (i.e. add global functions which construct function objects in the manner you are hoping for); or

2) Give the PR an "ABI" tag and wait to merge it until we have other worthy ABI breaking changes.

lionel-:
oh ok. I'm new to compiled languages so I'm not familiar with this ABI stuff.

I'm ok with waiting for another ABI breaking changes.

eddelbuettel:
Last time we accidentally broke the ABI I tried to add a test with a prebuilt package. That was a good idea "in theory" but I (still) don;t know if it catches enough stuff "in practice".

jjallaire:
I think that approach will only catch ABI breakages affecting the code in the prebuilt package. So it would have only catched this one if the code in the prebuilt package calls the Function constructor.

eddelbuettel:
Yes, that makes a lot of sense. Good old (lack of) coverage problem.

jjallaire:
It's possible that there is no ABI breakage after-all since these changes are all within header files. Once we get a bit more tooling in place for this we can revisit the PR and take it if there is no ABI change.

jjallaire:
The fact that these changes are header-only do indeed make them ABI safe.

lionel-:
great :-)

PR #265, opened 2015-02-24
sourceCpp: build c++ files corresponding to local shared header files

jjallaire:
This brings sourceCpp closer into line with source("foo.R") in that multiple translation units can be strung together from within the "main" C++ translation unit that is sourced. While packages are still preferred for binary redistribution, unit testing, etc. this extension to sourceCpp will make working with Rcpp at the REPL a bit more scalable. This also matches better with user intuition about what *should* work as users could never fathom why compiling in sourceCpp didn't allow for simple local code sharing.

A bit more detail on the mechanism: local includes (e.g. "#include foo.hpp" with quotes rather than <>) are considered an implicit specification that you want e.g. foo.cpp included in the compilation if it exists. We still track the file modification times for all dependent hpp and cpp files so rebuilds will always occur when necessary.

This should only affect sourceCpp so the revdep check against CRAN will not likely catch any regressions. I rebuilt the entire Rcpp Gallery with the change (and discovered one subtle defect that is now fixed by doing so).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,7 +2,8 @@
 
         * src/attributes.cpp: Guard against includes referencing themselves
         (and thus creating an endless loop of include processing); Process
-        attributes in included files.
+        attributes in included files; Automatically build implementation files
+        (*.cc; *.cpp) corresponding to local header files if they exist.
         
 2015-02-20  Lionel Henry  <lionel.hry@gmail.com>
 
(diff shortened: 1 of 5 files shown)


eddelbuettel:
Nice.  Noobs and non-package-builder will appreciate this.

What ever happened to "just tell them to build packages" mantra, though?   ;-)

jjallaire:
Packages are of course preferred, but I still feel like (rightly or
wrongly) most users feel like it's a huge step to write a package. I also
think there are some challenges related to incremental development (working
on both the standalone file and the re-usable package(s) at the same time)
and of course linking (if the types in the package aren't header only).

So in short I think packages are the right answer but not always slam dunk
at a practical level. If we can make sourceCpp a bit more scalable then I
think it will help the overall cause of using C++ with R and hopefully as a
byproduct result in more packages.




On Tue, Feb 24, 2015 at 1:31 PM, Dirk Eddelbuettel <notifications@github.com

PR #269, opened 2015-03-01
Feature/print function

eddelbuettel:
This adds a new function print() as a wrapper for Rf_PrintValue

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2015-03-01  Dirk Eddelbuettel  <edd@debian.org>
+
+        * src/api.cpp: New function print() as a wrapper around Rf_PrintValue()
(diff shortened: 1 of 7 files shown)


jjallaire:
One question: why not make this an inline function and dispense with registration/routines bits?

eddelbuettel:
That is what I did initially. 

I have no strong feeling either way.

jjallaire:
Got it. My bias is to keep as much inline as possible to minimize the probability of version mishmash issues. That said, if someone is using this function they clearly depend on Rcpp 0.11.5 and so the results of running against 0.11.4 are at best undefined. Also agree that having a simple worked example is valuable.

eddelbuettel:
Yeppers -- agree on all counts.

PR #274, opened 2015-03-11
Use Rtools 3.3 for compilation under R 3.2

jjallaire:


eddelbuettel:
Thanks for doing that.  I guess 0.11.6 needs to go out before R 3.2.0 then?

jjallaire:
It will be okay as long as 1 of these 3 things is true:

1. The user has Rtools 3.2 (which is still valid for R 3.2)

2. The user is running in RStudio (which in older versions will bootstrap
them to R 3.2 and in newer versions will bootstrap them to R 3.3).

3. The user installs Rtools 3.3 to C:\Rtools and puts C:\Rtools\bin on
their PATH

So any user that is running R 3.2 AND installs Rtools 3.3 AND does not
install to C:\Rtools AND is not running under RStudio will not be able to
use sourceCpp. This should be a pretty small group and workarounds are
possible (e.g. just do an old school style Rtools install which is probably
what most non-RStudio users are doing anyway) so I don't think we need to
be out before but more within a couple of weeks.



On Wed, Mar 11, 2015 at 2:03 PM, Dirk Eddelbuettel <notifications@github.com

eddelbuettel:
Thanks for the clarification, and for spelling it out here.

PR #275, opened 2015-03-12
move more headers into major/minor protection

kevinushey:
This PR moves three of our headers into the `major/minor` 'protection' block. When `-std=c++11` is specified, those headers pull in the `major/macro` defines (not sure what chain gets them). This PR fixes that.


eddelbuettel:
Excellent -- will merge.

PR #279, opened 2015-03-18
Default ctor for Matrix return 0x0 matrix

1dayac:
Commit to fix the issue - "Matrix default ctor does not make a matrix #129"

### DIFF
--- a/inst/include/Rcpp/vector/Matrix.h
+++ b/inst/include/Rcpp/vector/Matrix.h
@@ -44,7 +44,7 @@ public:
     typedef typename VECTOR::Proxy Proxy ;
     typedef typename VECTOR::const_Proxy const_Proxy ;
 
-    Matrix() : VECTOR() {}
+    Matrix() : VECTOR(Dimension(0, 0)), nrows(0) {}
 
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Looks good to me.  Does not do a whole lot, but appears to fix #129.

Any seconds -- @kevinushey or @jjallaire ?

kevinushey:
Looks good to me.

eddelbuettel:
Now folding in PR after second LGTM over email.

As for GSoC we need more discussion and a plan.  If you are interested, I would invite to open the discussion with a new issue ticket in the Rcpp GitHub repository where you flesh out the proposal, and we can comment.

We on the Rcpp Core are prudent about making breaking changes.  As such, some proposals and ideas are better than others.  The _new DataFrame class_ may be a good candidate as we can test it at length in a add-on package rather than having to make changes in Rcpp itself.  For the same reason _R_xlen_t_ is tricky.  I just expanded the [Extend Rcpp](https://github.com/rstats-gsoc/gsoc2015/wiki/Extend-Rcpp) entry on the GSoC WIki a little in regards. to this.

PR #280, opened 2015-03-19
Using UTF-8 in String ctor

thirdwing:
This is for https://github.com/RcppCore/Rcpp/issues/189 . String ctor now uses UTF-8.

There seems much more work if we want to support different encoding in String.

Feel free to close this PR.

### DIFF
--- a/inst/include/Rcpp/String.h
+++ b/inst/include/Rcpp/String.h
@@ -63,7 +63,8 @@ namespace Rcpp {
         }
 
         /** construct a string from a single CHARSXP SEXP */
-        String(SEXP charsxp) : data(charsxp), valid(true), buffer_ready(false) {
+        String(SEXP charsxp) : valid(true), buffer_ready(false) {
+            data = Rf_mkCharCE(Rf_translateCharUTF8(charsxp), CE_UTF8) ;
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Looks good to me though utf-8 support probably needs a _lot_ more work in more places.  

Any thoughts on how to get from here to there?

Any seconds on whether to fold this in, @kevinushey or @jjallaire ?

kevinushey:
I think this is a step in the right direction, although I think we need to find a compromise between:

1. `String` objects that just store some bytes, alongside whatever declared encoding they came with, and
2. `String` objects that are always internally stored as UTF-8, but are converted back to the appropriate locale on demand.

I think performance-intensive applications will want to avoid round-trip translations between various encodings, so I am somewhat hesitant to accept this PR right away. Thoughts, @jjallaire?

jjallaire:
The other problem with this is that on Windows the OS interfaces, R itself,
and many other libraries assume that single-character (char) strings use
the system encoding. If we start auto-converting to UTF-8 we will surely
break things that currently work.

So any conversion to UTF8 will need to explicit (e.g. an encoding parameter
on the constructor, a static construction function, etc.)

thirdwing:
Thanks @jjallaire  and @kevinushey .There are many things I don't know before.

PR #290, opened 2015-04-22
don't enable C++11 features in tinyformat

kevinushey:
This should fix compiler errors as seen [here](https://github.com/hadley/readxl/issues/78) -- we force C++98 compliance for `tinyformat`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2015-04-22  Kevin Ushey  <kevinushey@gmail.com>
+
+        * inst/include/Rcpp/utils/tinyformat.h: don't use C++11 features
+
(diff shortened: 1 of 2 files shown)


eddelbuettel:
I was trying to to follow up on the link given by Kevin. Is this a side-effect of the "turning C++11 on without requesting C++11" hack?

In any event the PR is good.  More defensive compilation can't hurt given that these wonderful CentOS installation from the 1960s (or so they feel ;-) are still around.

kevinushey:
That's right -- IIUC, using `-std=c++0x` causes `tinyformat` to pick up that C++11 features are 'enabled', and so attempts to use them. Unfortunately, on older compilers, `tinyformat` ends up using C++11 features that aren't fully implemented, and so we get the odd errors we saw there.

PR #303, opened 2015-06-02
R_xlen_t as the vector length to support long vector

thirdwing:
Most of changes are about `R_xlen_t` and `Rf_xlength`.

Please take a careful review on `Vector.h` and `is_arithmetic` added. 

The type trait is used because many packages depend the implicit conversion in `Vector` constructor.

### DIFF
--- a/inst/include/Rcpp/api/meat/DottedPairImpl.h
+++ b/inst/include/Rcpp/api/meat/DottedPairImpl.h
@@ -53,7 +53,7 @@ namespace Rcpp{
 		} else {
 			if( ref.isNULL( ) ) throw index_out_of_bounds() ;
 
-                        if( static_cast<R_xlen_t>(index) > ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
+                        if( static_cast<R_xlen_t>(index) > ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds() ;
 
 			size_t i=1;
 			SEXP x = ref.get__() ;
@@ -70,7 +70,7 @@ namespace Rcpp{
     template <typename T>
 	void DottedPairImpl<CLASS>::replace( const int& index, const T& object ) {
 	    CLASS& ref = static_cast<CLASS&>(*this) ;
-        if( static_cast<R_xlen_t>(index) >= ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
+        if( static_cast<R_xlen_t>(index) >= ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds() ;
 
         Shield<SEXP> x( pairlist( object ) );
         SEXP y = ref.get__() ;
@@ -84,7 +84,7 @@ namespace Rcpp{
 	template <typename CLASS>
     void DottedPairImpl<CLASS>::remove( const size_t& index ) {
         CLASS& ref = static_cast<CLASS&>(*this) ;
-        if( static_cast<R_xlen_t>(index) >= Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
+        if( static_cast<R_xlen_t>(index) >= Rf_xlength(ref.get__()) ) throw index_out_of_bounds() ;
         if( index == 0 ){
             ref.set__( CDR( ref.get__() ) ) ;
         } else{
--- a/inst/include/Rcpp/api/meat/DottedPairImpl.h
+++ b/inst/include/Rcpp/api/meat/DottedPairImpl.h
@@ -53,7 +53,7 @@ namespace Rcpp{
 		} else {
 			if( ref.isNULL( ) ) throw index_out_of_bounds() ;
 
-			if( static_cast<R_len_t>(index) > ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
+                        if( static_cast<R_xlen_t>(index) > ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
 
 			size_t i=1;
 			SEXP x = ref.get__() ;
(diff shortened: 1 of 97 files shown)


inst/include/Rcpp/traits/is_arithmetic.h
```diff
@@ -0,0 +1,101 @@
...
+
+    template<>
+    struct is_arithmetic<const long> : public true_type { };
+    
+    template<>
+    struct is_arithmetic<unsigned long> : public true_type { };
+
+    template<>
+    struct is_arithmetic<const unsigned long> : public true_type { };
+
```
eddelbuettel: The following lines need to be protected by something like `#if define(RCPP_HAS_LONG_LONG_TYPES)`. We may even use `rcpp_long_long_type` etc -- see the sibbling file `traits/longlong.h`.

eddelbuettel:
Not a single failure is due to the `R_xlen_t` change.   

A few are due to my 'cache' of packages having been emptier than usual.   Rerunning now, previous result summary committed in rcpp-logs repo.

inst/include/Rcpp/api/meat/DottedPairImpl.h
```diff
@@ -53,7 +53,7 @@ namespace Rcpp{
 		} else {
 			if( ref.isNULL( ) ) throw index_out_of_bounds() ;
 
-			if( static_cast<R_len_t>(index) > ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
+                        if( static_cast<R_xlen_t>(index) > ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
```
kevinushey: Here you cast to `R_xlen_t` but check against `::Rf_length`. Is that intentional? Should we be using `::Rf_xlength`?

inst/include/Rcpp/api/meat/DottedPairImpl.h
```diff
@@ -70,7 +70,7 @@ namespace Rcpp{
     template <typename T>
 	void DottedPairImpl<CLASS>::replace( const int& index, const T& object ) {
 	    CLASS& ref = static_cast<CLASS&>(*this) ;
-        if( static_cast<R_len_t>(index) >= ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
+        if( static_cast<R_xlen_t>(index) >= ::Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
```
kevinushey: Same as above.

inst/include/Rcpp/api/meat/DottedPairImpl.h
```diff
@@ -84,7 +84,7 @@ namespace Rcpp{
 	template <typename CLASS>
     void DottedPairImpl<CLASS>::remove( const size_t& index ) {
         CLASS& ref = static_cast<CLASS&>(*this) ;
-        if( static_cast<R_len_t>(index) >= Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
+        if( static_cast<R_xlen_t>(index) >= Rf_length(ref.get__()) ) throw index_out_of_bounds() ;
```
kevinushey: Same as above.

inst/include/Rcpp/sugar/functions/is_na.h
```diff
@@ -33,11 +33,11 @@ class IsNa : public ::Rcpp::VectorBase< LGLSXP, false, IsNa<RTYPE,NA,VEC_TYPE> >
 
 	IsNa( const BASE& obj_) : obj(obj_){}
 
-	inline int operator[]( int i ) const {
+        inline int operator[]( R_xlen_t i ) const {
```
kevinushey: quip: indentation changed here

inst/include/Rcpp/sugar/matrix/as_vector.h
```diff
@@ -30,8 +30,8 @@ inline Rcpp::Vector<RTYPE>
 as_vector__impl( MatrixBase<RTYPE,NA,T>& t, Rcpp::traits::false_type ){
     T& ref = t.get_ref() ;
     int nc = ref.ncol(), nr = ref.nrow() ;
-    Vector<RTYPE> out (nr*nc) ;
-    int k =0;
+    Vector<RTYPE> out (((R_xlen_t)nr) * nc) ;
```
kevinushey: We usually prefer C++-style casts (`static_cast<R_xlen_t>(nr)`) over C-style casts.

inst/include/Rcpp/sugar/matrix/as_vector.h
```diff
@@ -43,7 +43,7 @@ template <int RTYPE, bool NA, typename T>
 inline Rcpp::Vector<RTYPE>
 as_vector__impl( MatrixBase<RTYPE,NA,T>& t, Rcpp::traits::true_type ){
     Matrix<RTYPE>& ref = t.get_ref() ;
-    int size = ref.ncol()*ref.nrow() ;
+    R_xlen_t size = ((R_xlen_t)ref.ncol())*ref.nrow() ;
```
kevinushey: Same as above.

inst/include/Rcpp/sugar/matrix/as_vector.h
thirdwing: Of course, C++-style is better.

kevinushey:
I made a quick review and all looks well to me. Thanks for your hard work -- this has been sorely needed for quite some time!

I'll give a :+1: to merging this after @eddelbuettel's tests run through; we might want to do a second pass and scrutinize the use of `int` anywhere in the code base that's used for indexing (since we likely either want `R_len_t` or `R_xlen_t`).

Also, we could try rebuilding the Rcpp gallery with this just as one last thing to check.

inst/include/Rcpp/sugar/functions/is_na.h
thirdwing: Sorry for the mistake. The original indention is tab and I used spaces here. Any preference for indention?

eddelbuettel: We are not terribly consistent.  Four spaces.  Doing without tabs seems better (as we learned over the years), but we are religious enough to filter all files.  

I use Emacs to adjust indentation when I open a file and see something wrong.

jjallaire:
I've merged this into feature/r_xtlen_t-vector-length here: 

https://github.com/RcppCore/Rcpp/tree/feature/r_xlen_t-vector-length

We should make further fixes and tweaks on that branch. Hopefully this can be merged to master soon!

PR #318, opened 2015-07-02
add .clang-format

kevinushey:
This PR adds a `.clang-format` file which can be used alongside `clang-format` to format files according to what we can call the 'official' Rcpp format rules. Preferably, for anything new contributed to Rcpp, we should use this to enforce consistent styling.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -12,4 +12,5 @@ vignettes/notyet
 doxyfile
 .travis.yml
 .dir-locals.el
+.clang_format
 vignettes/getCurrentVersionsOfCitedPackages.R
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Thanks for starting this.  Do we have a concrete "before/after" for a particularly "oddly" formatted file?

eddelbuettel:
@kevinushey I am going to merge this as we all agree it is a good idea.

That said, it would probably be an equally good idea to show a particular file 'before' and 'after', especially while we (as I suspect) iterate over what the final 'format spec' becomes.   Do you want to point at a particular file?   Or, maybe, even "draw" up a fictitious file just to show how we'd indent ideally?

PR #319, opened 2015-07-02
inject message recorders into Rcpp namespace on eval

kevinushey:
NOTE: There is some black magic in this PR, so be warned...

This PR adds a mechanism for injecting R functions into the `Rcpp` namespace, and uses it to ensure that message recorders are injected and used for recording messages.

The injection of these functions should help ensure compatibility with newer versions of Rcpp that might not have these functions in the namespace by default.

That said, the environment unlocking / re-locking may be dangerous...

### DIFF
--- a/inst/include/Rcpp/api/meat/Rcpp_eval.h
+++ b/inst/include/Rcpp/api/meat/Rcpp_eval.h
@@ -20,13 +20,110 @@
 
 #include <Rcpp/Interrupt.h>
 
-namespace Rcpp{
+namespace Rcpp {
+
+namespace internal {
+
+#define RCPP_FRAME_LOCK_MASK (1<<14)
+#define RCPP_FRAME_IS_LOCKED(e) (ENVFLAGS(e) & RCPP_FRAME_LOCK_MASK)
+#define RCPP_LOCK_FRAME(e) SET_ENVFLAGS(e, ENVFLAGS(e) | RCPP_FRAME_LOCK_MASK)
+#define RCPP_UNLOCK_FRAME(e) SET_ENVFLAGS(e, ENVFLAGS(e) & (~ RCPP_FRAME_LOCK_MASK))
+
+class UnlockFrameScope {
+public:
+    
+    UnlockFrameScope(const std::string& name) {
+        Shield<SEXP> nameSEXP(::Rf_mkString(name.c_str()));
+        ns_ = ::R_FindNamespace(nameSEXP);
+        if (ns_ && TYPEOF(ns_) == ENVSXP)
+            RCPP_UNLOCK_FRAME(ns_);
+    }
+    
+    ~UnlockFrameScope() {
+        if (ns_ && TYPEOF(ns_) == ENVSXP) {
+            RCPP_LOCK_FRAME(ns_);
+        }
+    }
+    
+private:
+    
+    SEXP ns_;
+    
+};
+
+inline bool inject(const std::string& functionName,
+                   const std::string& namespaceName,
+                   const std::string& functionDefn)
+{
+    UnlockFrameScope rcppScope(namespaceName);
+    
+    std::string code = std::string() +
+        "ns <- asNamespace('" + namespaceName + "')\n" +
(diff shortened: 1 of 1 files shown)


inst/include/Rcpp/api/meat/Rcpp_eval.h
```diff
@@ -20,13 +20,110 @@
...
+namespace Rcpp {
+
+namespace internal {
+
+#define RCPP_FRAME_LOCK_MASK (1<<14)
+#define RCPP_FRAME_IS_LOCKED(e) (ENVFLAGS(e) & RCPP_FRAME_LOCK_MASK)
+#define RCPP_LOCK_FRAME(e) SET_ENVFLAGS(e, ENVFLAGS(e) | RCPP_FRAME_LOCK_MASK)
+#define RCPP_UNLOCK_FRAME(e) SET_ENVFLAGS(e, ENVFLAGS(e) & (~ RCPP_FRAME_LOCK_MASK))
+
+class UnlockFrameScope {
```
jjallaire: Better make this class noncopyable!

jjallaire:
I like effort to spare our dependencies from sudden and unexpected breakage! I don't have a strong handle on the various mechanisms possible for injecting code into a locked environment so can't comment on that piece. @eddelbuettel What's your take?

eddelbuettel:
I don't really play in those sandboxen either, so at best a firm "unsure" from me.

kevinushey:
There is no exported mechanism for unlocking an environment, although an appropriate macro is listed here: https://github.com/wch/r-source/blob/trunk/src/main/envir.c#L105-L108, and that's what I borrow from to make this bit of code possible.

@wch might have some information to share as I think he's messed around in this realm before. Just to further set the landscape of what we're trying to do:

When we evaluate an R function within Rcpp (from C++), we do it by wrapping the call in the moral equivalent of:

```
withCallingHandlers(
    warning = .rcpp_warning_recorder,
    <executeCall>
)
```

That call is generated within the header files. The problem there is if a package is compiled against a newer Rcpp, but an older one is currently loaded, it's possible that `.rcpp_warning_recorder` doesn't exist and so this call will fail.

To alleviate this, I'm trying to amend our evaluation code to perform a one-time check of the existence of e.g. `.rcpp_warning_recorder`; if it doesn't exist we 'inject' it by unlocking the Rcpp namespace, assigning the appropriate function definition, and then locking it back up at the end.

This is also done in a 'header-only' way by generating the functions as strings, parsing those strings, and then assigning the parsed result (which is the function we wanted to generate) into the Rcpp namespace.

jjallaire:
@wch Lest you think this sounds sketchy or hacky, recall the carnage that
occurs with plyr, shiny, httpuv, etc. when:

1. A new version of Rcpp with headers that depend on functions in the
package is released;

2. A package (e.g. shiny) is recompiled on CRAN with the new version

3. User installs shiny and already has Rcpp (older version)

4. BOOM! New headers meet old Rcpp library and package (e.g. shiny) stops
working.

This hackery is to avoid that carnage which is no fun for anyone.

wch:
I believe it's a real problem -- it's closely related to one that I asked about on r-devel just last night, except that my question stayed within the realm of pure R code:
  http://r.789695.n4.nabble.com/Are-downstream-dependencies-rebuilt-when-a-package-is-updated-on-CRAN-td4709349.html

I don't think it should cause any problems unlock the namespace and insert a new object. If you're modifying an *exported* object, there's more to it, though.

jjallaire:
No, were not updating an exported object.

Kevin, I think this is good to merge if you want to go ahead and pull the
trigger.

kevinushey:
@jjallaire, I'm going to make a couple more changes -- I think we should tag injected functions with the version of Rcpp that injected them (ie as an attribute), and also use that when injecting functions (so that e.g if we needed to change them in a future version of Rcpp we can still overwrite them; currently we only inject if it doesn't already exist)

jjallaire:
Okay, that sounds like a very, very good idea!

J

Issue #223, opened 2015-01-02
Sugar var() only knows numeric


eddelbuettel:
This fails to compile:

```{.R}
cppFunction('double myVar(IntegerVector x) { return(var(x)); }') 
```

wush978:
Dear Dirk,

I study the related header and have some ideas of implementing a template specialization function `sugar::Var<INTSXP,NA,T> var( const VectorBase<INTSXP,NA,T>& t)` to resolve this issue.

Directly replace the `REALSXP` to `INTSXP` will produce the wrong result because the `Minus_Vector_Primitive<RTYPE,NA,T> mm( object, m) ;` will trim result to `int`. We need to rewrite the `get` method.

The simplest approach is to internally transform the `IntegerVector` to `NumericVector` and apply the original `var`. 

How do you think?

eddelbuettel:
But an internal cast would create a copy and that is not ideal.

I think the best way, sadly, is to add four specialization as I did for `mean()` (cf #219 and #221).   It's one of those things I may get to one day when there is nothing more urgent on the plate.  

If you want to play with it, this (like a few of the other open issue tickets) is a nice one to create a contribution.  

And thanks also for switching to discussion first rather than pull request first ;-)

wush978:
I'll take this. And thanks for your hint, I'll study your implementation of `mean()` as a good start.

Ya, discuss first does help!

wush978:
In R, the `stats::var` drops the imaginary part, so I do the same thing in 0eb6b3a . 

If the unit test in 41fea1e is all what we need (test 4 different type and compare the result with `stats::var`), I will start to test the implementation 0eb6b3a on major machines and compilers.

wush978:
Dear all,

I don't know if the test is adaquate or not, so I come here for helps. It would be great if any of you can spend 1 minute to review 41fea1e . It compare the result of `Rcpp::var` on 4 different types with the `stats::var`.

If the unit test is all what we need, could I send a PR after making the test passed on major machines and compilers?

eddelbuettel:
Yes, that is the general idea. 

In practice most people only use `NumericVector` which is how we got by so far.  But it would be nice to complement it.  Not sure anybody ever uses `var(myLogicalVector)` ... but hey, nice to have.

MattPD:
Question: Hasn't this commit per chance introduced a regression corresponding to the switch from formula (1) to formula (2) in the following:
[...]

PR #320, opened 2015-07-06
Update var.h

MattPD:
Variance: Change from the unstable formula back to the stable (two-pass) formula.
Details: https://github.com/RcppCore/Rcpp/issues/223

### DIFF
--- a/inst/include/Rcpp/sugar/functions/var.h
+++ b/inst/include/Rcpp/sugar/functions/var.h
@@ -34,10 +34,10 @@ public:
     Var( const VEC_TYPE& object_ ) : object(object_){}
 
     double get() const{
-        double
-            m = mean(object).get(),
-            ssq = sum( pow(object, 2.0) ).get();
-        return (ssq - m * m * object.size()) / (object.size() - 1);
+        const double
+            average = mean(object).get(),
+            sum_squared_deviations = sum( pow(object - average, 2.0) ).get();
+        return sum_squared_deviations / (object.size() - 1);
     }
 
 private:
--- a/inst/include/Rcpp/sugar/functions/var.h
+++ b/inst/include/Rcpp/sugar/functions/var.h
@@ -34,10 +34,12 @@ public:
     Var( const VEC_TYPE& object_ ) : object(object_){}
 
     double get() const{
-        const double
-            average = mean(object).get(),
-            sum_squared_deviations = sum( pow(object - average, 2.0) ).get();
-        return sum_squared_deviations / (object.size() - 1);
+        const double average = mean(object).get();
+        const R_xlen_t sample_size = object.size();
+        double sum_squared_deviations = 0.0;
+        for (R_xlen_t i = 0; i != sample_size; ++i)
+            sum_squared_deviations += std::pow(object[i] - average, 2.0);
+        return sum_squared_deviations / (sample_size - 1);
     }
 
 private:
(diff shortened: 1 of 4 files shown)


inst/include/Rcpp/sugar/functions/var.h
```diff
@@ -34,10 +34,10 @@ class Var : public Lazy< double , Var<RTYPE,NA,T> > {
...
     double get() const{
-        double
-            m = mean(object).get(),
-            ssq = sum( pow(object, 2.0) ).get();
-        return (ssq - m * m * object.size()) / (object.size() - 1);
+        const double
+            average = mean(object).get(),
+            sum_squared_deviations = sum( pow(object - average, 2.0) ).get();
+        return sum_squared_deviations / (object.size() - 1);
     }
```
eddelbuettel: :+1:

Could you do the same for the complex case (next few lines) and also add ChangeLog entry in the PR?

inst/include/Rcpp/sugar/functions/var.h
```diff
@@ -34,10 +34,10 @@ class Var : public Lazy< double , Var<RTYPE,NA,T> > {
     Var( const VEC_TYPE& object_ ) : object(object_){}
 
     double get() const{
-        double
-            m = mean(object).get(),
-            ssq = sum( pow(object, 2.0) ).get();
-        return (ssq - m * m * object.size()) / (object.size() - 1);
```
eddelbuettel: I sort-of missed we switched to this.   My bad.

This can indeed lead to over/under-flow.

inst/include/Rcpp/sugar/functions/var.h
MattPD: Just added a comment, it's indeed next in line in .plan :-)

MattPD: Looks like this might have introduced a regression for `IntegerVector`, investigating.

MattPD: Found the reason.

For `IntegerVector` temporary `object - average` has type `Rcpp::sugar::Minus_Vector_Primitive<13, true, Rcpp::Vector<13, Rcpp::PreserveStorage>>.

In contrast, for `NumericVector` temporary `object - average` has type `Rcpp::sugar::Minus_Vector_Primitive<14, true, Rcpp::Vector<14, Rcpp::PreserveStorage>>`.

This results in invalid deviations calculation (thus resulting in invalid `sum_squared_deviations`).

Rewriting.

eddelbuettel:
I think this one is good to go (though I'd like a `ChangeLog` and/or `inst/NEWS.Rd` entry too).  

Any seconds, @kevinushey or @jjallaire ?

MattPD:
@eddelbuettel: sounds great; done & done! :-)

eddelbuettel:
Sorry:  I fixed this in 384e7e43faae111f892cc944c267c670182039fa but I hope that doesn't create a merge conflict for you now.

MattPD:
Right! Can I fix this on my side?

eddelbuettel:
I would think so by re-syncing your fork with what I have now.

MattPD:
Let's hope this works :-)

eddelbuettel:
Looks good!  While I usually wait for either/both of @kevinushey and @jjallaire to nod, I think this one is good to go -- thanks for the contribution!

MattPD:
:-)

PR #323, opened 2015-07-13
remove calling handlers from Rcpp_eval

kevinushey:
(Sorry for the noisy diff -- the gist of this PR is that we no longer evaluate with `withCallingHandlers`; instead just with `tryCatch`).

This PR supercedes https://github.com/RcppCore/Rcpp/pull/319. The main idea:

1. It appears that when `options(warn = 2)` is set, the warning is converted to an error before escaping the `tryCatch` call and so `tryCatch` handles it properly,

2. When we attach calling handlers to the call, recursive evaluation of R expressions causes loads of problems that make it very difficult to properly forward errors. It looks like R 'does the right thing' when we don't attach calling handlers, so I believe it is safe to just remove them.

AFAIK, a `warning` (or `message`) call hence will not  `longjmp` out of the current context -- only R errors will do that. I think any handler that attempts to call `stop()` from R code will still get caught by our top-level `tryCatch` and so I think we are safe.

The only problem I've found thus far is that messages called using the Rcpp function class are not properly suppressed, e.g. given this:

```cpp
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
void test() {
    
    Function msg("message");
    msg("Rcpp message");
    
    Shield<SEXP> messageSEXP(Rf_mkString("R message"));
    Shield<SEXP> messageCall(Rf_lang2(
        Rf_install("message"),
        messageSEXP
    ));
    
    Rf_eval(messageCall, R_GlobalEnv);
}

// [[Rcpp::export]]
void exec(Function f) { f(); }

/*** R
test()
suppressMessages(test())
*/
```

I get:

```
> Rcpp::sourceCpp('~/Untitled.cpp')

> test()
Rcpp message
R message

> suppressMessages(test())
Rcpp message
```

So I believe this is an artefact of how the `Rcpp::Function` class is written and not of this evaluation mechanism.

### DIFF
--- a/inst/include/Rcpp/api/meat/Rcpp_eval.h
+++ b/inst/include/Rcpp/api/meat/Rcpp_eval.h
@@ -22,98 +22,58 @@
 
 namespace Rcpp {
 
-struct EvalCall {
-    
-    // Related to call object
-    SEXP expr;
-    SEXP env;
-    SEXP result;
-    
-    // Related to error recording / forwarding
-    bool error_occurred;
-    std::string error_message;
-};
+inline SEXP Rcpp_eval(SEXP expr, SEXP env) {
 
-inline void Rcpp_eval(void* data) {
-
-    EvalCall* evalCall = (EvalCall*) data;
-
-    SEXP tryCatchSym               = ::Rf_install("tryCatch");
-    SEXP evalqSym                  = ::Rf_install("evalq");
-    SEXP conditionMessageSym       = ::Rf_install("conditionMessage");
-    SEXP errorSym                  = ::Rf_install("error");
-    
-    // get the Rcpp error recorder
-    SEXP errorRecorder = Rf_findFun(
-        Rf_install(".rcpp_error_recorder"),
-        Environment::Rcpp_namespace()
+    // 'identity' function used to capture errors, interrupts
+    SEXP identity = Rf_findFun(
+        ::Rf_install("identity"),
+        R_BaseNamespace
     );
     
-    if (errorRecorder == R_UnboundValue) {
-        evalCall->error_occurred = true;
-        evalCall->error_message =
-            "Failed to find Rcpp error recorder: please ensure\n"
-            "you have the latest version of Rcpp installed.";
-        return;
+    if (identity == R_UnboundValue) {
+        stop("Failed to find 'base::identity()'");
     }
 
     // define the evalq call -- the actual R evaluation we
     // want to execute
     Shield<SEXP> evalqCall(Rf_lang3(
-        evalqSym,
-        evalCall->expr,
-        evalCall->env
+        ::Rf_install("evalq"),
+        expr,
+        env
     ));
     
     // define the call -- enclose with `tryCatch` so we can record
     // and later forward error messages
-    Shield<SEXP> call(Rf_lang3(
-        tryCatchSym,
+    Shield<SEXP> call(Rf_lang4(
+        ::Rf_install("tryCatch"),
         evalqCall,
-        errorRecorder
+        identity,
+        identity
     ));
-    SET_TAG(CDDR(call), errorSym);
+    SET_TAG(CDDR(call), ::Rf_install("error"));
+    SET_TAG(CDDDR(call), ::Rf_install("interrupt"));
 
     // execute the call
     Shield<SEXP> res(::Rf_eval(call, R_GlobalEnv));
 
     // check for error
-    if (error_occured()) {
-        Shield<SEXP> current_error        (rcpp_get_current_error());
-        Shield<SEXP> conditionMessageCall (::Rf_lang2(conditionMessageSym, current_error));
-        Shield<SEXP> condition_message    (::Rf_eval(conditionMessageCall, R_GlobalEnv));
-        evalCall->error_occurred = true;
-        evalCall->error_message = std::string(CHAR(::Rf_asChar(condition_message)));
-    } else {
-        evalCall->error_occurred = false;
-        evalCall->result = res;
+    if (Rf_inherits(res, "error")) {
+        
+        Shield<SEXP> conditionMessageCall(::Rf_lang2(
+           ::Rf_install("conditionMessage"),
+           res
+        ));
+        
+        Shield<SEXP> conditionMessage(::Rf_eval(conditionMessageCall, R_GlobalEnv));
+        throw eval_error(CHAR(STRING_ELT(conditionMessage, 0)));
     }
     
-    // Reset the current error for the next evaluation
-    reset_current_error();
-
-}
-
-inline SEXP Rcpp_eval(SEXP expr, SEXP env) {
-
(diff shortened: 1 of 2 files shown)


jjallaire:
One important note was that we had two failing tests (one in dplyr I think)
until we added the withCallingHandlers to collect warnings. The tests that
failed had to do with a call to testthat::expect_warning where the warning
didn't occur. Note that it did *eventually* occur just not within the
test's evaluation context.

I could easily argue (and I think did at the time we had the failing tests)
that this is by design and the tests should be updated to reflect this. In
other words, the withCallingHandlers cure may have turned out to be worse
than the disease of delayed propagation of warnings. In any case, if we do
go with this PR we need to deal with those failing tests as part of the
effort.

kevinushey:
I think I agree -- at least, if the only fallout from this change is that some tests might need to be updated in client packages, that seems better than the pain of properly handling warning and message handlers.

kevinushey:
@jjallaire, thinking more, the initial reason we moved to `R_ToplevelExec` was so we could catch interrupts and ensure they didn't force a `longjmp` (cc: https://github.com/RcppCore/Rcpp/pull/244) -- is that correct?

However, I just discovered that we should be able to capture interrupts at the R level with `tryCatch`, e.g.

```R
tryCatch(Sys.sleep(10), interrupt = function(e) print(e))
```

So I wonder if the _real_ solution here is to just add an interrupt handler to our wrapping `tryCatch` call to ensure that doesn't cause a `longjmp` out of the context of the current execution? If that's the case, we can get rid of the `R_ToplevelExec` call and ensure that pre-existing handlers and what-not are properly respected.

jjallaire:
It's to catch interrupts and to catch Rf_error / stop calls

kevinushey:
@jjallaire my point here is that we can actually catch interrupts within an R-level `tryCatch` (and so we can add a handler there) and also `Rf_error` calls with the `error` handler.

I believe this implies that we don't actually need `R_ToplevelExec` if we're wrapping calls in `tryCatch` -- we can just set `error` and `interrupt` handlers and that will protect us from any `longjmp`ing, and any pre-existing condition handlers (e.g. for warnings or messages) will just be properly handled.

Any handler that attempts to throw an error or interrupt will also be caught by our enclosing `tryCatch`.

In other words -- instead of using `R_ToplevelExec` we can just use `tryCatch` with `error` and `interrupt` handlers, and we avoid `longjmp`s from any R code executed within an Rcpp evaluation.

kevinushey:
Ah, but the `tryCatch` doesn't capture `Rf_error` calls (suddenly that conversation is coming back to me); hence why we adopted `R_ToplevelExec`.

It seems unfortunate that we don't have a way of preserving existing calling handlers but also 'capturing' `longjmp`s.

romainfrancois:
Well it's all there on the contexts api R is keeping to itself, because essentially nobody can be trusted to use it right. 😡

jjallaire:
So it seems like we're back to `R_ToplevelExec`, is that correct?

romainfrancois:
I think kevin's proposal is right.

kevinushey:
Thinking about this a bit more, I think that either `tryCatch` or a lower-level handler attachment through `.Internal(.addCondHands(...))` is the correct solution here.

This PR currently works for all evaluation that happens through `Rcpp_eval` -- we are protected from `longjmp`s as the `tryCatch` handlers will catch those errors before a `longjmp` to top level can occur. If we wanted to avoid the overhead of `tryCatch`, we could just mock the same process using `.Internal(.addCondHands(...))`.

The other option is to use `R_ToplevelExec`, but the problem with that function is that is also detaches all existing handlers (something undesired for people who want to emit / control message and warning output from Rcpp function calls, which we've definitely seen). Since we _can_ capture both errors and interrupts using R handlers (thereby avoiding `longjmp`s) I think that's preferable to using `R_ToplevelExec`.

(Of course, if we have access to R's contexts, we could use `R_ToplevelExec` to create a top-level context and then re-attach existing handlers -- this is what @romainfrancois has effectively done in Rcpp11 here: https://github.com/Rcpp11/Rcpp11/blob/master/inst/include/Rcpp/Context.h#L92-L121)

The error I was thinking of before -- inability to catch top-level calls to `Rf_error()` -- is something that I don't think we can protect users from (ie, if a user attempts to call a `longjmp`ing R API function, they either need to make sure the evaluation is 'protected' with `Rcpp_eval` or use `tryCatch` / `R_ToplevelExec` themselves). There are a couple 'blunt' shields we could use, however:

- Redefine `Rf_error()` to `Rcpp::stop` -- this should be okay since anyone using Rcpp will be using it in a C++ compilation unit, and so any calls to `Rf_error()` would effectively be 'wrong',

- Add an `executeSafely` function that takes a lambda function and performs a `tryCatch` or `R_ToplevelExec` to ensure any R API functions called can't `longjmp` out of the current executing context.

jjallaire:
This all makes good sense to me. Note that before we commit anything
we should probably do the full revdep check.

In terms of redefining Rf_error I think if we can do it cleanly that's
a great idea. Users will continue to call Rf_error from C++ code from
now until eternal as regrettably we can't expect people to RTFM about
this.

I think executeSafely is a nice idea (we have that in RStudio for
times you absolutely need to call R code that might longjmp). I
wouldn't pass it a lambda though but rather use the same interface as
R_ToplevelExec i.e. "void (*fun)(void *), void *data)" for broader
compatibility. In RStudio we wrap some of this nastiness by allowing
you to pass a boost::function but we can't do that in Rcpp -- perhaps
we could conditionally compile a labmda version when C++11 is
supported?

eddelbuettel:
Just wanted to say that I like where this is going ...    Can possibly turn on a rev.dep this eve/tomorrow.

Rcpp now at 411 reverse depends (where I remove 5 false positives).

kevinushey:
Awesome -- thanks for handling the revdep check, @eddelbuettel.

eddelbuettel:
@kevinushey:  Oddly, revdep stopped at midnight.  Seems spurious. Restarted.  Score was at time was 268 out of 414 tests with 248 good and 20 bad.  None of the 20 seemed Rcpp-related: mostly missing Suggests in tests or examples.   Will summarize when the re-run finishes in a few hours.

eddelbuettel:
Results are in:  390 passed, 24 fail.  Of those 24 most are known cases of tests or examples 'failing' because packages (from Suggests:) are assumed present, but I only install Depends: and Imports:.  
I will summarise this as usual (later though) in the rcpp-logs repo.

So the "tl;dr" is that I think we can merge this PR.

jjallaire:
Let's be sure to also add the requisite items to NEWS.rd and ChangeLog

eddelbuettel:
Maybe @kevinushey wants to take a stab at that?  We can just do that in the master branch.

kevinushey:
Thanks for merging! I'll add relevant news items on master.

PR #324, opened 2015-07-13
import rcpp-scripts to Rcpp repo

kevinushey:
This PR brings the `rcpp-scripts` repository into `Rcpp`.

IMO it's preferable to keep Rcpp and its tools in one place; otherwise it's easy to 'lose' them or let them get out of sync.

@eddelbuettel, if you agree with bringing `rcpp-scripts` would you mind merging this? Otherwise feel free to close the PR.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -14,3 +14,4 @@ doxyfile
 .dir-locals.el
 .clang_format
 vignettes/getCurrentVersionsOfCitedPackages.R
+^internal/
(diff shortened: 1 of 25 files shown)


eddelbuettel:
I fear they _are_ out of sync.   How about if I as a first step bring `eddelbuettel/rcpp-scripts` into `RcppCore/rcpp-scripts` and we take it from there?   I don't think that this PR (which also drops history unless I overlook something) is the best way.   I could be convinced otherwise...

eddelbuettel:
Actually, rcpp-scripts *is* already part of the RcppCore repo.    I am still not convinced we want the scripts inside the Rcpp repo though ... so closing this now.

PR #329, opened 2015-07-21
initialize underlying SEXP when construct String from char*/std::string

thirdwing:
This should fix cccp bugs.

String class should be rewritten using Storage meta class.


eddelbuettel:
Running a rev.dep check. Tried cccp manually first; it does of course pass.   We're getting very close!

eddelbuettel:
Rev.dep ran fine, no new issues.  Merging -- thank you!

PR #330, opened 2015-07-22
add 'updateChangeLog' utility function

kevinushey:
This PR adds an (unexported) function to Rcpp, `updateChangeLog()`, which can be used to take the set of modified files (based on a `git status --porcelain` call) and insert a `ChangeLog` skeleton.

Primarily intended for those of us who might be working on Rcpp without Emacs at a particular point in time...

The user name field is either pulled from a devtools option if it exists, or (as a default) uses `Sys.getenv("USER")`. A similar scheme is used for email (falling back to `Sys.getenv("MAIL")`).

### DIFF
--- a/R/tools.R
+++ b/R/tools.R
@@ -48,3 +48,55 @@ asBuildPath <- function(path) {
     
     return(path)
 }
+
+updateChangeLog <- function(dir = getwd(), text = "") {
+    
+    clPath <- file.path(dir, "ChangeLog")
+    if (is.null(clPath))
+        return()
+    
+    if (!file.exists(clPath))
+        file.create(clPath)
+    
+    on.exit(file.edit(clPath))
+    
+    git <- Sys.which("git")
+    stopifnot(git != "")
+    
+    gitOutput <- system("git status --porcelain", intern = TRUE)
+    if (!length(gitOutput))
+        return()
+    
+    changedFiles <- sort(substr(gitOutput, 4, nchar(gitOutput)))
+    
+    userName <- getOption("devtools.name", default = Sys.getenv("USER"))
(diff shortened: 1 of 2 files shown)


R/tools.R
```diff
@@ -48,3 +48,55 @@ asBuildPath <- function(path) {
...
+    git <- Sys.which("git")
+    stopifnot(git != "")
+    
+    gitOutput <- system("git status --porcelain", intern = TRUE)
+    if (!length(gitOutput))
+        return()
+    
+    changedFiles <- sort(substr(gitOutput, 4, nchar(gitOutput)))
+    
+    userName <- getOption("devtools.name", default = Sys.getenv("USER"))
```
eddelbuettel: I have found Gabor Csardi's 'whoami' package excellent for this.  That said, we may not want to bring extra dependencies in.  Then again, maybe `updateChangeLog()` wants to be in its own package?

eddelbuettel:
Maybe the [rcpp-scripts](https://github.com/RcppCore/rcpp-scripts) repo is a better candidate?

kevinushey:
The main motivation for including this as part of `Rcpp` itself is that it also makes it much easier to encourage new contributors to update the ChangeLog as part of their edits, as the tooling built right into Rcpp itself (regardless of what editor they use).

The 'selfish' reason is that it makes it really easy for me to update the ChangeLog when editting Rcpp from RStudio; of course, that doesn't imply that this should live in Rcpp...

eddelbuettel:
I really thing this belongs into rcpp-scripts, or another-to-be-created utils package.  It is nothing an Rcpp _user_ would want.

kevinushey:
Fair enough!

Issue #309, opened 2015-06-23
sourceCpp changes working directory


kendonB:
I don't think this should be happening. I run something like:

    sourceCpp("C/foo.cpp")

and the new wd is ``"old_folder/C"``

Hope this helps!

eddelbuettel:
What version of Rcpp and what OS?   It does not happen here:

```r
R> library(Rcpp)
R> getwd()
[1] "/home/edd/Dropbox/src/R"
R> sourceCpp("/tmp/alexios.cpp")
R> getwd()
[1] "/home/edd/Dropbox/src/R"
R> 
```

and it is not a relative-vs-absolute path issue:

```r
R> setwd("~/git/src-latex/conferences/uzuerich_2015/")
R> getwd()
[1] "/home/edd/git/src-latex/conferences/uzuerich_2015"
R> sourceCpp("code/cumsum.cpp")
R> getwd()
[1] "/home/edd/git/src-latex/conferences/uzuerich_2015"
R> 
```

kendonB:
I've recently upgraded from Rcpp 0.11.3 to 0.11.5 and the problem seems to no longer occur.

gforge:
I'm not sure this should be closed - I'm currently experiencing the same issue when running on my Win 8.1 machine with Rcpp 0.12.0 and R 3.2.1

jjallaire:
I'm not able to reproduce this locally (Rcpp 0.12.0, and R 3.2.1 on both Win7 and OS X Mavericks). Could you provide a full reproducible example including the source file you are compiling, the exact call to sourceCpp, and the working directory before and after the compilation.

gforge:
Wow this was a little tricky once I digged a little deeper. It **only** appears after sourceCpp the same file **twice**!? I.e. the following code causes the error:

```R
library(Rcpp)
getwd()
sourceCpp("C:/Users/Max/Documents/R/test/cpp/test.cpp")
getwd()
sourceCpp("C:/Users/Max/Documents/R/test/cpp/test.cpp")
getwd()
```

Gives the following:

```R
> library(Rcpp)
> getwd()
[1] "C:/Users/Max/Documents/R/test"
> sourceCpp("C:/Users/Max/Documents/R/test/cpp/test.cpp")

> timesTwo(42)
[1] 84
> getwd()
[1] "C:/Users/Max/Documents/R/test"
> sourceCpp("C:/Users/Max/Documents/R/test/cpp/test.cpp")

> timesTwo(42)
[...]

PR #335, opened 2015-08-02
very simple fix for #309

eddelbuettel:
if no rebuild is needed, still set an on.exit() handler to get back to the
starting directory

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-08-02  Dirk Eddelbuettel  <edd@debian.org>
+
+        * R/Attributes.R (sourceCpp): In no-rebuild-needed case, also set a
+        default directory restoration via on.exit()
+
 2015-07-24  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION: Release 0.12.0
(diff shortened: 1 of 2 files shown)


jjallaire:
Let me take a closer look at this before we merge. From first glance it appears that we haven't yet done the setwd in the case of the `else` clause you put in however clearly it's happening somewhere (perhaps on the native code side?). Want to make sure I fully understand the source of the bug before we apply the fix.

eddelbuettel:
Sure. I instrumented a local copy with simply i/o to see which `setwd()` was being triggered (which I'll email you in a minute).  That made it pretty apparent that we lacked one in the 'no rebuild' case.   But by all means take a closer look; my analysis was clearly of the Sunday-morning variety :)

jjallaire:
The fix was as expected correct (I had missed that we also set the working directory as when sourcing embedded R code). I was puzzled why we might not have seen this more often in the field but it requires the combination of:

1) Sourcing a file from outside the current directory more then once; and
2) The presence of an embedded R code chunk in the file.

Certainly plausible that this doesn't happen all that often.

PR #338, opened 2015-08-12
Fix  querying  and creation of a large matrix due to integer overflow

fplaza:


eddelbuettel:
Thanks for the PR, that looks interesting.

Would you be able to supply an example where the old code break, and ideally where the new code passes?  Even better, could you add that as a unit test?

thirdwing:
I think `nrow` and `ncol` should be `int` to be consistent with R source code.

https://github.com/wch/r-source/blob/trunk/src/main/array.c#L202-#L213

fplaza:
Actually, I think that almost all occurences of 'int' and 'size_t' should be replaced by 'R_xlen_t' which is a typedef of 'ptr_diff_t' that "acts as the signed counterpart of std::size_t"

thirdwing:
In my opinion, if `nrow` and `ncol` are both `R_xlen_t` and large enough, the overall size of matrix, `nrow * ncol`, can still be too big to fit into a `R_xlen_t` type. 

So in R's own source code, `nrow` and `ncol` are `int` and the overall size is `R_xlen_t`.

https://github.com/wch/r-source/blob/trunk/src/main/array.c#L202-#L213

fplaza:
You are right.
ncol and nrow are "int" and bounded to 2^31-1 
However, some casts to "R_xlen_t" are required to avoid integer overflows while multiplying dimensions.

fplaza:
This code doesn't produce the expected result.
```c++
#include <Rcpp.h>
using namespace Rcpp;


// [[Rcpp::export]]
NumericMatrix test() {
  const int nrow = 4000000;
  const int ncol = 700;
  NumericMatrix v(nrow,ncol);
  return v;
}

/*** R
test()
*/
```
```
> test()
Error in .Primitive(".Call")(<pointer: 0x000000006f941db0>) : 
  negative length vectors are not allowed
````

**Comment by kevinushey** [comment]
I agree that changing from `int` to `size_t` isn't the right choice here; we should match the interface with (current) R's. But we do need to perform some casts to prevent overflow.

E.g.

https://github.com/wch/r-source/blob/trunk/src/main/array.c#L148-L151
https://github.com/wch/r-source/blob/trunk/src/main/array.c#L134

**Comment by eddelbuettel** [comment]
Quick check:  

```r
R> 4000000 * 700 * 8 / 1e9
[1] 22.4
R> 
```

22 gb is not all  that common, but not impossible.   And on board with the rest of the discussion: we ought to have the proper interface to R.

fplaza:
This code probably produces a segfault but I will not have an access to a machine with enought RAM to test it until tomorrow.
````c++

#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
void test(const NumericMatrix& m) {
  Rcout << m(m.nrow()-1,m.ncol()-1)  << std::endl;
}

/*** R
m=matrix(rep(1.1,4000000*700), nrow=4000000,ncol=700)
test(m)
*/
````

**Comment by fplaza** [comment]
For your information, a lot laboratories are using R to analyse matrices with much more than 2^31-1 elements

**Comment by thirdwing** [comment]
The problem is in the dimension class: https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/Dimension.h#L59-L61

`nrow` and `ncol` should be `int`, and their production should be a `R_xlen_t`.

**Comment by eddelbuettel** [comment]
Nicely done, KK.  Agree with the issue in Dimension.h.

And yes, the code does segfault:

```r
R> cppFunction("void mattest(NumericMatrix m) { Rcout <<  m(m.nrow()-1,m.ncol()-1)  << std::endl; }")  
R> m <- matrix(1.1, nrow=4000000,ncol=700)    
R> mattest(m)

 *** caught segfault ***
address 0x7f96c8671040, cause 'memory not mapped'

Traceback:
 1: .Primitive(".Call")(<pointer: 0x7f9ec8675c40>, m)
 2: mattest(m)

Possible actions:
1: abort (with core dump, if enabled)
2: normal R exit
3: exit R without saving workspace
4: exit R saving workspace
Selection: 
```

fplaza:
This constructor is also problematic
```c++
    template <typename Iterator>
    Matrix( const int& nrows_, const int& ncols, Iterator start ) :
        VECTOR( start, start + (nrows_*ncols) ), // Potential overflow. Need to cast to R_xlen_t
        nrows(nrows_)
    {
        VECTOR::attr( "dim" ) = Dimension( nrows, ncols ) ;
    }
```

thirdwing:
So can you update your PR according our discussion? @fplaza

fplaza:
What about those operators?
Should 'size_t' be replaced  by int?
```c++
    inline Proxy operator()( const size_t& i, const size_t& j) {
      return static_cast< Vector<RTYPE>* >( this )->operator[]( offset( i, j ) ) ;
    }
    inline const_Proxy operator()( const size_t& i, const size_t& j) const {
       return static_cast< const Vector<RTYPE>* >( this )->operator[]( offset( i, j ) ) ;
    }
````

**Comment by thirdwing** [comment]
They should be `int` and actually are cast to `int` by `offset`.

https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Matrix.h#L160

**Comment by fplaza** [comment]
I am not very familiar with git.
Could you briefly indicate me how to update my PR?

**Comment by eddelbuettel** [comment]
Change your file. Commit it again.

File: inst/include/Rcpp/vector/Matrix.h [hunk 1]
```diff
@@ -58,7 +58,7 @@ class Matrix : public Vector<RTYPE, StoragePolicy>, public MatrixBase<RTYPE, tru
 
     template <typename Iterator>
     Matrix( const int& nrows_, const int& ncols, Iterator start ) :
-        VECTOR( start, start + (nrows_*ncols) ),
+        VECTOR( start, start + (static_cast<R_xlen_t>(nrows)_*ncols) ),
```
thirdwing: You have a typo here, `static_cast<R_xlen_t>(nrows_)`

inst/include/Rcpp/vector/Matrix.h
```diff
@@ -58,7 +58,7 @@ class Matrix : public Vector<RTYPE, StoragePolicy>, public MatrixBase<RTYPE, tru
 
     template <typename Iterator>
     Matrix( const int& nrows_, const int& ncols, Iterator start ) :
-        VECTOR( start, start + (nrows_*ncols) ),
+        VECTOR( start, start + (static_cast<R_xlen_t>(nrows_)*ncols) ),
```
eddelbuettel: Or rather   `VECTOR( start, start + (static_cast<R_xlen_t>(nrows_ * ncols)))` ?

thirdwing: I think we need to cast before multiply, otherwise, there is still an overflow, right?

eddelbuettel: Ahh.  Good point.

fplaza:
Do the changes satisfy you?
If so, when do you plan to publish a new version of Rcpp which includes them?
Regards.

inst/include/Rcpp/Dimension.h
```diff
@@ -39,14 +39,14 @@ namespace Rcpp{
 	            dims = other.dims ;
 	        return *this ;
 	    }
-	    Dimension(const size_t& n1) : dims(1){
+	    Dimension(const int& n1) : dims(1){
```
eddelbuettel: But `int` is signed and `size_t` is not. I think you just reduced the range of numbers which can be expressed here.

eddelbuettel:
I think it needs more discussion.  

`ChangeLog` and `NEWS.Rd` will give you a good idea about our release frequency.

inst/include/Rcpp/Dimension.h
fplaza: Actually, dims is an 'int' vector so 'size_t' will be downcasted to 'int' at the end

eddelbuettel: We are not that lighthearted about changing existing interfaces.  Right now I am inclined not to take the patch.   Lines 59 and 60 in `Dimension.h` look fine as does line 61 in `Matrix.h`.  I am not so sure about the rest.

fplaza: The patches make the Rcpp interface coherent with the RInternals (which is currently not the case.)
Tell me what are your concerns and I will answer it.

kevinushey: For coherence with R's interface, I think it's best to use `int` here. The `std::size_t` items used would still get casted to `int` later which of course would cause problems if we had a `std::size_t` which did not fit in `int`.

fplaza:
I am closing this PR.
I will create an issue to create a consensus about the changes needed.

PR #340, opened 2015-08-14
Add bounds checking in Matrix operator()

fplaza:
This patch makes the Matrix interface coherent with the Vector interface.

### DIFF
--- a/inst/include/Rcpp/vector/Matrix.h
+++ b/inst/include/Rcpp/vector/Matrix.h
@@ -129,10 +129,10 @@ public:
     }
 
     inline Proxy operator()( const size_t& i, const size_t& j) {
-      return static_cast< Vector<RTYPE>* >( this )->operator[]( offset( i, j ) ) ;
+      return static_cast< Vector<RTYPE>* >( this )->operator()(i, j) ;
     }
     inline const_Proxy operator()( const size_t& i, const size_t& j) const {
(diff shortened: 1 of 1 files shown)


eddelbuettel:
May I suggest that we adopt a different pattern now that you seem to request our semi-permanent attention?

It is hard to gather context from a two-line diff, and the PR has precisely **zero** motivation or rationale.  Let's not do this.

Start with issue ticket.  Convince us that something needs fixing ie establish consensus about whether something needs change.   Only then consider code change submissions.

fplaza:
Fine, I don't want to bother you.
I will proceed as you mention next time.

Here is a small example
```c++
#include <Rcpp.h>
using namespace Rcpp;


// [[Rcpp::export]]
void test_vec()
{
  
  const int length = 2;
  NumericVector v1(length);
  //This fails
  Rcout << v1(length) << std::endl;
}

// [[Rcpp::export]]
void test_mat()
{
  const int num_rows = 2;
  const int num_cols = 2;
  NumericMatrix m1(num_rows, num_cols);
  //This should fail tooo
  Rcout << m1(2,2) << std::endl;
}




/*** R
test_vec()
test_mat()
*/
```
Result
```
> test_vec()
Error in eval(expr, envir, enclos) : index out of bounds
> test_mat()
1.58802e-315
````

PR #344, opened 2015-08-15
Fix conversion from char to string

fplaza:


eddelbuettel:
Ah. Yes. Good point.  To merge it I'd need a PR first :)

Issue #345, opened 2015-08-15
Enable e.g. `x[...] = y[...]`


kevinushey:
Re: http://stackoverflow.com/questions/31992883/in-rcpp-why-cant-i-assign-a-subsetted-vector-to-another-subsetted-vector

It's possible that this could be implemented with an appropriate `operator=` taking a `SubsetProxy` object: https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Subsetter.h

fplaza:
Adding this operator solves the issue
```c++
	SubsetProxy& operator=(const SubsetProxy& other) {
        if (indices_n != other.indices_n) stop("index error");
		for (int i=0; i < indices_n; ++i)
			lhs[ indices[i] ] = other.lhs[other.indices[i]];
        return *this;
    }
```

fplaza:
By the way, I think that
```c++
    template <int OtherRTYPE, template <class> class OtherStoragePolicy>
    SubsetProxy& operator=(const Vector<OtherRTYPE, OtherStoragePolicy>& other) {
        int n = other.size();
        if (indices_n != n) stop("index error");
        if (n == 1) {
            for (int i=0; i < n; ++i) {
                lhs[ indices[i] ] = other[0];
            }
        } else if (n == indices_n) {
            for (int i=0; i < n; ++i) {
                lhs[ indices[i] ] = other[i];
            }
        } else {
            stop("index error");
        }
        return *this;
    }
```
could be replaced by
```c++
template <int OtherRTYPE, template <class> class OtherStoragePolicy>
    SubsetProxy& operator=(const Vector<OtherRTYPE, OtherStoragePolicy>& other) {
        int n = other.size();
        if (indices_n != n) stop("index error");
		for (int i=0; i < n; ++i) {
			lhs[ indices[i] ] = other[i];
		}
        return *this;
    }
```

fplaza:
And the error message from "index error" to "number of items to replace is not a multiple of replacement length"

fplaza:
However, this code doesn't produce the expected result
```c++
#include <Rcpp.h>
using namespace Rcpp;


// [[Rcpp::export]]
NumericVector valuesOverThree(NumericVector b){
  
  NumericVector a(b.size());
  
  NumericVector temp = b[b > 3];
  a[b <= 3] = temp;
  return a;
}

// [[Rcpp::export]]
NumericVector valuesOverThree2(NumericVector b){
[...]

PR #349, opened 2015-08-25
Allow assignment of size 1 with Subsetter

fplaza:
R allows singleton vectors to be assigned to vectors subsets with more than one element.

Here is an example:
```R
> v1=c(1,2,3)
> v2=c(4)
> v1[v1<4]=v2
> v1
[1] 4 4 4
```

This feature is currently not available in Rcpp as previously discussed in #345 

Code:
```c++
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
IntegerVector test()
{
  IntegerVector v1 = IntegerVector::create(1,2,3);
  IntegerVector v2 = IntegerVector::create(4);
  v1[v1<4] = v2;
  
  return v1;
}

/*** R
test()
*/
```

Result:
```
test()
Error in eval(expr, envir, enclos) : index error
```

This PR fixes this issue by returning the same result as R:
```
test()
[1] 4 4 4
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2015-09-01  Florian Plaza Onate <florian.plaza@jouy.inra.fr>
+        * inst/include/Rcpp/vector/Subsetter.h: Allow logical subsets to be
+       	assigned to a vector of size 1 
+		(pull request #349, discussed in issue #345)
+        * inst/unitTests/cpp/Subset.cpp: Add unit test for above
+        * inst/unitTests/runit.subset.R: Ditto
+
 2015-08-31  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll Version and Date
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 2015-09-01  Florian Plaza Onate <florian.plaza@jouy.inra.fr>
         * inst/include/Rcpp/vector/Subsetter.h: Allow logical subsets to be
-       	assigned to a vector of size 1 
-		(pull request #349, discussed in issue #345)
+        assigned to a vector of size 1 
+        (pull request #349, discussed in issue #345)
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Florian, can we please try it the other way around as previously discussed:
- identify an issue
- create consensus that a code fix is needed
- work out a code fix
- create a pull request
- ensure the pull request is complete with
    - unit tests
    - documentation
    - ChangeLog entry

This one once again start from the wrong (though I agree that it is of course a very small one...).

fplaza:
We briefly discussed this issue here https://github.com/RcppCore/Rcpp/issues/345 but I will create a new one

eddelbuettel:
We can keep this open, it looks like a valid patch.   I just do _not_ like "random" pull requests that fly in without prior discussion.  In this case, if #345 is/was indeed the place where this was discussed, then the fault is on me for not remembering -- and you for not referencing it.  I have to deal with more than one discussion at a time, so helpful pointers are appreciated.

In any event, documentation, tests, changelog entries etc pp are also clearly missing from the PR.  So please reopen and augment as you see fit.

fplaza:
Could you have a look to my last commits?

eddelbuettel:
The patch is fine as far as it mechanics go.  But I can't make up my mind on this:
- on the hand it is nice to have a new subsetter
- but on the other hand it is a very specific case of recycling
    - what about vectors other than length one?  
    - shouldn't an element other than position 0 be assignable?
   
Recycling has very specific meaning in R.  We have avoided it here, mostly, as it is not natural for C++.

I am not yet convinced this really helps.

@kevinushey @jjallaire

kevinushey:
Looks good to me! The PR fixes a bug in the old implementation (the goal was to allow recycling in the first place; it just was not implemented correctly)

eddelbuettel:
Thanks for that.

eddelbuettel:
Ok, with this and the parallel email discussion this is good to go.

Issue #350, opened 2015-08-26
Recent R changes and multiple Modules with the same name


eddelbuettel:
We can cause a segmentation violation in R 3.2.2 and R-devel simple by running

```
runit.Module.client.package.R
runit.modref.R
```

after one another in a script.   There seems to be some interaction with commit r68802 in R itself but after discussion with R Core the best cause of action seems to be to sidestep the issue by not having several test Modules in Rcpp which all define a Module named `World`.

eddelbuettel:
closed by #351

PR #351, opened 2015-08-26
address issue #350 by renaming Modules

eddelbuettel:
affects four different unit tests
also very minor cleanups

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,21 @@
+2015-08-26  Dirk Eddelbuettel  <edd@debian.org>
+
+        * inst/unitTests/testRcppClass/src/rcpp_module.cpp: Renamed Module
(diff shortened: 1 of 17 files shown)


eddelbuettel:
Interesting.  So it fails at Travis while it passes here at on my box.   I'll separate out more modules then -- now that `World` is taken care of this crashed over `Num`.

eddelbuettel:
@kevinushey @jjallaire Let me know what you think -- this ought to be ready to merge.

kevinushey:
Looks good to me.

eddelbuettel:
Rev.deps finished fine, logs and summary committed in https://github.com/RcppCore/rcpp-logs/commit/ebbce0c3a6e6d4af0057b24103ba46e185bb3bd5 

No new issues that were caused by Modules; there is one package which fails it tests for an odd reason I couldn;t immediately attribute.

PR #352, opened 2015-08-28
Enable e.g. 'a[b > 3] = b[b > 3]'

thirdwing:
This for #345 .

It is just a combination of code from @kevinushey and @fabian-s .

I think converting to `Vector` is better than defining `length()` and `get()` for `SubsetProxy`.

### DIFF
--- a/inst/include/Rcpp/vector/Subsetter.h
+++ b/inst/include/Rcpp/vector/Subsetter.h
@@ -100,8 +100,24 @@ public:
     template <int RTYPE_OTHER, template <class> class StoragePolicyOther,int RHS_RTYPE_OTHER, bool RHS_NA_OTHER, typename RHS_T_OTHER>
     SubsetProxy& operator=(const SubsetProxy<RTYPE_OTHER, StoragePolicyOther, RHS_RTYPE_OTHER, RHS_NA_OTHER, RHS_T_OTHER>& other) {
 
-        Vector<RTYPE, StoragePolicyOther> other_vec = other;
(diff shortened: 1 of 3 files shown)


kevinushey:
Thanks!

What happens when we attempt to assign non-conforming types, e.g. if `x` is a `NumericVector` and `y` is an `IntegerVector`? We should have test cases that ensure these throw or convert properly (seems like it could be hairy)

thirdwing:
```
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
IntegerVector subset_assign_subset(NumericVector x) {
    IntegerVector y(x.size());
    y[x > 3] = x[x > 3];
    return y;
}

/*

> Rcpp::sourceCpp("test.cpp")
> subset_assign_subset(1:6)
[1] 0 0 0 4 5 6
> x <- c(1.2, 2.3, 3.4, 4.5)
> subset_assign_subset(x)
[1] 0 0 3 4

*/
```

kevinushey:
It looks like type coercion is already properly handled so I think this is good to go. E.g. this gives a compiler error, as hoped:

```
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
IntegerVector subset_assign_subset(CharacterVector x) {
  IntegerVector y(x.size());
  y[x > 3] = x[x > 3];
  return y;
}
```

eddelbuettel:
Having a few unit tests would be nice.

eddelbuettel:
Appreciate the extra tests!   Merging now.

PR #353, opened 2015-08-29
Feature/module load

eddelbuettel:
This contains code emailed in by Kurt Hornik to catch up with changed on the R side.

It has been tested against reverse dependencies on both r-release and r-devel and it appears that it is  not creating any new issues.  There are one or two breakages we have not seen before but they may be tickled by something else as well.

### DIFF
--- a/R/loadModule.R
+++ b/R/loadModule.R
@@ -67,7 +67,9 @@ loadModule <- function( module, what = character(), loadNow,
         loadNow <- !is(loadM, "error")
     }
     if(loadNow) {
-        .botched <- isBotchedSession()
+        ## .botched <- isBotchedSession()
+        .botched <- FALSE
+
         if(is.null(loadM))
             loadM <- tryCatch(Module( module, mustStart = TRUE, where = env ),
                               error = function(e)e)
--- a/R/loadModule.R
+++ b/R/loadModule.R
@@ -1,4 +1,4 @@
-# Copyright (C) 2010 - 2012 John Chambers, Dirk Eddelbuettel and Romain Francois
+# Copyright (C) 2010 - 2015  John Chambers, Dirk Eddelbuettel and Romain Francois
 #
 # This file is part of Rcpp.
 #
--- a/inst/include/Rcpp/Reference.h
+++ b/inst/include/Rcpp/Reference.h
@@ -59,7 +59,8 @@ namespace Rcpp{
         Reference_Impl( const std::string& klass ) {
             SEXP newSym = Rf_install("new");
             Shield<SEXP> call( Rf_lang2( newSym, Rf_mkString( klass.c_str() ) ) );
-            Storage::set__( Rcpp_eval( call ) ) ;
+            //Storage::set__( Rcpp_eval( call ) ) ;
+            Storage::set__( Rcpp_eval( call , Rcpp::internal::get_Rcpp_namespace()) ) ;
         }
 
         void update( SEXP x){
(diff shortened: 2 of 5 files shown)


jjallaire:
Looks good to me. @kevinushey ?

eddelbuettel:
There is still some headscratchery in other placed.  Eg `R CMD check` fails on the tests in RcppDE under R 3.2.2 if and only if the last replication size (50) is used.  It does not fail when gdb is used.  It also does not fail under R-devel.  In short, some ugly race condition somewhere, possibly in R.  I also have one or two packages now puke in their unit tests so it is not clear this is us.

We need a better tool chain to run (and report) these tests.  Is Gabor done yet?  ;-)

R/loadModule.R
```diff
@@ -67,7 +67,9 @@ loadModule <- function( module, what = character(), loadNow,
         loadNow <- !is(loadM, "error")
     }
     if(loadNow) {
-        .botched <- isBotchedSession()
+        ## .botched <- isBotchedSession()
+        .botched <- FALSE
```
kevinushey: Do we know why we had this check?

The comment above `isBotchedSession()`:

```
## the following items are to get around some insanity in the
## CMD check of packages using Rcpp that dies in loadModule()
## because some code somewhere can't find the methods package
isBotchedSession <- function()
    ! ("package:methods" %in% search())
```

This, at least to me, implies that there is code within Rcpp modules that requires `methods` not only to be loaded but also attached; do we have some way of verifying that Rcpp modules work when methods is loaded but not attached?

inst/include/Rcpp/Reference.h
```diff
@@ -59,15 +59,15 @@ namespace Rcpp{
         Reference_Impl( const std::string& klass ) {
             SEXP newSym = Rf_install("new");
             Shield<SEXP> call( Rf_lang2( newSym, Rf_mkString( klass.c_str() ) ) );
-            Storage::set__( Rcpp_eval( call ) ) ;
+            Storage::set__( Rcpp_eval( call , Rcpp::internal::get_Rcpp_namespace()) );
```
kevinushey: I guess the intention of this is to ensure that `new` is picked up from the `methods` namespace (which `Rcpp` exports)?

I would prefer something more explicit here, but I guess that's fine.

R/loadModule.R
eddelbuettel: That is why I brought this to the `rcpp-core` list and John -- AFAIK he added it.

inst/include/Rcpp/Reference.h
eddelbuettel: Ditto here.   If you can think of better alternatives I'd be all ears :)   I guess worked Kurt worked this out "the hard way" as recent changes in R-devel required it.

In his email to me, he stressed the use of `R_DEFAULT_PACKES=NULL` with `debug(Rcpp::loadModule)` which lead RcppBDT loading to fail -- via either  `loadNamespace()` or `library()` leading to this `Module( module, mustStart = TRUE, where = env )` failing for
lack of `new`.

Personally, I find this whole dance around the methods package mysterious. If it were me, I'd just depend on it...

kevinushey: I'm guessing we avoid depending on `methods` because we only need it when modules are used, and not all Rcpp users require modules?

Either way, I think this is fine.

eddelbuettel: (I was sort-of-kinda-didn't-quite-say-it complaining about other `methods` cases too; for example nobody understands the nonsense of `Rscript` not calling `methods` by default leading to a bunch of scripts doing funny stuff ...   As I said, I'd just load the bugger -- which is just [what littler does](https://github.com/eddelbuettel/littler/blob/master/autoloads.R#L1).)

eddelbuettel: @kevinushey How about if I merge that and run another rev.dep. check just in case?

PR #357, opened 2015-08-29
refreshed tinyformat.h against upstream, keeping local mods (closes #356)

### LINKED ISSUE #356: tinyformat refresher
It would appear that [tinyformat](https://github.com/c42f/tinyformat/blob/master/tinyformat.h)  got updated so we may want to take a look.   Judging from `git blame` it is mostly one import followed by several surgical operations / insertions.  Sounds about right to you, @kevinushey ?

eddelbuettel:


kevinushey:
Looks good to me, thanks!

eddelbuettel:
Yes, should not bite.  Will merge now.

PR #368, opened 2015-09-07
Nullable object (closes #363)

### LINKED ISSUE #363: Nullable object
In quite some user-space code the R side has something like

```{.r}
weightedLeastSquares(X, y, w)      # ie y ~ X given weights w
```

where the third argument is options, ie 

```{.r}
weightedLeastSquares(X, y, NULL)
```

In that case we generally cannot have the third strongly typed as `NumericMatrix` or alike and have to keep it a `SEXP` in the interface.  Nothing wrong with that but a bit of an eyesore. 
  
Following some discussion with @dcdillon I have written a really simple wrapper class `Nullable` that contains a `SEXP` as well as a `bool` for state.  With that we can do (stylized)

```{.cpp}
...., NumericVector & w) {
//...
if (!w.isNull()) {
    // do stuff with w
}
```

If there is no protest -- or someone reminds of something much better we already have ;-) -- then I'll fold that in.

eddelbuettel:
This (simple) PR has been tested in pending code that will may go into [mvabund](https://cran.rstudio.com/package=mvabund) in a few days.  (I may use a copy there to not have to rely on the very newest Rcpp there.)  But as the feature is generally useful (and I also already spotted use cases for it in [RcppDE](https://cran.rstudio.com/package=RcppDE)).

The basic need was described in #363, and we had a lively discussion there.  @dcdillon and went back and forth over some design with a templated class but this doesn't quite work:  the is circular as we need a `NULL` protection layer around our proxy objects.  As soon as we template with them, access when `NULL` leads to an instantiation which ... throws as before.   Similarly, I got nowhere with the `safe_bool_idiom` approach.

So what is here works, but (as always) could be improved if someone is so inclined.  I added a few tests to describe how it is used.

### DIFF
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -68,7 +68,7 @@ namespace Rcpp {
          * @throw 'not initialized' if object has not been set
          */
         inline operator SEXP() {
-            checkIfSet();
+            isSet();
             return m_sexp;
         }
 
@@ -83,7 +83,7 @@ namespace Rcpp {
         }
 
         /**
-         * boolean test for NULL
+         * Boolean test for NULL
          *
          * @throw 'not initialized' if object has not been set
          */
@@ -93,13 +93,19 @@ namespace Rcpp {
         }
 
         /**
-         * boolean test for not NULL
+         * Boolean test for not NULL
          *
          * @throw 'not initialized' if object has not been set
          */
         inline bool isNotNull() {
             return ! isNull();
         }
+
+        /**
+         * Test function to check if object has been initialized
+         *
+         */
+        inline bool isSet(void) { return m_set; }
 
     private:
         SEXP m_sexp;
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -2,7 +2,7 @@
 //
 // Nullable.h: Rcpp R/C++ interface class library -- SEXP container which can be NULL
 //
-// Copyright (C) 2015         Dirk Eddelbuettel
+// Copyright (C) 2015         Dirk Eddelbuettel and Daniel C. Dillon
 //
 // This file is part of Rcpp.
 //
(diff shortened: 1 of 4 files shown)


dcdillon:
Given that `isNull()` throws if it has never been set, should you also add an `isSet()` function?  This would give the user the ability to poll for the "third state" without resorting to exception handling.

eddelbuettel:
You could you be "not set" but "not null" ?

inst/include/Rcpp/Nullable.h
```diff
@@ -0,0 +1,122 @@
...
+            return *this;
+        }
+
+        /**
+         * operator SEXP() to return nullable object
+         *
+         * @throw 'not initialized' if object has not been set
+         */
+        inline operator SEXP() {
+            isSet();
```
kevinushey: Should this be `checkIfSet()`?

kevinushey:
My only thought: the class name `Nullable` does not really evoke the type of the underlying object; e.g. I would much rather see `Nullable<SEXP>` to indicate that this is an object wrapping a `SEXP` that may be either unset or `R_NilValue`. It would be even better if this were a template class, e.g. `Nullable<T>` where `T` could be any Rcpp class type.

I'm still not quite sure on the utility -- can you provide some other examples where this would be used? (The example in #363 was a bit bare bones; a full example would be useful)

inst/include/Rcpp/Nullable.h
eddelbuettel: Absolutely. Now fixed.  

I blame @dcdillon for enticing me to make changes, which I then changed midflight...  ;-)

inst/include/Rcpp/Nullable.h
```diff
@@ -0,0 +1,122 @@
...
+         */
+        inline bool isNotNull() {
+            return ! isNull();
+        }
+
+        /**
+         * Test function to check if object has been initialized
+         *
+         */
+        inline bool isSet(void) { return m_set; }
```
kevinushey: nit: probably don't need `void` in the parameter list

eddelbuettel:
`Nullable<SEXP>` could work whereas `Nullable<RcppANYTHING>` does not seem to work given how our types are set up -- they bark at run-time when you inquire about their state and the payload was `NULL`.

One use case in in [this PR](https://github.com/aliceyiwang/mvabund/pull/4) where I carried a local copy over.  We can now do [this in compact way](https://github.com/aliceyiwang/mvabund/blob/feature/nullability/src/Rinterface.cpp#L151-L155) whereas the [current code](https://github.com/aliceyiwang/mvabund/blob/master/src/Rinterface.cpp#L156-L164) is much uglier -- and exposes a `SEXP` in the nice Rcpp Attributes interface.

Suggestions for finer tuning welcome.  What is there now is so far good enough for me.

kevinushey:
Okay -- LGTM!

eddelbuettel:
Thanks.  @dcdillon and I went back and forth on this.  We both really wanted `Nullable<T>` but you just instrument it as `Nullable<NumericMatrix>` and then pass `NULL` at run-time. It gets you back to `NULL` being used to instantiate `NumericMatrix`  :-/  

Maybe it can still be done -- suggestions welcome.  We now have a few units describing what this should be good for, and one test use over in [RcppGSL] -- and those may help provide a framework for something better.   But for now I'd say let's put this one too.

eddelbuettel:
Ok, being stubborn helps sometimes.  

We had another crack at this and @dcdillon spotted a really bad `NULL` that really wanted to be a `R_NilValue` plus a few more little things -- so we WILL have a templated version coming.

PR #373, opened 2015-09-15
fixed compiler warning on icpc

wrathematics:


eddelbuettel:
Thanks for the PR. Unfortunately it came without a preceding issue ticket which could have clarified  what error you're seeing from what compiler.

Also, [upstream R-devel still uses the exact same code](https://github.com/wch/r-source/blob/trunk/src/extra/tzone/localtime.c#L382) which makes me lean against the PR. Why would I want to differ from the widely used R code?

wrathematics:
Yes sorry, I should have added more information.  Using Intel composer XE 2015, the following warning is generated:

```
Date.cpp(562): warning #437: reference to local variable of enclosing function is not allowed
  		      2 * sizeof *sp + 4 * TZ_MAX_TIMES];
  		                  ^
```

I'm not much of a C++ guy, but it looks to me like one of those things that's not allowed in C++ but valid in C.  As to whether or not you want to change it, if you aren't getting strange behavior then maybe there's no good reason not to.  Either way, no biggie, and sorry for the bother.

eddelbuettel:
I will test it in g++ and clang++ and reconsider if they bark.  If they don't then I may not be inclined to make a change.

eddelbuettel:
As the current versions of the two compilers families used in the vast majorities of deployments do not warn, and as the code still stands as is in R itself, I will close this.  

I appreciate you bringing this though.  I take warnings seriously though, but I also trust R Core, g++ and clang++.   

Below is cut&paste copy from trying this in a current Docker container with g++-5.2 and clang++-3.7.

```
root@722d2dda8a30:/mnt/src# clang++-3.7 -c -Wall -pedantic Date.cpp -I../inst/include -I/usr/share/R/include/
root@722d2dda8a30:/mnt/src# g++-5 -c -Wall -pedantic Date.cpp -I../inst/include -I/usr/share/R/include/
root@722d2dda8a30:/mnt/src# 
```

wrathematics:
No worries; makes perfect sense.  Sorry for the trouble.

PR #374, opened 2015-09-18
fix wrong return type in product of dimensions

fplaza:
Sorry guys...
I should have tested it first.


thirdwing:
It doesn't matter. 

Very few people are using that large vector or matrix. It will be appreciated it you can test more.

fplaza:
I am working on an R package which processes very large matrices.
Everything works fine with this patch.

thirdwing:
@eddelbuettel Can you look at this?

eddelbuettel:
@thirdwing I sort-of had.  I think it is good as is.

kevinushey:
Looks good to me.

Issue #380, opened 2015-10-08
is__module_object leads to error when checking non-module object


tqchen:
```c++
#include <Rcpp.h>

using namespace Rcpp;

class MyClass{
};

RCPP_EXPORT_CALSS(MyClass)

// [[Rcpp::export]]
bool check(RObject m)
{
    return is<MyClass>(m);
}

RCPP_MODULE(test) {
     class_<MyClass>("MyClass");
}
```
If we call check from R's side by passing in a non-module object. It will report error "External pointer is expected", instead of returning false. 

The problem is here https://github.com/RcppCore/Rcpp/blob/0a43747be6c7b88fec6717a45034e2749012dddf/inst/include/Rcpp/api/meat/is.h#L153

Maybe need a explicit check of ext ptr type first before converting to XPtr

thirdwing:
`if (TYPEOF(m) == EXTPTRSXP)`

tqchen:
Directly check won't work, since that object is a S4, and I guess we need to check type of env.get(".cppclass")

thirdwing:
See the code below, I think it is what you want.

```
#include <Rcpp.h>                                                               
                                                                                 
// [[Rcpp::export]]                                                                                                 
void check(Rcpp::RObject m) {                                                   
    SEXP attrs = ATTRIB(m);                                                     
    Rcpp::List lst(attrs);                                                      
    std::vector<std::string> names = lst.names();                               
    for (size_t i = 0; i < names.size(); i++) {                                 
        Rcpp::Rcout << names[i] << std::endl;                                   
    }                                                                           
                                                                                
    Rcpp::List lst2 = Rcpp::List(lst[1]);                                       
                                                                                
    for (size_t i = 0; i < lst2.size(); i++) {  
        Rcpp::Rcout << Rcpp::as<std::string>(lst2[i]) << std::endl;   
    }                                                                           
}       
```

```
> x = as.array(c(1,2,3))
> mat = mx.nd.array(x, mx.cpu(0))
> check(mat)
.xData
class
Rcpp_MXNDArray
```

tqchen:
Yap :) However, I think this may be something that should be solved by ```Rcpp::is<NDArray>(mat)```
in the future

eddelbuettel:
But 

```{sh}
edd@don:~/git/rcpp(master)$ ag NDArray
edd@don:~/git/rcpp(master)$ 
```

[...]

PR #381, opened 2015-10-09
Enhance is check for module object

tqchen:
fix to #380

### DIFF
--- a/inst/include/Rcpp/api/meat/is.h
+++ b/inst/include/Rcpp/api/meat/is.h
@@ -150,10 +150,13 @@ namespace Rcpp {
 
     inline bool is_module_object_internal(SEXP obj, const char* clazz){
         Environment env(obj);
-        XPtr<class_Base> xp( env.get(".cppclass"));
+        SEXP sexp = env.get(".cppclass");
+        if (TYPEOF(sexp) != EXTPTRSXP) return false;
+        XPtr<class_Base> xp(sexp);
         return xp->has_typeinfo_name(clazz);
     }
     template <typename T> bool is__module__object(SEXP x) {
+        if (!is__simple<S4>(x)) return false;
         typedef typename Rcpp::traits::un_pointer<T>::type CLASS;
         return is_module_object_internal(x, typeid(CLASS).name());
     }


eddelbuettel:
That probably requires a reverse-dependency check but I am traveling right now and may not get to it for some time.

tqchen:
please take your time. we have applied a local specialization path to our package for now. Thanks

thirdwing:
BTW, maybe you can add a simple test in https://github.com/RcppCore/Rcpp/tree/master/inst/unitTests/testRcppModule/src

eddelbuettel:
+1 for adding unit test support to this (or any, really) pull request.

kevinushey:
This does not seem like the right change. I think the only reason you're getting punted into the module method is because you haven't let Rcpp understand how to handle `as` / `wrap`.

Please read https://cran.r-project.org/web/packages/Rcpp/vignettes/Rcpp-extending.pdf for more information on how you can let `Rcpp::as<T>` and `Rcpp::wrap` understand how to interact with custom classes; I am guessing your problem will go away if you do this.

thirdwing:
A very simple example to make things clear. Just use our `RcppModuleWorld`.

`check_module` is nothing but `Rcpp::is<RcppModuleWorld>(x)`.

```
> m <- new(RcppModuleWorld)
> m
C++ object <0x2858180> of class 'RcppModuleWorld' <0x306a480>
> check_module(m)
[1] TRUE
> check_module(1)
Error: expecting an external pointer
```

If other things pass into this function, we will get a warning, since `env.get(".cppclass")` return a `NULL`.

@tqchen thought we need to check whether it is `NULL`. If it is `NULL`, we should return `FALSE`, not a warning.

tqchen:
I think @thirdwing explains my point. Ideally it would be great ```Rcpp::is<RcppModuleWorld>```   only return true or false. So we can use it to check parameters. 

For example, we can use this to check if input parameter is a certain type, and give more friendly error message to the user.

```c++
void dosomething(Rcpp::List param_list) {
  for (size_t i = 0; i < param_list.size(); ++i) {
    if (!Rcpp::is<RcppModuleClassA>(param_list[i]))
        throw Rcpp::exception("The input parameter is expected to be list of RcppModuleClassA")
  }
}
```

kevinushey:
Thanks, it indeed looks like the `RCPP_EXPOSED_CLASS` is responsible for exposing `as` / `wrap` methods, hence where the dispatch is coming from. The PR looks fine to me, then.

eddelbuettel:
Agreed -- will merge now and launch a full reverse-dependency test.  We can always revert should something bubble up (which we have no reason to expect).

PR #393, opened 2015-11-07
incorporate @thirdwing's suggestion for sugar ops (closes #392)

### LINKED ISSUE #392: Sugar operator+ and SubsetProxy
The following fails to compile (x86_64-redhat-linux, g++ 4.8.3, R 3.2.2),  
```c++
#include <Rcpp.h>
// [[Rcpp::export]]
Rcpp::NumericVector SubsetAdd(Rcpp::NumericVector lhs, Rcpp::NumericVector rhs) {
    Rcpp::LogicalVector idx = !Rcpp::is_na(lhs) & !Rcpp::is_na(rhs);
    return lhs[idx] + rhs[idx];        // error
} 
``` 
with error 
``` 
no match for ‘operator+’ (operand types are 
‘Rcpp::SubsetProxy<14, Rcpp::PreserveStorage, 10, true, Rcpp::Vector<10, Rcpp::PreserveStorage> >’ 
and 
‘Rcpp::SubsetProxy<14, Rcpp::PreserveStorage, 10, true, Rcpp::Vector<10, Rcpp::PreserveStorage> >’)
```
=====

This compiles and evaluates correctly: 
```c++
Rcpp::NumericVector SubsetAdd(Rcpp::NumericVector lhs, Rcpp::NumericVector rhs) {
    Rcpp::LogicalVector idx = !Rcpp::is_na(lhs) & !Rcpp::is_na(rhs);
    
    return static_cast<const Rcpp::NumericVector&>(lhs[idx]) + 
        static_cast<const Rcpp::NumericVector&>(rhs[idx]);           // Ok
}
``` 
but surely is not ideal. I was unsuccessful in coming up with a fix for this, but hopefully someone can point me in the direction of a clever solution? (Presumably this functionality is desirable & possible.)

kevinushey:
Enables e.g. `x[y]` + `x[z]`.

Should we add other binary operators to the list? (I just grabbed the four main arithmetic ones)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ src/*.o
 src/*.so
 src/*.dll
 .Rhistory
+.RData
 
 ## QtCreator
 Rcpp.pro
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2015-11-06  Kevin Ushey <kevinushey@gmailcom>
+
+        * inst/include/Rcpp/vector/Subsetter.h: Add sugar math operators
+        * inst/unitTests/runit.subset.R: Unit tests
+        * inst/unitTests/cpp/Subset.cpp: Unit tests
+        * inst/NEWS.Rd: NEWS entry
+
 2015-11-01  Qiang Kou <qkou@umail.iu.edu>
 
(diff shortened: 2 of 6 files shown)


kevinushey:
Oops -- looks like the implementation actually updates the original object rather than generating a new one with the result! Will update.

PR #401, opened 2015-11-24
adding slack token

eddelbuettel:
playing with [slackhq](http://slack.com) integration which I have come to like for Rblpapi


eddelbuettel:
That worked as expected and is otherwise a minimal change, so I will merge this directly.

PR #404, opened 2015-11-26
const_iterators for CharacterVector now work analogously to iterators

dcdillon:
Happy Thanksgiving.

Been holding on to this one for a while and since we're between releases, now seemed like a good time.

This basically just makes `const_iterators` for `CharacterVector` work just like their non-const counterparts.


eddelbuettel:
Thanks, Dan.

eddelbuettel:
This worked out swimmingly too so merging now.

Issue #266, opened 2015-02-24
Additional ways to bind to external dependencies in sourceCpp


jjallaire:
There is a way for CRAN packages to declare system dependencies and then meet them at configure time (e.g gputools has a SystemDependency on Nvidia's CUDA toolkit which is ultimately bound using a configure script).

However, there is nothing analogous for sourceCpp, making it difficult to bind to system libraries for ad-hoc work unless they are already wrapped in an R package. A couple of things we could do to bridge this gap:
1. Create an `Rcpp::pkgconfig` attribute that queries pkg-config for whatever named dependencies it is passed. For example:

``` cpp
// [[Rcpp::pkgconfig(opencv)]]
```
1. As an escape hatch for when pkg-config isn't available, create an `Rcpp::buildenv` attribute that can be used to define arbitrary build environment variables:

``` cpp
// [[Rcpp::buildenv(PKG_CXXFLAGS = "...")]]
// [[Rcpp::buildenv(PKG_LIBS = "...")]]
```

Granted we have plugins already for the scenarios where you need arbitrary environment variables however plugins that aren't built into Rcpp are somewhat awkward to register as you need to execute some R code in your session to do so (e.g. in .Rprofile). The `buildenv` attribute would be more self-contained (albeit somewhat more hard-coded).

At first I didn't think we needed this sort of mechanism because "everything can be in an R package" but as it turns out even CRAN doesn't hold code to that standard (and some stuff is very difficult to get into a package that's distributed on CRAN).

eddelbuettel:
This is a *very hard* problem.

Even the CRAN solution does not work as the declaration is "loose".  Any such dependency often differs in name between distros or OSs -- thing eg how to encode the (simpler) dependency for, say, MySQL/MariaDB headers and library, for PostgreSQL, ...   There simply is no standard.

I fear we have have the same problem here.  That said, I quite like that you at least attempt to pass _some_ configuration options (which may well be local and not portable) down to the build.  This goes the right way.   (Then again people always had `Sys.setenv()` and I am not sure how many people used that....)

jjallaire:
Yes, I'm thinking that this is inherently a single-platform solutions
(although I do think for some things pkg-config will give you reasonable
portability between OSX and Linux).

Sys.setenv definitely works, but similar to registering plugins you have to
execute it out-of-band (e.g. it's not a documented/reproducible part of the
translation unit). Allowing a syntax for setting environment variables
directly in the C++ file would IMHO be an improvement.

[...]

Issue #163, opened 2014-08-06
Plugin for on-the-fly #define statements ?


eddelbuettel:
With PR #162 adding support for `std::function` conditional on a corresponding `#define`, I was wondering if Attributes should have a helper send down such a `#define`.    

So thinking out aloud in the direction of @jjallaire: Would a `plugin()`, or a new function adding to `CXXFLAGS` make sense?

PR #409, opened 2015-12-22
Attributes.R: allow findPlugin to getInlinePlugin

Thell:
- Reduce need of registering plugins for sourceCpp when the plugin_name
  of //Rcpp::plugins(plugin_name) matches a package name that provides
  an inlineCxxPlugin.

Signed-off-by: Thell Fowler <thell@tbfowler.name>

--->8---

Instead of complaining that `foo` isn't registered and forcing a manual call to `registerPlugin(foo, Rcpp:::.getInlinePlugin(foo))` just do it...

I wasn't sure of what a good test would be or where it would go, so let me know if you want one.

### DIFF
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -552,6 +552,11 @@ sourceCppFunction <- function(func, isVoid, dll, symbol) {
 .findPlugin <- function(pluginName) {
 
     plugin <- .plugins[[pluginName]]
+    if (is.null(plugin)) {
+        plugin <- .getInlinePlugin(pluginName)
+        if (!is.null(plugin))
+            registerPlugin(pluginName, plugin)
+    }
     if (is.null(plugin))
         stop("Inline plugin '", pluginName, "' could not be found ",
              "within the Rcpp package. You should be ",


eddelbuettel:
That is ... pretty clever :)   

What was your use case?  A local package?  Or one of ours where we forgot register, or you hadn't registered yet?

I'll wait for @jjallaire to chime in.

Thell:
Finally back around to RcppMultiPrecision and am doing it right (_better_) this time in regard to TDD...

~~~
# Context Accessibility 
ok 1 Builds using sourceCpp and Rcpp::depends(RcppMP). 
ok 2 Builds using sourceCpp by finding plugin. 
ok 3 Builds using cxxfunction by finding builtin plugin. 
ok 4 Builds using sourceCpp with explictly registered plugin. 
ok 5 Builds using cxxfunction with explicitly registered plugin. 
# Context C++ 
ok 6 C++ tests pass 
ok 7 Exported tests pass 

Tests complete
~~~

Recently did some playing with a `gprof` settings plugin since I kept forgetting to set the various environment flags, only to forget to registerPlugin. This should help my failing focus. :pray:

jjallaire:
This basically makes `Rcpp::plugin` behave just like `Rcpp::depends` if it can't find an already registered plugin. If you prefer to house your plugin in a package why not just use `Rcpp::depends` in the first place?

My concern here is that the use case for an entire R package that does nothing other than provide a plugin might be pretty narrow. But perhaps your package actually does quite a bit more than that and this use case is entirely legit. A bit more context would be helpful.

eddelbuettel:
I've also been mulling this over and realized that I wasn't sure when a provided plugin was _not_ already registered.  Isn't this just a matter of you having to do The Right Thing in the startup of your package?  But maybe I am missing something.  

As @jjallaire said, context would be nice...

Thell:
Perhaps I'm looking at things slightly askew... To me the primary difference in `Rcpp:depends` and `Rcpp::plugins` is __how__ they effect the compilation environment, _not_ if the are 'findable' or not.

`dependsAttribute`:  
> The include directories of the specified packages are added to the CLINK_CPPFLAGS environment variable. **In addition, if the referenced package provides an inline plugin it is called to determine additional environment variables required to successfully build.**

`pluginsAttribute`:  
> Rcpp::plugins - When included within a sourceCpp translation unit, the **configuration-related fields of the plugin (e.g. env and LinkingTo) are utilized**, however **the code-generation fields (e.g. includes and body) are not**.

Am I wrong?

@jjallaire your concern about the use case for a package that does nothing other than provide a plugin being a narrow one is spot on. It is narrow, yet used. Consider the builtin plugins now available. We don't register those, we just use them and they are found; but they don't break the definition of a `pluginsAttribute` becoming a `dependsAttribute`.

Like I alluded to earlier... It's a `PR` born out of my frustration over a period of time of wishing I didn't need to `Sys.setenv` so often, which led to making some plugin definitions, which led to frustration of _having_ to register them.

If your still wary let me know before closing this out and I'll create a more compelling argument.

jjallaire:
How about if plugins could be found via standard function lookup (sort of like we do for R Markdown custom formats). Something like this:

```cpp
// [[Rcpp::plugins(foo::bar)]]
```

Thell:
@jjallaire Thanks for the further consideration. As long as it doesn't complain and tell _me_ to do something when it could 'just do it'. :smile:

Doing a lookup makes me think of some of the other attributes proposals (like #163, #37, and #266), and I have to say, the _simplicity_ of just checking `.getInlinePlugin` prior to failing without altering the mechanics of plugins has its merit.

jjallaire:
I'd like to think on this a bit more. To me having an entire R package
dedicated to simply providing the value of a couple of environment
variables feels pretty heavyweight. It seems like if you really wanted this
behavior you could also just manually register your plugin on package load.
This would have the advantage of being able to provide multiple plugins in
a single package.

Thell:
While you think on it a bit more, could you also consider letting me know the _drawback_ of simply having it do what it is telling me to do when it is clear it can do it without me?

jjallaire:
It's not telling you to write an R package that contains a single plugin
and nothing else. It's just saying register your plugin *somehow* (likely
in .Rprofile or in onLoad of a package). The drawback is encouraging a
style of creating an R package for a *single* plugin which I think would be
an abuse of the R package system.

Thell:
@jjallaire  Ahhh, thanks, now I understand your concern. I don't agree with it, but I understand it. Give it some thought after the holiday. Remember I'm just trying to resolve an occasional annoyance in a __very simple__ and, in my mind, obvious way...

I'll create an example...

eddelbuettel:
I tend do think about the need or usefulness of plugins in the sense of packages providing headers to be 'plugged in'.   When we have a package, registration is natural.

Where there is no package, I think one can indeed cook up the code for the plugin and its registration in an ad-hoc code snippet, maybe set in `~/.Rprofile`.

Thell:
@jjallaire 
I've created [an __example__ plugin package](https://github.com/Thell/PluginGperf). There's only one file of consequence, the [`R/inline.R`](https://github.com/Thell/PluginGperf/blob/master/R/inline.R).

This `PR` would _only_ remove one function and one step from the usage (`register.plugin`) for this and similar uses (of which I _can_ think of more) but during iterative testing/developing across multiple sessions in multiple disposable docker containers, with different configs, it removes _countless_ times of saying "Oh yeah! Register." :anguished:

@eddelbuettel LMAO, that's __exactly__ what brought this PR into existence! If it wasn't frustrating I surely would __not__ have thought to make it automatic. :wink: 

> Recently did some playing with a gprof settings plugin since I kept forgetting to set the various environment flags, only to forget to registerPlugin. This should help my failing focus.

Anyhow; that settings plugin is well disposed of since it was completely aimed at a particular project.

jjallaire:
The side effect of your plugin registration is a call to setHook. If
the plugin is called multiple times within a given R session then you
will get multiple copies of the hook registered. Further, this hook
will exist for *all* invocations of sourceCpp in that session whether
they have anything to do with gprof.

I think heavyweight things like global hooks that produce console
output on every single call to sourceCpp is something that users need
to opt into explicitly. You probably want another global function that
establishes this hook.

I think what is needed here is a package that registers the plugin
onLoad and has an additional R function which installs the hook. In
terms of your docker containers, etc. all you need is a single line in
your .Rprofile:

library(PluginGperf)

If you want the hook installed as well then:

library(PluginGperf)
installOutputHook()

I think this is a straightforward scheme that is both explicit and reproducible.

Thell:
@jjallaire Glad to see you saw the first line of the 'todo' in the README, but a little disappointed you totally glossed over the point. Truly; I appreciate your viewpoint of _"The drawback is encouraging a
style of creating an R package for a *single* plugin which I think would be
an abuse of the R package system."_ and perhaps the example should only have set the compiler flags, but without knowing where the dynLib was built it seemed, I don't know... pointless?

Honestly, __all__ I want to do is reduce an irritant.

jjallaire:
Now that I understand the whole pipeline better I'd actually suggest that
you implement this in a slightly different way that would overcome a lot of
the friction. Since the user already needs to in some fashion "opt-in" to
the scheme by (i.e. adding an attribute and running google-pprof over the
build output) how about creating a wrapper function that:

1. Sets (and unsets on.exit) the environment variables needed by Gperf

2. Do the same for the build hook (that way it doesn't hang around after
it's used either)

2. Runs the google-pprof utility at the end of the build

Thell:
Thanks again for the further feedback. I'm afraid I made a mistake in using gprof as the example, especially with the helper output... Please don't focus on 'gprof', this could just as easily be any other linked-in tool with its own frontend.

Just because it annoyed me too... [inline.R](https://github.com/Thell/PluginGperf/blob/master/R/inline.R) is altered to better handle the multi-registration issue using `onLoad` and an `active` state option variable that toggles with the calling of `inlineCxxPlugin` and the execution of the hook.

Regarding your point 2, I thought about that this morning, and while I think it would be fantastic to have a `setHookOnce` (ala `debugonce`) that doesn't exist and with the way hooks are handled I don't see how one would `unset` a particular hook. Is that even possible? (Something for another time perhaps...).

Thell:
hmmm it _is_ possible to 'unset' a hook by providing a named formal then iterating over the list in an `on.exit` function and setting the list elem to `NULL`; but I'd bet dollars to donuts it breaks `Attributes.R` looping over the hooks...

eddelbuettel:
@Thell this has gotten a little stale since December -- shall we just close it?

Thell:
I'd reiterate https://github.com/RcppCore/Rcpp/pull/409#issuecomment-167132860  and ask that this stayed open just to prod future thoughts of improvements to the plugins system. But if it is hurting some stat, feel free...

eddelbuettel:
Correct me if I am wrong but it seems to me that @jjallaire gently pushed the ball back into your court, and I had seen no movement.  I don't believe in leaving things open for the sole purpose of a Post It-note.  If there is something new to discuss we can reopen, or start a properly tagged issue ticket.  As far as this PR goes, if we won't apply it I'd be inclined to close it.

Issue #410, opened 2015-12-29
RCPP_CTOR_ASSIGN compiler warning in Rcpp/Module.h


jpritikin:
I might be able to figure out the cause of this warning, but you could probably figure it out much faster.

````
/opt/R/x86_64-pc-linux-gnu-library/3.2/Rcpp/include/Rcpp/Module.h: In copy constructor ‘Rcpp::CppClass::CppClass(const Rcpp::CppClass&)’:
/opt/R/x86_64-pc-linux-gnu-library/3.2/Rcpp/include/Rcpp/Module.h:404:23: warning: base class ‘class Rcpp::S4_Impl<Rcpp::PreserveStorage>’ should be explicitly initialized in the copy constructor [-Wextra]
      RCPP_CTOR_ASSIGN(CppClass)
                       ^
/opt/R/x86_64-pc-linux-gnu-library/3.2/Rcpp/include/Rcpp/macros/interface.h:27:1: note: in definition of macro ‘RCPP_CTOR_ASSIGN’
 __CLASS__( const __CLASS__& other ){                                           \
 ^
/opt/R/x86_64-pc-linux-gnu-library/3.2/Rcpp/include/Rcpp/Module.h: In copy constructor ‘Rcpp::CppObject::CppObject(const Rcpp::CppObject&)’:
/opt/R/x86_64-pc-linux-gnu-library/3.2/Rcpp/include/Rcpp/Module.h:416:26: warning: base class ‘class Rcpp::S4_Impl<Rcpp::PreserveStorage>’ should be explicitly initialized in the copy constructor [-Wextra]
         RCPP_CTOR_ASSIGN(CppObject)
                          ^
/opt/R/x86_64-pc-linux-gnu-library/3.2/Rcpp/include/Rcpp/macros/interface.h:27:1: note: in definition of macro ‘RCPP_CTOR_ASSIGN’
 __CLASS__( const __CLASS__& other ){                                           \
 ^
````

eddelbuettel:
1. Complete reproducible example, please, or it is no bug report or issue.
2. What compiler? What OS? What version of things?
3. Just add `-Wno-extra` to your compiler options to suppress this. It is a _warning_.

jpritikin:
Rcpp 0.12.2, R 3.2.2, gcc (Tanglu 5.2.1-16) 5.2.1 20150903

eddelbuettel:
g++-5 is known to be pickier.  Pull requests welcome.

[...]

PR #411, opened 2015-12-29
Initialize base class in the copy constructor

jpritikin:
See issue #410 for background.


eddelbuettel:
Looks ok -- I had in the meanwhile looked at #410 and the actual file and wondered if we always had `Base` around.  This makes it clearer.

I'll pull this in manually and run a full reverse depends check on it.  If that doesn't throw anything I vote in favor of including this.

PR #412, opened 2016-01-08
Add extra checks to demangler_one

vonosmas:
Be more defensive about the output of backtrace_symbols, passed to demangler_one: don't assume the string would always have plus sign and space (e.g. the symbol may be unavailable) to prevent against crashes (e.g. resize would throw std::length_error for std::string::npos - 1).

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -42,10 +42,13 @@ using namespace Rcpp;
         static std::string demangler_one(const char* input) {
             static std::string buffer;
             buffer = input;
-            buffer = buffer.substr(0, buffer.find_last_of('+'));
+            const auto last_plus =  buffer.find_last_of('+');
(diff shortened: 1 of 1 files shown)


eddelbuettel:
We prefer issue tickets being opened before pull requests.  

Can you (and post it just here) please motivate _why_ we want a change and _what_ you want to change?   A reproducible example would not hurt either.  

Lastly, you may have noticed that your pull request **failed** the continuous integration test, so in order for this to be considered you probably want to correct that.

kevinushey:
`auto` is a C++11-ism; Rcpp needs to conform to the C++98 standard. Can you please explicitly declare the types you need?

vonosmas:
Sure, sorry for the noise. I will follow up with more precise problem description, and updated patch shortly.

PR #414, opened 2016-01-09
Feature/adding contributing.md 

eddelbuettel:


kevinushey:
Looks great! Thanks for the minor tweaks.

eddelbuettel:
So ... shall we fold it in?

kevinushey:
Let's perhaps wait for one last look from @jjallaire and @thirdwing and then fold it in.

eddelbuettel:
:+1:

thirdwing:
:+1:

eddelbuettel:
Thank you both.  I will fold it now.  Release 0.12.3 still on track and grinding through rev.dep. checks.

PR #417, opened 2016-01-13
const support for Nullable<>

krlmlr:
cargo-culted from many other implementations of `operator SEXP() const`

This came up when trying to use `const Nullable<...>&` in https://github.com/rstats-db/RMySQL/pull/90/files#r49576837

Seemed too minor to open an issue first. What would a test for this look like?

CC @peternowee

### DIFF
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -77,7 +77,8 @@ namespace Rcpp {
          * @throw 'not initialized' if object has not been set
          */
         inline operator SEXP() const {
-            return get();
+            checkIfSet();
+            return m_sexp;
         }
 
         /**
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -76,7 +76,7 @@ namespace Rcpp {
          *
          * @throw 'not initialized' if object has not been set
          */
-        inline operator SEXP() {
+        inline operator SEXP() const {
             checkIfSet();
             return m_sexp;
         }
@@ -86,7 +86,7 @@ namespace Rcpp {
          *
          * @throw 'not initialized' if object has not been set
          */
-        inline SEXP get() {
+        inline SEXP get() const {
             checkIfSet();
             return m_sexp;
         }
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -77,8 +77,7 @@ namespace Rcpp {
          * @throw 'not initialized' if object has not been set
          */
         inline operator SEXP() const {
-            checkIfSet();
-            return m_sexp;
+            return get();
         }
 
         /**
--- a/inst/unitTests/cpp/misc.cpp
+++ b/inst/unitTests/cpp/misc.cpp
@@ -179,22 +179,22 @@ void test_stop_variadic() {
 }
 
 // [[Rcpp::export]]
-bool testNullableForNull(Nullable<NumericMatrix> M) {
+bool testNullableForNull(const Nullable<NumericMatrix>& M) {
     return M.isNull();
 }
 
 // [[Rcpp::export]]
-bool testNullableForNotNull(Nullable<NumericMatrix> M) {
+bool testNullableForNotNull(const Nullable<NumericMatrix>& M) {
     return M.isNotNull();
 }
 
 // [[Rcpp::export]]
-SEXP testNullableOperator(Nullable<NumericMatrix> M) {
+SEXP testNullableOperator(const Nullable<NumericMatrix>& M) {
     return M;
 }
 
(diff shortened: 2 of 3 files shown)


eddelbuettel:
@dcdillon Care to look at this (and tests) ?

dcdillon:
I will take a peek soon.  May not have time until the weekend (depends on the wife).

inst/unitTests/cpp/misc.cpp
```diff
@@ -179,22 +179,22 @@ void test_stop_variadic() {
 }
 
 // [[Rcpp::export]]
-bool testNullableForNull(Nullable<NumericMatrix> M) {
+bool testNullableForNull(const Nullable<NumericMatrix>& M) {
```
eddelbuettel: Methinks we may want these as `const` variants (post PR) as well as non-`const` variants just to be sure.

inst/include/Rcpp/Nullable.h
```diff
@@ -76,9 +76,8 @@ namespace Rcpp {
          *
          * @throw 'not initialized' if object has not been set
          */
-        inline operator SEXP() {
-            checkIfSet();
```
eddelbuettel: Why did you remove `checkIfSet()` ?

krlmlr: Now forwarded to get().

eddelbuettel: I can see that.  Let me rephrase:  Why did you make that change? What is the reasoning?

I see no relation in that and the `const` change.  So why mess with it?

inst/unitTests/cpp/misc.cpp
krlmlr: It's no big deal to re-add the pre-PR versions, but I wonder under what circumstances code can succeed for `const&` but fail for a non-const object?

dcdillon:
The question I have is with regards to the return type of SEXP of the `get()` function.  It doesn't provide any safety...so what is the use case of a `const Nullable<T>` object?

inst/include/Rcpp/Nullable.h
krlmlr: You're right, it's an unrelated change. I included it, because it affects code in close proximity to the code that I was touching anyway. The reason for that change: in the unlikely event that the implementation of get() changes, you will also want to change the implementation of operator SEXP(), because they seem synonymous.

That said, I'm totally fine with excluding that particular change from this PR, if you prefer that.

eddelbuettel:
As for your motivation in https://github.com/rstats-db/RMySQL/pull/90/files#r49576837, why not just initialize with empty strings a la `std::string foo = ""`.   Why insinst on `Nullable` and require this change?

dcdillon:
What I would propose is adding an `T as()` function and a `T as() const` function to `Nullable` where the first one simply returns a `T(m_sexp)` and the second one returns a "make deep copy of SEXP and then create a `T` out of it" and return it.  This way we get actual const safety out of a `const Nullable<T>`.  It means that the efficiency of `T as() const` function would be really crappy, but I see no other way to make `const Nullable<T>` provide any safety at all.

krlmlr:
The strings are passed to a C API which makes a difference between `NULL` and `""`. `Rcpp::Nullable` seemed handy.

eddelbuettel:
@dcdillon I like that.  May look into what `Rcpp::clone()` does.

@krlmlr The idiom `std::string foo = "hello, world"; const char bar = foo.c_str();` is about as old `std::string` and gets you a C API string.  Again, no motivation for `Nullable` yet.

krlmlr:
In that C API, passing `NULL` means "get your data elsewhere", while passing `""` means "this is my data, it is an empty string". How do I encapsulate that difference in a `std::string`?

eddelbuettel:
Right. `std::string` alone can't do that as it can't hold a `NULL` (pointer).

eddelbuettel:
Going to merge this as it should be of limited impact in its current form.

kevinushey:
Also agree that the conversion operator should be `const`. :+1:

dcdillon:
@kevinushey why should something that returns a value that allows you to modify the contents be `const`?  I know it's done as a convenience across Rcpp, but it's bastardizing `const` pretty horribly.  That is to say, if you can get a `SEXP` from a `const` function, the world is your oyster.  This would seem to go against any "traditional" meaning of `const`.

eddelbuettel:
@dcdillon "This is R. Everything is bastardized."

We are also cheating somewhat with function interfaces like `double foo(const & NumericVector x)` because we cannot really undo the P in the `SEXP`.  It is what it is, but we all feel it is better to adhere to, and converge to, standard idioms.  As such, no harm in `const` qualifiers.

dcdillon:
Yeah...i was more just wondering the logic, not arguing with it.

PR #423, opened 2016-01-17
Added as() and clone() to Nullable

### LINKED ISSUE #421: Nullable additions
I would like to add two convenience member functions to `Rcpp::Nullable<T>` specifically:

```c++
T Nullable::as()
{
    return T(m_sexp);
}

T Nullable::clone() const
{
    return T(Rcpp:::clone(m_sexp));
}
```

These are analogous to the `Rcpp::as` and `Rcpp::clone` functions and just add convenience to the `Nullable` class.

Any objections?

dcdillon:
Fixes #421

### DIFF
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -116,6 +116,16 @@ namespace Rcpp {
          */
         inline bool isSet(void) const { return m_set; }
 
+        /**
+         * Returns m_sexp as a T
+         */
+        inline T as() { return Rcpp::as< T >(get()); }
+
+        /**
+         * Return a clone of m_sexp as a T
+         */
+        inline T clone() const { return Rcpp::clone(as()); }
+
     private:
         SEXP m_sexp;
         bool m_set;
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -124,7 +124,7 @@ namespace Rcpp {
         /**
          * Return a clone of m_sexp as a T
          */
-        inline T clone() const { return Rcpp::clone(as()); }
+        inline T clone() const { return Rcpp::clone(Rcpp::as< T >(get())); }
 
     private:
         SEXP m_sexp;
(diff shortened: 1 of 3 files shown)


eddelbuettel:
LGTM

inst/include/Rcpp/Nullable.h
```diff
@@ -116,6 +116,16 @@ namespace Rcpp {
...
 
+        /**
+         * Returns m_sexp as a T
+         */
+        inline T as() { return Rcpp::as< T >(get()); }
+
+        /**
+         * Return a clone of m_sexp as a T
+         */
+        inline T clone() const { return Rcpp::clone(Rcpp::as< T >(get())); }
```
dcdillon: So technically the call is just AWFUL it:

* Shallow copies in the return from `Rcpp::as()`
* Deep copies inside the `Rcpp::clone()' function
* Shallow copies in the return from `Rcpp::clone()`
* Shallow copies in the return from `Nullable::clone()`
* Probably shallow copies again when the end user receives it.

There should really be a better way.  Best I can come up with is a `SEXP Rcpp::clone(SEXP obj)` function which would remove the need for some of this.

Any other thoughts?

eddelbuettel: Don't we usually call `wrap()` to convert to a `SEXP` which would make feeding in `clone()` more straightfoward?

dcdillon:
OK.  It's a lot better now.  We keep it as a `SEXP` until `return` time.  So now it's down to:

* Construct T
* Copy in return

Much more satisfying.

dcdillon:
And same performance tweak to `as()`.  Now I think it's good.

eddelbuettel:
Well, nobody cared to second this but on the upside nobody howled either :) 

Planning to fold this in later this morning.

dcdillon:
No problem.  It's just convenience stuff.

dcdillon:
Added `isUsable()` to indicate whether or not the `Nullable<T>` can be used as a `T`.  This is equivalent to checking `isSet()` followed by `!isNull()` but won't throw an exception.

eddelbuettel:
No one shouted so will fold this one now.

Issue #424, opened 2016-01-18
Sugar median function


nathan-russell:
In light of a recent Stack Overflow question, I put together a sugar function corresponding to `stats::median` - more specifically, `median.default`. At the moment, I have two implementations worked out: one which has an extra boolean argument for specifying whether or not to remove missing values (like `median.default`'s `na.rm` option), and one which does not (it just returns `NA` for input containing one or more missing values). I've posted what are more or less final drafts of these source files [here](https://gist.github.com/nathan-russell/6e8824240bfecdef6ac8). 

-----

#### With the na.rm option 
The main benefit of this version is of course having the option to remove missing values - thus more closely conforming to `median.default`. As a result, however, the interface function is a little ugly: 

```c++ 
template <int RTYPE, bool NA, typename T>
inline typename sugar::median_detail::result<RTYPE>::type
median(const Rcpp::VectorBase<RTYPE, NA, T>& x, bool na_rm = false) {
    switch (static_cast<int>(na_rm)) {
        case 0:
            return sugar::Median<RTYPE, NA, T, false>(x);
        case 1:
            return sugar::Median<RTYPE, NA, T, true>(x);
        default:
            return Rcpp::traits::get_na<RTYPE>();
    }
}
```
The `na.rm` is encoded as a template parameter, and after playing around with some TMP for a while and failing to find a more elegant solution, I settled on the `switch` statement logic above. I'm actually not sure that this is even avoidable, but am certainly open to suggestions if there are better alternatives. 

----- 

#### Without the na.rm option 
While this version obviously lacks some of the above functionality, it is more consistent with similar sugar functions such as `mean` and `max` which do not provide an `na.rm` option. Accordingly, the interface is more sugar-esque: 

```c++ 
template <int RTYPE, bool NA, typename T>
inline sugar::Median<RTYPE, NA, T>
median(const Rcpp::VectorBase<RTYPE, NA, T>& x) {
    return sugar::Median<RTYPE, NA, T>(x);
}
```

----- 

#### General remarks 

In the testing I've done so far, both versions seem to mirror `median.default` pretty closely. These are some slight differences that I've come across:  

+ Attributes are not preserved on `Date` and `POSIXt` objects; this is also true of the `Rcpp::mean`, however. 
[...]

PR #425, opened 2016-01-18
Sugar median with unit tests

nathan-russell:
As discussed in #424, this adds a sugar function `Rcpp::median` with corresponding unit tests.

### DIFF
--- a/inst/include/Rcpp/sugar/functions/median.h
+++ b/inst/include/Rcpp/sugar/functions/median.h
@@ -88,7 +88,7 @@ class Median {
 public:
     typedef typename median_detail::result<RTYPE>::type result_type;
     typedef typename Rcpp::traits::storage_type<RTYPE>::type stored_type;
-    enum { RTYPE2 = median_detail::result<RTYPE>::rtype };
+    enum { RESULT_RTYPE = median_detail::result<RTYPE>::rtype };
     typedef T VECTOR;
     
 private:
@@ -100,11 +100,11 @@ public:
     
     operator result_type() {
         if (x.size() < 1) {
-            return Rcpp::traits::get_na<RTYPE2>();
+            return Rcpp::traits::get_na<RESULT_RTYPE>();
         }
         
         if (Rcpp::any(Rcpp::is_na(x))) {
-            return Rcpp::traits::get_na<RTYPE2>();
+            return Rcpp::traits::get_na<RESULT_RTYPE>();
         }
         
         R_xlen_t n = x.size() / 2;
@@ -126,7 +126,7 @@ class Median<RTYPE, NA, T, true> {
 public:
     typedef typename median_detail::result<RTYPE>::type result_type;
     typedef typename Rcpp::traits::storage_type<RTYPE>::type stored_type;
-    enum { RTYPE2 = median_detail::result<RTYPE>::rtype };
+    enum { RESULT_RTYPE = median_detail::result<RTYPE>::rtype };
     typedef T VECTOR;
     
 private:
@@ -138,7 +138,7 @@ public:
     
     operator result_type() {
         if (!x.size()) {
-            return Rcpp::traits::get_na<RTYPE2>();
+            return Rcpp::traits::get_na<RESULT_RTYPE>();
         }
         
         R_xlen_t n = x.size() / 2;
@@ -160,7 +160,7 @@ class Median<RTYPE, false, T, NA_RM> {
 public:
     typedef typename median_detail::result<RTYPE>::type result_type;
     typedef typename Rcpp::traits::storage_type<RTYPE>::type stored_type;
-    enum { RTYPE2 = median_detail::result<RTYPE>::rtype };
+    enum { RESULT_RTYPE = median_detail::result<RTYPE>::rtype };
     typedef T VECTOR;
     
 private:
@@ -172,7 +172,7 @@ public:
     
     operator result_type() {
         if (x.size() < 1) {
-            return Rcpp::traits::get_na<RTYPE2>();
+            return Rcpp::traits::get_na<RESULT_RTYPE>();
         }
         
         R_xlen_t n = x.size() / 2;
@@ -281,14 +281,8 @@ public:
 template <int RTYPE, bool NA, typename T>
 inline typename sugar::median_detail::result<RTYPE>::type
 median(const Rcpp::VectorBase<RTYPE, NA, T>& x, bool na_rm = false) {
-    switch (static_cast<int>(na_rm)) {
-        case 0:
-            return sugar::Median<RTYPE, NA, T, false>(x);
-        case 1:
-            return sugar::Median<RTYPE, NA, T, true>(x);
-        default:
-            return Rcpp::traits::get_na<RTYPE>();
-    }
+    if (!na_rm) return sugar::Median<RTYPE, NA, T, false>(x);
+    return sugar::Median<RTYPE, NA, T, true>(x);
 }
 
 } // Rcpp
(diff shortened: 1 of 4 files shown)


inst/include/Rcpp/sugar/functions/median.h
```diff
@@ -0,0 +1,297 @@
+// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; tab-width: 8 -*-
```
eddelbuettel: I have a much newer version of this line but then I never said anything official about it...

inst/include/Rcpp/sugar/functions/median.h
```diff
@@ -0,0 +1,297 @@
...
+//
+// median.h: Rcpp R/C++ interface class library -- median
+//
+// Copyright (C) 2012 - 2013 Dirk Eddelbuettel and Romain Francois
```
eddelbuettel: 2012-2013 makes no sense (and just shows the copy and pasty).  If the file is all yours may as well drop us.  Or keep us in as a courtesy. Your file, your call.

inst/include/Rcpp/sugar/functions/median.h
```diff
@@ -0,0 +1,297 @@
...
+// Copyright (C) 2016 - Nathan Russell
```
eddelbuettel: If there is one year only, no hyphen.  (Not that it matters but as I am in nitpick mode...)

eddelbuettel:
This looks good. I had in the back of my mind a though that R's `median()` had a bazillion further options, but maybe I was thinking about `quantile()` and its (crazy but impressive) `type=` argument.

So this is probably good too but one or two seconds would be nice.

nathan-russell:
Okay, that sounds good. In the meantime I will make the appropriate changes regarding your comment above. Is this the newer version of the formatting line you referred to: 

`-*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-`

eddelbuettel:
Yep. Hard tabs are evil, which we learned eventually.  But I leave commits touching every single file to @kevinushey :)  -- in short, bulk changes aren't worth it either.

There isn't too much about code style you don't already know or do.  Four spaces is better than too, indentation pretty much like R Core and other sensible people do etc pp.  No biggies.

nathan-russell:
Agreed on the spaces vs. tabs - it was just sloppy copy & paste from another sugar file, and being a `vim` guy (sorry!) the outdated formatting settings didn't catch my eye.

kevinushey:
FWIW we now use [`.dir-locals.el`](https://github.com/RcppCore/Rcpp/blob/master/.dir-locals.el) to enforce indentation rules when editting with Emacs; does Vim have an analogue for directory-local variables / indentation rules?

inst/include/Rcpp/sugar/functions/median.h
```diff
@@ -0,0 +1,296 @@
...
+    typedef typename Rcpp::traits::storage_type<RTYPE>::type stored_type;
+    enum { RTYPE2 = median_detail::result<RTYPE>::rtype };
+    typedef T VECTOR;
+    
+private:
+    VECTOR x;
+    
+public:
+    Median(const VECTOR& xx)
+        : x(Rcpp::clone(xx)) {}
```
kevinushey: Just to confirm -- we clone the vector here because the `std::` algorithm we use will modify the passed vector?

inst/include/Rcpp/sugar/functions/median.h
```diff
@@ -0,0 +1,296 @@
...
+
+} // median_detail
+
+// base case
+template <int RTYPE, bool NA, typename T, bool NA_RM = false>
+class Median {
+public:
+    typedef typename median_detail::result<RTYPE>::type result_type;
+    typedef typename Rcpp::traits::storage_type<RTYPE>::type stored_type;
+    enum { RTYPE2 = median_detail::result<RTYPE>::rtype };
```
kevinushey: Maybe instead of `RTYPE2` we could use `RESULT_RTYPE` or `result_rtype` or something?

inst/include/Rcpp/sugar/functions/median.h
```diff
@@ -0,0 +1,296 @@
...
+        return x[n];
+    }
+};
+
+} // sugar
+
+template <int RTYPE, bool NA, typename T>
+inline typename sugar::median_detail::result<RTYPE>::type
+median(const Rcpp::VectorBase<RTYPE, NA, T>& x, bool na_rm = false) {
+    switch (static_cast<int>(na_rm)) {
```
kevinushey: How come we can't just use an `if-else` on the `na_rm` parameter? The `static_cast<int>` is kind of funky.

thirdwing:
Maybe we can also have a `.vim.custom` to set indent.

But if we really want to have a consistent coding style, we might think about cpplint from Google.

eddelbuettel:
Re `.dir-locals.el`, the fanciest / most general scheme I know is via [editorconfig](http://editorconfig.org/) where, as I understand it, you define a 'policy' or outcome and the editor-specific plugins implement it in the editor of choice. 

Alternatively, and there clang-format but all of that is a bit of the heavy-handed side.

eddelbuettel:
My bad for the close and re-open.  Hit the wrong button.

kevinushey:
Other than some minor comments this looks good to me.

nathan-russell:
@kevinushey Thanks for the feedback - I do believe it is necessary to call `Rcpp::clone` in the constructor as above. I just did a rebuild without the use of `clone` and it appears to modify the original vector: 

```c++ 
#include <Rcpp.h>

// [[Rcpp::export]]
double median_dbl(Rcpp::NumericVector x, bool na_rm = false) {
    return Rcpp::median(x, na_rm);
}

// [[Rcpp::export]]
double median_int(Rcpp::IntegerVector x, bool na_rm = false) {
    return Rcpp::median(x, na_rm);
}

// [[Rcpp::export]]
Rcomplex median_cx(Rcpp::ComplexVector x, bool na_rm = false) {
    return Rcpp::median(x, na_rm);
}

// [[Rcpp::export]]
Rcpp::String median_ch(Rcpp::CharacterVector x, bool na_rm = false) {
    return Rcpp::median(x, na_rm);
}


/*** R

set.seed(123)

(xx <- rnorm(5))
#[1] -0.56047565 -0.23017749  1.55870831  0.07050839  0.12928774

median_dbl(xx)
#[1] 0.07050839

xx
#[1] -0.56047565 -0.23017749  0.07050839  0.12928774  1.55870831

set.seed(123)

(xx <- as.integer(rpois(5, 20)))
#[1] 17 25 12 20 27

median_int(xx)
#[1] 20

xx
#[1] 17 12 20 25 27

set.seed(123)

(xx <- rnorm(5) + 2i)
#[1] -0.560476+2i -0.230177+2i  1.558708+2i  0.070508+2i  0.129288+2i

median_cx(xx)
#[1] 0.070508+2i

xx
#[1] -0.560476+2i -0.230177+2i  0.070508+2i  0.129288+2i  1.558708+2i

set.seed(123)

(xx <- sample(letters, 5))
#[1] "h" "t" "j" "u" "w"

median_ch(xx)
#[1] "t"

xx
##[1] "h" "j" "t" "u" "w"

*/
```

----- 

As for the other points, I'll set `RTYPE2` to `RESULT_RTYPE` and swap out the `switch` statement with an `if` statement and commit the changes.

I have not used / am not aware of anything analogous to `.dir-locals.el` or the Emacs configuration lines embedded in the Rcpp headers; I just have the standard `~/.vimrc` and `~/.vim/ftplugin/*.vim` files on my machines.

eddelbuettel:
Time to pull this one in!

nathan-russell:
:+1:

eddelbuettel:
No, thumbs up to _you_ -- that was really well done (but I was a little tied up yesterday and the day before).

PR #432, opened 2016-01-30
fix the error in subset under gc. close #431

### LINKED ISSUE #431: SubsetProxy should validate input vector sizes
Currently, the `get_indices()` functions (which are called in the constructor) do not validate the input sizes We should validate early that the right-hand side vector is either length 1, or same length as the left-hand side vector.

https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Subsetter.h#L40

Note that we do perform this check when actually performing the subset (e.g. https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Subsetter.h#L56-L66) but I think this is happening too late (we should validate before we do any work)

thirdwing:
This is the simpler solution. We can close this if we figure out how to reuse the IntergerVector correctly.

@kevinushey @eddelbuettel

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2016-01-30 Qiang Kou <qkou@umail.iu.edu>
+
+        * inst//include/Rcpp/vector/Subsetter.h: fix the error under gc
+        * inst/unitTests/cpp/Vector.cpp : Add tests
+        * inst/unitTests/runit.Vector.R : Idem
+
 2016-01-29  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Nice work.  Do you think this can go in without fear of side-effects?  

And I can always run another rev.dep check as well.

kevinushey:
Looks good to me -- I don't think there will be any issues.

eddelbuettel:
Right. I'll merge now and set it up to burn some cycles.

Thanks _a big fat bunch_ to both of you for chasing this one down so quickly.

PR #433, opened 2016-02-07
Improved OS X Rcpp Compiler Information

coatless:
Per discussion on the Rcpp dev list, here is a list of proposed changes for the OS X compilation section.

### DIFF
--- a/vignettes/Rcpp-FAQ.Rnw
+++ b/vignettes/Rcpp-FAQ.Rnw
@@ -128,7 +128,7 @@ Specific per-platform notes:
     above detect this and set appropriate flags.
   \item[OS X] users, as noted in the `R Administration' manual \citep[Appendix
     C.4]{R:Administration}, need to install the Apple Developer Tools
-    (\textsl{e.g.}, \texttt{Xcode}) (as well as \texttt{gfortran} if \proglang{R} or
+    (\textsl{e.g.}, \href{https://itunes.apple.com/us/app/xcode/id497799835?mt=12}{Xcode} (OS X $\le 10.8$) or \href{https://developer.apple.com/library/ios/technotes/tn2339/_index.html}{Xcode Command Line Tools} (OS X $\ge 10.9$) (as well as \texttt{gfortran} if \proglang{R} or
     Fortran-using packages are to be built); also see \faq{q:OSX} and
     \faq{q:OSXArma} below. Depending on whether on OS X release before or after
     Mavericks is used, different additional installation may be needed. Consult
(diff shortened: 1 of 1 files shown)


eddelbuettel:
OS X users: @kevinushey @jjallaire @thirdwing   Any comments or good as is?

thirdwing:
:+1:

hrbrmstr:
In the patched Rnw file, when I compile it to PDF I'm not seeing any of the command lines in the 

    <<osx_xcode,engine='bash', eval = F>>=

and:

    <<osx_openmp, engine="Rcpp", eval = F>>=

blocks (but this could just be me not compiling the Rnw file properly)

There are some refs to `Maverick` when it should be `Mavericks`.

    \url{thecoatlessprofessor.com/programming/r-compiler-tools-for-rcpp-on-os-x/}

needs an `http:` in front of it to be "clickable" in Preview.app

+100 for inclusion of the 4.8.2 `gfortran` d/l & install instructions.

Apart from that, these are _really_ helpful instructions for folks on OS X, are spot-on, and pretty much (except for OpenMP which I don't use right now) all things I've had to do (and mimic what I've done).

I'd make the Rnw changes myself but remember just enough latex to be dangerous.

coatless:
@eddelbuettel 
* It seems as if the FAQ is not 100% knitr based? 
    * e.g. some section have `<<lang=bash>>` and `<<lang=cpp>>` instead of `<<engine='bash'>>` and `<<engine='Rcpp'>>`

@hrbrmstr 

* Fixed the one maverick reference & update url to have http://
* The Coatless Professor will be around for a long time as it is my personal website and playground for new tech.
* Instead of posting to r-sig-mac, it probably would be better if R updated: https://cran.r-project.org/bin/macosx/tools/ as it is currently sitting on gfortran 4.2.3

hrbrmstr:
Wasn't implying anything abt you speficially @coatless :-) I've got a longstanding personal site as well but wouldn't suggest linking to it as an official documentation source as I'm just a person vs an org. Just a personal pref.

eddelbuettel:
Just in from a longish run.  To you young'nes:  There was a time when we all wrote vignettes with Sweave.  Nuttin' wrong with.  Outside of `R CMD foo ...` it works eg via the (old) shell script below. I would not (have) let a tool as immature as knitr at my vignettes :)  (but do know for shorter/other ones as eg in drat).

Will comment on substantial issues in a bit.

```shell
edd@max:~/bin$ cat sweave 
#!/bin/bash -e

function errorexit () {
    echo "Error: $1"
    exit 1
}

function filetest () {
    if [ ! -f $1 ]; then
       errorexit "File $1 not found"
    fi
    return 0
}
                       
               
if [ "$#" -lt 1 ]; then
    errorexit "Need to specify argument file"
fi
                           
                   
BASENAME=$(basename $1 .Rnw)
                           
RNWFILE=$BASENAME.Rnw
filetest $RNWFILE
echo "library(tools); Sweave(\"$RNWFILE\")" \
      | R --no-save --no-restore --slave
                               
LATEXFILE=$BASENAME.tex
filetest $LATEXFILE && pdflatex $LATEXFILE
                               
PDFFILE=$BASENAME.pdf
#filetest $PDFFILE && acroread $PDFFILE &
#filetest $PDFFILE && xpdf $PDFFILE &
#filetest $PDFFILE && kpdf $PDFFILE &
edd@max:~/bin$ 
```

eddelbuettel:
Turns out that script no longer works due to extra `$` hanging around. Tst tst.

Oh wait. It's because I am loosing my marbles -- it is because we use the even-less-standard driver based on highlight. My apologies.  *This* script does it:

```r
edd@max:~/git/rcpp/vignettes(master)$ cat ~/bin/hilightsweave.R 
#!/usr/bin/Rscript

## use given argument(s) as target files, or else default to .Rnw files in directory
files <- if (length(commandArgs(TRUE)) == 0) dir(pattern="*.Rnw") else commandArgs(TRUE)

## convert all files from Rnw to pdf using the highlight driver
invisible(sapply(files, function(srcfile) {
    Sweave(srcfile, driver=highlight::HighlightWeaveLatex(boxes=TRUE))
    tools::texi2pdf(gsub(".Rnw", ".tex", srcfile))
    tools::texi2pdf(gsub(".Rnw", ".tex", srcfile), texi2dvi="pdflatex")
}))
edd@max:~/git/rcpp/vignettes(master)$ 
```

But for ease-of-use I replaced the 'published' version on my site with the current draft, so [see here](http://dirk.eddelbuettel.com/code/rcpp/Rcpp-FAQ.pdf).   Time to brew  a new coffee and go over that.

With that, big thanks to @coatless and @hrbrmstr for updates and reviews!

eddelbuettel:
Some of James' markup was not of the right kind for our (admittedly non-standard) use of highlight. 

I fixed that in https://github.com/RcppCore/Rcpp/commit/08ceaaf76ddd79a399e8681005971302744dbc22. New version copied to the [same URL as above](http://dirk.eddelbuettel.com/code/rcpp/Rcpp-FAQ.pdf).

PR #434, opened 2016-02-07
more verbose use of seq(along.with=

eddelbuettel:
Mostly for @jjallaire:

Following a moment of madenss, I now use
```r
    options(#
            # ...
            warnPartialMatchAttr=TRUE,
            warnPartialMatchDollar=TRUE,
            warnPartialMatchArgs=TRUE,
            # ...
    ) 
```

which tickled some warnings from `seq(along=...)` and this PR fixes one.  There may be others, but this is a start.


eddelbuettel:
:+1:

PR #437, opened 2016-02-20
Resolve misleading-indentation warnings from gcc 6

ellert:
This pull request resolves the warnings about misleading indentation issued by gcc 6.
~~~
In file included from ../inst/include/Rcpp/stats/stats.h:32:0,
                 from ../inst/include/Rcpp.h:71,
                 from Date.cpp:31:
../inst/include/Rcpp/stats/norm.h: In function 'double Rcpp::stats::dnorm_0(double, int)':
../inst/include/Rcpp/stats/norm.h:48:5: warning: statement is indented as if it were guarded by... [-Wmisleading-indentation]
     if(!R_FINITE(x)) return R_D__0;
     ^~
../inst/include/Rcpp/stats/norm.h:45:5: note: ...this 'if' clause, but it is not
     if (ISNAN(x) )
     ^~
In file included from ../inst/include/Rcpp/stats/stats.h:37:0,
                 from ../inst/include/Rcpp.h:71,
                 from Date.cpp:31:
../inst/include/Rcpp/stats/lnorm.h: In function 'double Rcpp::stats::plnorm_1(double, double, int, int)':
../inst/include/Rcpp/stats/lnorm.h:85:2: warning: statement is indented as if it were guarded by... [-Wmisleading-indentation]
  if (x > 0)
  ^~
../inst/include/Rcpp/stats/lnorm.h:81:5: note: ...this 'if' clause, but it is not
     if (ISNAN(x) || ISNAN(meanlog) )
     ^~
In file included from ../inst/include/Rcpp/stats/stats.h:38:0,
                 from ../inst/include/Rcpp.h:71,
                 from Date.cpp:31:
../inst/include/Rcpp/stats/weibull.h: In function 'double Rcpp::stats::pweibull_1(double, double, int, int)':
../inst/include/Rcpp/stats/weibull.h:59:2: warning: statement is indented as if it were guarded by... [-Wmisleading-indentation]
  x = -::pow(x , shape);
  ^
../inst/include/Rcpp/stats/weibull.h:57:5: note: ...this 'if' clause, but it is not
     if (x <= 0)
     ^~
In file included from ../inst/include/Rcpp/stats/stats.h:39:0,
                 from ../inst/include/Rcpp.h:71,
                 from Date.cpp:31:
../inst/include/Rcpp/stats/logis.h: In function 'double Rcpp::stats::dlogis_0(double, int)':
../inst/include/Rcpp/stats/logis.h:35:2: warning: statement is indented as if it were guarded by... [-Wmisleading-indentation]
  e = ::exp(-::fabs(x));
  ^
../inst/include/Rcpp/stats/logis.h:31:5: note: ...this 'if' clause, but it is not
     if (ISNAN(x))
     ^~
In file included from ../inst/include/Rcpp/stats/stats.h:41:0,
                 from ../inst/include/Rcpp.h:71,
                 from Date.cpp:31:
../inst/include/Rcpp/stats/exp.h: In function 'double Rcpp::stats::d_exp_0(double, int)':
../inst/include/Rcpp/stats/exp.h:39:3: warning: statement is indented as if it were guarded by... [-Wmisleading-indentation]
   return give_log ? (-x) : ::exp(-x);
   ^~~~~~
../inst/include/Rcpp/stats/exp.h:37:6: note: ...this 'if' clause, but it is not
      if (x < 0.)
      ^~

~~~

### DIFF
--- a/inst/include/Rcpp/stats/exp.h
+++ b/inst/include/Rcpp/stats/exp.h
@@ -27,46 +27,45 @@
 
 namespace Rcpp {
 namespace stats {
-	inline double d_exp_0( double x, int give_log){
+	inline double d_exp_0(double x, int give_log) {
 
 		#ifdef IEEE_754
-		    /* NaNs propagated correctly */
-		    if (ISNAN(x) ) return x + 1.0 ;
+		/* NaNs propagated correctly */
+		if (ISNAN(x)) return x + 1.0;
 		#endif
 
-    	if (x < 0.)
-		return R_D__0;
+		if (x < 0.)
+			return R_D__0;
 		return give_log ? (-x) : ::exp(-x);
 	}
-	inline double q_exp_0( double p, int lower_tail, int log_p){
+	inline double q_exp_0(double p, int lower_tail, int log_p) {
 		#ifdef IEEE_754
 		if (ISNAN(p)) return p + 1.0;
 		#endif
 
-		if ((log_p	&& p > 0) || (!log_p && (p < 0 || p > 1)) ) return R_NaN ;
+		if ((log_p && p > 0) || (!log_p && (p < 0 || p > 1))) return R_NaN;
 		if (p == R_DT_0)
-		return 0;
+			return 0;
 
 		return - R_DT_Clog(p);
 	}
 	inline double p_exp_0(double x, int lower_tail, int log_p) {
-#ifdef IEEE_754
-    if (ISNAN(x) )
-	return x + 1.0 ;
-#endif
+		#ifdef IEEE_754
+		if (ISNAN(x)) return x + 1.0;
+		#endif
 
-    if (x <= 0.)
-	return R_DT_0;
-    /* same as weibull( shape = 1): */
-    x = -x;
-    if (lower_tail)
-	return (log_p
-		/* log(1 - exp(x))  for x < 0 : */
-		? (x > -M_LN2 ? ::log(-::expm1(x)) : ::log1p(-::exp(x)))
-		: -::expm1(x));
-    /* else:  !lower_tail */
-    return R_D_exp(x);
-}
+		if (x <= 0.)
+			return R_DT_0;
+		/* same as weibull(shape = 1): */
+		x = -x;
+		if (lower_tail)
+			return (log_p
+				/* log(1 - exp(x))  for x < 0 : */
+				? (x > -M_LN2 ? ::log(-::expm1(x)) : ::log1p(-::exp(x)))
+				: -::expm1(x));
+		/* else:  !lower_tail */
+		return R_D_exp(x);
+	}
 
 } // stats
 } // Rcpp
--- a/inst/include/Rcpp/stats/lnorm.h
+++ b/inst/include/Rcpp/stats/lnorm.h
@@ -33,15 +33,15 @@ inline double dlnorm_0(double x, int log_p){
 
 #ifdef IEEE_754
     if (ISNAN(x))
-	return x + 1.0 ;
+        return x + 1.0;
 #endif
 
-    if(x <= 0) return R_D__0;
+    if (x <= 0) return R_D__0;
 
-    y = ::log(x) ;
+    y = ::log(x);
     return (log_p ?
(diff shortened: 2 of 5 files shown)


eddelbuettel:
I'll look into this some more once I get home.

nathan-russell:
FWIW this can also be silenced by tacking on a `-Wno-misleading-indentation`, as [here](http://melpon.org/wandbox/permlink/fxsScGMDmq6ouITX), or (preferably, IMO) with explicit braces for control structures, as done [here](http://melpon.org/wandbox/permlink/kxfLVU8HSHqnUtOt).

ellert:
Of course there are other ways to fix this. A pull request is a proposed solution, not the one and only solution. I usually try to make minimal changes when I propose patches to projects where I am not the upstream. This proposed fix is not quite minimal, since you could fix this with fewer changed lines. I opted for making the indentation consistent throughout the file rather than fixing only the lines where the compiler issued warnings. I hope this was not considered too intrusive.

I think that disabling warnings is a bad idea, since it potentially hides real issues. In this case it is just confusing indentation - none of the places where the compiler issues warnings was - as far as I can tell - intended to actually be something else than what is written. (This warning have however flagged real issues that I have reported in other projects, so I think it is a useful warning.) And since these are installed headers, other projects #including them could get the warnings too. even if you disable the warning in this project.

By the way, do you have a coding standard regarding indentation in this package? I only looked at the 5 files that triggered this warning, and most of them had a mixture of indent-with-tabs and indent-with-spaces. For C++ this is of course not an error, but it is confusing. For my patch I chose the one that resulted in the fewest number of changes when I normalised each file, but this was not the same choice for all of the 5 files.

eddelbuettel:
> do you have a coding standard regarding indentation in this package? 

I just mentioned that [over here in #439](https://github.com/RcppCore/Rcpp/pull/439#issuecomment-187158285).

kevinushey:
It looks like this is auto-generated code (https://github.com/ellert/Rcpp/blob/indent-warnings/inst/include/Rcpp/stats/exp.h#L4) so if we do want to fix up the whitespace it should probably occur in the generator file.

eddelbuettel:
The Debian experimental container for Docker has gcc-6 currently uninstallable, some underlying components are not yet in.   So I'll let this rest for now. 

As @kevinushey mentioned, the _real_ fix has to be in the code generators. By the uncoordinated nature of the r-forge-to-github transition, these ended up [here](https://github.com/RcppCore/rcpp-scripts/tree/master/scripts).

eddelbuettel:
I now have `g++-6` from Feb 27 in a (Ubuntu xenial) Docker container, and I can replicate this.

@kevinushey 's point still stands, but I guess I'll bite the bullet and go over this -- semi-manually, with my good friend Emacs -- to get rid of the tabs and give it saner indentation.

kevinushey:
Thanks @eddelbuettel!

eddelbuettel:
And when I said 'go over this' I meant after first accepting the PR which is after all valid (all our grumblings about really having to change the generator notwithstanding).

ellert:
_A question regarding the generator script:_
Am I understanding it correctly that the statement in the files reading "auto generated file (from script/stats.R)" refers to the script/stats.R in the https://github.com/RcppCore/rcpp-scripts repository?
If this is the case, the script in the repo does not seem to be up to date.
The output of the script checked out from git does not match the headers in the Rcpp sources. In particular the headers in the Rcpp sources say "Copyright (C) 2010 - 2011", while those generated by the script say "Copyright (C) 2010". This this doesn't seem to be the version of the script that was used co generate the current versions of the headers. Am I looking in the wrong place?

eddelbuettel:
How can I put this delicately?  Let's try this:  _Rcpp was mostly written by two authors who put different levels of emphasis on these different components._  Maybe I just leave it there. 

At some point I made some catch-up commits to the scripts to reflect later updates.  I do apologise for one confusion---these scripts may either be in rcpp-scripts or the sibling rcpp-core repository. 

But yes. we should clean these up.  Help (in testing and everything else) welcome!

eddelbuettel:
Time for the merge and post-cleanup in a subsequent PR.

PR #438, opened 2016-02-20
Fix R shebangs

ellert:
I am not sure what this is supposed to be. I assume /usr/bin/r is a misspelt /usr/bin/R since these are all R macros. But many of them say "/usr/bin/R -t" and -t is not a valid option to R.
~~~
$ R -t
WARNING: unknown option '-t'
~~~
So what is this -t flag supposed to do?

The shebang with /usr/bin/r doesn't work
~~~
$ inst/examples/FastLM/benchmark.r
bash: inst/examples/FastLM/benchmark.r: /usr/bin/r: bad interpreter: No such file or directory
~~~
Only a few of the files with an /usr/bin/r shebang (15 of 52) actually have the executable permissions set. Why does some have executable permissions set and some not - is this just a mistake, or is this inconsistency deliberate. A shebang in a file without executable permissions does not make sense, so I would suggest either to set the executable permission or remove the shebang.

### DIFF
--- a/inst/examples/ConvolveBenchmarks/exampleRCode.r
+++ b/inst/examples/ConvolveBenchmarks/exampleRCode.r
@@ -1,4 +1,4 @@
-#!/usr/bin/r
+#!/usr/bin/R
 
 suppressMessages(require(Rcpp))
 set.seed(42)
--- a/inst/examples/ConvolveBenchmarks/overhead.r
+++ b/inst/examples/ConvolveBenchmarks/overhead.r
@@ -1,4 +1,4 @@
-#!/usr/bin/r
+#!/usr/bin/R
 
 set.seed(42)
 a <- rnorm(100)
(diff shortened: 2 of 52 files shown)


dmbates:
/usr/bin/r is intentional. See https://github.com/eddelbuettel/littler

eddelbuettel:
@dmbates is entirely correct. The [littler](https://cloud.r-project.org/web/packages/littler/index.html) package is even a dependency of Rcpp.  There is an institutional reason why it can't provide `/usr/bin/r` but still -- this is a feature.

ellert:
Thank you for clarifying this. This was not at all clear from reading the DESCRIPTION file. There is no dependency on littler mentioned there.

eddelbuettel:
Apologies:  I confused it with `debian/control` which needs it.  And has it as [r-cran-rcpp](https://packages.debian.org/sid/r-cran-rcpp) is a working and correct Debian (and Ubuntu and ...) package.

PR #439, opened 2016-02-21
Permissions

ellert:
This adds a Suggests: littler to the DESCRIPTION since littler is used by the tests and examples.
It also sets executable permissions on all scripts with a shebang.


thirdwing:
Hi, @ellert 

Thank you for the contribution. I think we don't need `littler` in the suggestions, since all testing cases are handle by `RUnit`.

eddelbuettel:
@thirdwing is correct.

Think this through:  if the Depends were missing, how could this have been on CRAN for 8 years, or have passed 1100+ tests on Travis?

Also, please consider taking a look at [Contributing.md](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) as me think it is more efficient if you start with an issue ticket describing what you'd want to change.

ellert:
OK, I get that the littler is not strictly needed. This is why I proposed adding it to suggests, not requires.

The reason for adding it was intended as a way to avoid the confusion about /usr/bin/r. When I saw the /usr/bin/r it really looked like a typo for /usr/bin/R - something that could happen if the person writing the file was using a case insensitive file system - which is still common on e.g. MacOS X. With a reference in the DESCRIPTION file to a package that actually provides /usr/bin/r it could possibly help others not having the same wrong idea that I had. But of course this is not crucial, and if you think it is a bad idea that is perfectly fine. And it was not only about the tests - also many of the example files use /usr/bin/r.

What about the other commit in this pull request, setting the executable permissions on the files having a shebang. This was in my opinion the more important of the two, and the title of the the request was even named "Permissions". A shebang doesn't make much sense without the executable permission set. Even many packaging sanity check tools complain about it, e.g.:

$ rpmlint R-Rcpp
R-Rcpp.x86_64: E: non-executable-script /usr/lib64/R/library/Rcpp/discovery/cxx0x.R 644 /bin/env

eddelbuettel:
Regarding

1. modes:  I get the same warning on Debian and [override them](https://github.com/RcppCore/Rcpp/blob/master/debian/r-cran-rcpp.lintian-overrides).  Strictly speaking most of these do not need the shebang so we haven't cared.  Maybe I clean it up one day.

2. indentation:  Hundreds of files specify it in the Emacs header line.  We have a Emacs style file in the repo.  We have a clang format file.  But we not doctrinaire about and don't force reformat all files. (Well @kevinushey did once but we no longer talk about that <grin>).  Use common sense, look at other files, particularly newer ones.  An 'R Core-alike' style with four spaces is common in R and C(++).

Finally please DO file issues before PRs.

PR #443, opened 2016-03-04
Feature/switch to usr bin env r (closes #440)

### LINKED ISSUE #440: Use /usr/bin/env r
In the past we used `#!/usr/bin/r -ti` to pass switches such as `-t` or `-i`.  These were, as I recall, needed with the littler version used then -- we should no longer need this.

As such, we may as well use `#!/usr/bin/env r` and I'll start changing a few of the scripts (and will also overhaul them as I go along).

eddelbuettel:
This is mostly cleanup along the lines of the earlier (but too large) PR #438 by @ellert

I went over all examples and updated where needed; they now all use `#!/usr/bin/env r`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,16 @@
+2016-02-21  Dirk Eddelbuettel  <edd@debian.org>
+
+        * inst/examples/functionCallback/newApiExample.r (vecfunc): Switched to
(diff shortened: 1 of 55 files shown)


eddelbuettel:
There isn't anything too controversial here so I'll just fold this in now.

PR #444, opened 2016-03-05
Feature/pr437 postprocessing

eddelbuettel:
Passes cleanly with `g++-6 -Wall -pendantic` as per #437


eddelbuettel:
There is nothing super scary here so I'll just fold this in now.

PR #445, opened 2016-03-09
One character typo: d/q/p/q -> d/q/p/r

csgillespie:
Also: 
 * Re-order example to match: d/q/p/r
 * Fix odd indenting of foo function
 * Remove rogue additional line


eddelbuettel:
Thank you -- looks good.

Issue #407, opened 2015-12-07
Minimal cbind sugar function


nathan-russell:
I've been working on an implementation for a `cbind` sugar function, but since my current design lacks some of the features of `base::cbind` I'm not sure it's on par with the rest of the `Rcpp::sugar` functions, and was hoping to get some feedback and / or suggestions for improving it. The source can be viewed in [this gist](https://gist.github.com/nathan-russell/068b8a459833609c52c2) (I didn't want to dump 300 lines of code into this ticket), but to summarize the pros and cons: 

##### Good: 
+ Handles `Matrix`/`Matrix`, `Matrix`/`Vector`, `Matrix`/`scalar`, `Vector`/`Vector`, `Vector`/`Matrix`, `Vector`/`scalar`, `scalar`/`Matrix`, `scalar`/`Vector` argument pairs 
+ Does appropriate checking for equal number of rows (or length(s), in the case of `Vector`s) 

##### Shortcomings: 
+ Does not handle arbitrary number of arguments in a single function call, as in `base::cbind`'s `...` argument (I'm not sure how to address this without dipping into C++11, or if it's even possible in C++98) , although nested calls seem to work correctly (e.g. `cbind(cbind(cbind(arg1, arg2), arg3), arg4)`)
+ The `dimnames` attribute is not preserved 
+ Does not handle `DataFrame`s  

From the testing I've done so far, the supported features above correctly mimic the behavior of `base::cbind`, but I'm not sure what the general consensus is on including less-than-comprehensive sugar functions in the package. Is it worthwhile to have a version of `cbind` with limited functionality? And of course, I'd be more than happy to look into any suggestions for improving this.

eddelbuettel:
Nice!  And as a real quick reply: could we not 'trick it' by macro expansion to provide 2, 3, 4, 5, .... arguments?  I can live without dimnames, and support for data.frame objects...

kevinushey:
I get a compile error + a segfault in the following example:

```cpp
// [[Rcpp::export]]
NumericMatrix my_cbind(NumericVector x, NumericVector y) {
  return cbind(x, y);
}

/*** R
my_cbind(1, 2)
*/
```

```
> Rcpp::sourceCpp("Untitled.cpp")
Untitled.cpp:119:11: warning: binding reference member 'lhs' to a temporary value [-Wdangling-field]
    : lhs(lhs_), rhs(rhs_)
          ^~~~
[...]

PR #447, opened 2016-03-13
Sugar cbind with unit tests

nathan-russell:
The latest `cbind` implementation as discussed toward the bottom of #407.

### DIFF
--- a/inst/include/Rcpp/sugar/functions/cbind.h
+++ b/inst/include/Rcpp/sugar/functions/cbind.h
@@ -1,0 +1,830 @@
+// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-
+//
+// cbind.h: Rcpp R/C++ interface class library -- cbind
+//
+// Copyright (C) 2016 Nathan Russell
+//
+// This file is part of Rcpp.
+//
+// Rcpp is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// Rcpp is distributed in the hope that it will be useful, but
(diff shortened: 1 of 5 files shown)


eddelbuettel:
Looks pretty good to me.

inst/include/Rcpp/sugar/functions/cbind.h
```diff
@@ -0,0 +1,828 @@
...
+struct matrix_return
+    : public detail::matrix_return<T, B> {};
+
+template <typename T>
+struct matrix_return<T, false>
+    : public detail::matrix_return<T, false> {};
+
+} // cbind_impl
+
+#define MakeBindable(x) (cbind_impl::detail::MakeBindable(x)(x))
```
kevinushey: Should this macro get undefined after we're done using it?

kevinushey:
Looks great! One minor comment re: potentially cleaning up the `MakeBindable` macro. Awesome stuff!

nathan-russell:
@kevinushey Thanks for reviewing! I just committed the change.

eddelbuettel:
Something went marginally wrong with the commit as GH thinks you changed 830 lines...  Not to worry though.

nathan-russell:
Oh dear. I made this commit from a Windows machine -- I wonder if that has anything to do with it.

eddelbuettel:
That would do it.  Hint:  I believe even Windoze has some editors that behave like grownups :)

Here it doesn't matter.  It's a new file anyway.  In general, we've been pretty careful about not mass-committing whitespace changes.  (With exception: @kevinushey once did a mass commit as I recall; I recently updated a bunch of headers following @ellert's PR, .... -- but in general that is best avoided).

We do have a number of skeletons with mixed tab/spaces in the closet though.  But no point going crazy about them just for the heck of it...

I'll merge this now.  Nice work!

Issue #448, opened 2016-03-21
Rcpp compilation issue on Alpine distribution


lostie:
```
"Command failed (1)

Failed to run system command:

'/usr/lib/R/bin/R' --vanilla CMD INSTALL '/tmp/RtmpDPDGNf/Rcpp' --library='/app/packrat/lib/x86_64-alpine-linux-musl/3.2.3' --install-tests --no-docs --no-multiarch --no-demo
The command failed with output:
* installing *source* package 'Rcpp' ...
** package 'Rcpp' successfully unpacked and MD5 sums checked
** libs
x86_64-alpine-linux-musl-g++ -I/usr/lib/R/include -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer  -c Date.cpp -o Date.o
x86_64-alpine-linux-musl-g++ -I/usr/lib/R/include -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer  -c Module.cpp -o Module.o
x86_64-alpine-linux-musl-g++ -I/usr/lib/R/include -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer  -c Rcpp_init.cpp -o Rcpp_init.o
x86_64-alpine-linux-musl-g++ -I/usr/lib/R/include -DNDEBUG -I../inst/include/ -Os -fomit-frame-pointer    -fpic  -Os -fomit-frame-pointer  -c api.cpp -o api.o
api.cpp:40:30: fatal error: execinfo.h: No such file or directory
compilation terminated.
/usr/lib/R/etc/Makeconf:141: recipe for target 'api.o' failed
make: *** [api.o] Error 1
ERROR: compilation failed for package 'Rcpp'
* removing '/app/packrat/lib/x87_64-alpine-linux-musl/3.2.3/Rcpp'"
```

It seems related to https://github.com/RcppCore/Rcpp/issues/360 (although on a diferrent platform).

The code it is hitting seems to be https://github.com/RcppCore/Rcpp/blob/0.12.3/src/api.cpp#L40 

According to the [gnulib page](https://www.gnu.org/software/gnulib/manual/html_node/execinfo_002eh.html) it seems that for platforms that the file is missing it should only have a stub (do nothing).

> Alpine Linux is a security-oriented, lightweight Linux distribution based on musl libc
[...]

PR #449, opened 2016-03-21
Add support for MUSL based Operation Systems

lostie:
Addresses https://github.com/RcppCore/Rcpp/issues/448

**Final solution:** After some discussion we decided the best way would be to add a `__MUSL__` test and get Alpine distributions to add that to the `PKG_CXXFLAGS `

~~Using suggestion recommend by the `musl` F.A.Q.:~~ (Read below)

> Q: application XY misbehaves or crashes at runtime when linked against musl
A: Usually this is because the app has hardcoded glibc-specific assumptions or wrong #ifdefs. See Functional differences from glibc. The most common causes are expectations of gnu getopt behaviour, iconv usage on UCS2 with assumptions that BOM is processed and the byte order detected, assuming that off_t is 32 bit, and assumptions that pthread_create will create sufficiently large stacks by default (crash situation). A good approach to solving the issue is to watch out for #if and #ifdefs in the code and placing some debug #warning there to see which code paths are enabled by the preprocessor. Often the logic taken by #ifdef's is to check a blacklist of preprocessor defines #if __sun__ || __haiku__ || __irix__ || __qnx__ portable_code(); #else glibc_code() #endif. The logic should be the other way round: #ifndef __GLIBC__ portable_code() #else glibc_code(). There's also no point in checking __GLIBC__ version numbers without first making sure that __GLIBC__ is defined at all.

http://wiki.musl-libc.org/wiki/FAQ

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -34,26 +34,21 @@ using namespace Rcpp;
 #endif
 
 #if defined(__GNUC__)
-    #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX)
+    #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX) || defined(__MUSL__)
         // do nothing
     #else
-        #ifndef __GLIBC__
-          // do nothing
-        #else
-          #include <execinfo.h>
+        #include <execinfo.h>
 
-          static std::string demangler_one(const char* input) {
-              static std::string buffer;
-              buffer = input;
-              buffer.resize(buffer.find_last_of('+') - 1);
-              buffer.erase(
-                  buffer.begin(),
-                  buffer.begin() + buffer.find_last_of(' ') + 1
-              );
-              return demangle(buffer);
-          }
-        #endif
-
+        static std::string demangler_one(const char* input) {
+            static std::string buffer;
+            buffer = input;
+            buffer.resize(buffer.find_last_of('+') - 1);
+            buffer.erase(
+                buffer.begin(),
+                buffer.begin() + buffer.find_last_of(' ') + 1
+            );
+            return demangle(buffer);
+        }
     #endif
 #endif
 
@@ -261,7 +256,7 @@ SEXP rcpp_can_use_cxx11() {
 // [[Rcpp::register]]
 SEXP stack_trace(const char* file, int line) {
     #if defined(__GNUC__)
-        #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX)
+        #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX) || defined(__MUSL__)
             // Simpler version for Windows and *BSD
             List trace = List::create(_["file"] = file,
                                       _[ "line"  ] = line,
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -37,18 +37,22 @@ using namespace Rcpp;
     #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX)
         // do nothing
     #else
-        #include <execinfo.h>
+        #ifndef __GLIBC__
+          // do nothing
+        #else
+          #include <execinfo.h>
 
-        static std::string demangler_one(const char* input) {
-            static std::string buffer;
-            buffer = input;
-            buffer.resize(buffer.find_last_of('+') - 1);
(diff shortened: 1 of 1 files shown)


kevinushey:
This unfortunately does not compile on OS X (tested on El Capitan).

kevinushey:
Perhaps instead we need to guard this in https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/platform/compiler.h, with something like

```
#define RCPP_DEMANGLING_SUPPORT
#ifdef __ALPINE
# undef RCPP_DEMANGLING_SUPPORT
#endif
```

and then guard based on that?

thirdwing:
It failed because another change is missed. No `__GLIBC__` is defined on OSX. See https://github.com/thirdwing/Rcpp/commit/8db55e08dc7aa42bbd2841754c488827c000d0b2

@kevinushey From the FAQ (http://wiki.musl-libc.org/wiki/FAQ#Q:_why_is_there_no_MUSL_macro_.3F), it seems no way to detect usage of `musl`.

eddelbuettel:
We could put a else-if in there and have the Alpine user (at least for now) define that themselves...

kevinushey:
@thirdwing but backtraces _do_ currently work on OS X, so there must be a more scoped / targeted change. I loathe the idea of making a configure script to do it, though.

thirdwing:
@kevinushey Yes, OSX doesn't define `__GLIBC__` and it does have `execinfo.h`.

The only idea I have now is to detect `execinfo.h` in the makefile and add something like `-DHAVE_EXECINFO` into `PKG_CPPFLAGS`.

eddelbuettel:
As I said before, add a test for `__ALPINE`  [here](https://github.com/RcppCore/Rcpp/blob/0.12.3/src/api.cpp#L37) and have the Alpine folks add `-D__ALPINE=1` to their PKG_CXXFLAGS.  This will work for now until they do us the pleasure of defining something in Musl.

I have no intention of breaking OS X, and no appetite for byzantine build mechanisms.

thirdwing:
Totally agree.

On Mon, Mar 21, 2016 at 3:04 PM, Dirk Eddelbuettel <notifications@github.com
> wrote:

> As I said before, add a test for __ALPINE here
> <https://github.com/RcppCore/Rcpp/blob/0.12.3/src/api.cpp#L37> and have
> the Alpine folks add -D__ALPINE=1 to their PKG_CXXFLAGS. This will work
> for now until they do us the pleasure of defining something in Musl.
>
> I have no intention of breaking OS X, and no appetite for byzantine build
> mechanisms.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/RcppCore/Rcpp/pull/449#issuecomment-199427344>
>



-- 
Qiang Kou
qkou@umail.iu.edu
School of Informatics and Computing, Indiana University

lostie:
> As I said before, add a test for __ALPINE here and have the Alpine folks add -D__ALPINE=1 to their PKG_CXXFLAGS. This will work for now until they do us the pleasure of defining something in Musl.
>
> I have no intention of breaking OS X, and no appetite for byzantine build mechanisms

I have added a `__MUSL__` test instead of `__ALPINE` as the issue is directly related with the `musl` glib and [more linux distributions make use of it](http://wiki.musl-libc.org/wiki/Projects_using_musl), not just Alpine.

eddelbuettel:
That looks much better.  And I presume it works for you, ie `__MUSL__` is in fact defined?

(And please don't get me wrong -- I have nothing against Musl or Alpine. I use Docker quite a bit and have quite some sympathy for thinner builds but we got to this right and not break a bazillion existing use cases ...)

lostie:
> That looks much better. And I presume it works for you, ie __MUSL__ is in fact defined?

Yes!

For testing purposes (and given I'm using docker I have added the `-D__MUSL__` to the `R` compilation options (`PKG_CXXFLAGS `):

```shell
x86_64-alpine-linux-musl-g++ -I/usr/lib/R/include -DNDEBUG -D__MUSL__ -I../inst/include -DCOMPILING_DPLYR -Os -fomit-frame-pointer -I"/app/packrat/lib/x86_64-alpine-linux-musl/3.2
.3/Rcpp/include" -I"/app/packrat/lib/x86_64-alpine-linux-musl/3.2.3/BH/include"   -fpic  -Os -fomit-frame-pointer  -c RcppExports.cpp -o RcppExports.o
```

but once Alpine adds that to the `PKG_CXXFLAGS` it should work out of the box :+1:

eddelbuettel:
Err, yes, I understand the `PKG_CXXFLAGS` trick; after all it was me who suggested it to you :wink:

But what is missing here is someone checking with the good folks of the Musl library **to please add such an unmistakable identifier** as we cannot possibly do this piecemeal package by package.

lostie:
> But what is missing here is someone checking with the good folks of the Musl library to please add such an unmistakable identifier as we cannot possibly do this piecemeal package by package.
>
> So let me rephrase: Can you please run a recursive grep over the C library headers for Musl and check that __MUSL__ is there?

Here's the list of macros: https://gist.github.com/lostie/2b60cb4afb3bbba0b928

__MUSL__ is not present as stated in their F.A.Q.

All I can do is request MUSL community (I've already sent them an e-mail) to set by default in their library.

eddelbuettel:
Yes, please liaise with them. They will need this.

I'll fold this in now.  We have Rcpp 0.12.4 pending at CRAN so with be for the next release anyway.

Issue #452, opened 2016-03-28
Rcpp::Nullable<Rcpp::String> problems


asieira:
When I run this with a string, I get an error message saying `Error: 'getCharCE' must be called on a CHARSXP`:
```
// [[Rcpp::export]]
String test_nullable(Rcpp::Nullable<Rcpp::String> param = R_NilValue) {
  if (param.isNotNull())
    return String(param);
  else
    return String("");
}
```

This, however, works perfectly:
```
// [[Rcpp::export]]
CharacterVector test_nullable2(Rcpp::Nullable<Rcpp::CharacterVector> param = R_NilValue) {
  if (param.isNotNull())
    return CharacterVector(param);
  else
    return CharacterVector();
}
```

Is there a principled reason why `Rcpp::Nullable` should not work with the `String` class so it enforces a scalar value?

dcdillon:
How are you calling `test_nullable()`.  Can you add the R code?

asieira:
There you go:
```
> cppFunction('String test_nullable(Rcpp::Nullable<Rcpp::String> param = R_NilValue) {
+   if (param.isNotNull())
+     return String(param);
+   else
+     return String("");
+ }')
> test_nullable()
[1] ""
> test_nullable("blah")
Error in test_nullable("blah") : 
  'getCharCE' deve ser chamado em uma CHARSXP

Enter a frame number, or 0 to exit   

1: test_nullable("blah")

Selection: 
```

dcdillon:
Will have to look into this some more.  I'm sure it's an easy fix, but on the surface looks like a bug to me.

asieira:
:bow:

eddelbuettel:
Strings are really complicated.  We set up `Nullable` to work with `SEXP` types, so I suggest you do something like this instead:
```C++
#include <Rcpp.h>

using namespace Rcpp;

// [[Rcpp::export]]
std::string test_nullable(Nullable<CharacterVector> param = R_NilValue) {
  if (param.isNotNull()) {
    CharacterVector s(param);
    return std::string(s[0]);
  } else
    return std::string("");
}
```

This seems to work:

```r
R> sourceCpp("/tmp/nullstr.cpp")
R> test_nullable("abc")               
[1] "abc"
R> test_nullable(NULL)                       
[1] ""
R> 
```

kevinushey:
[...]

PR #453, opened 2016-03-29
Fixing issue with Rcpp::String and Rcpp::Nullable

dcdillon:
This addresses #452 in the manner suggested by @thirdwing.

Complete with unit tests.

### DIFF
--- a/inst/include/Rcpp/String.h
+++ b/inst/include/Rcpp/String.h
@@ -71,16 +71,35 @@ namespace Rcpp {
         }
 
         /** construct a string from a single CHARSXP SEXP */
-        String(SEXP charsxp) : data(charsxp), valid(true), buffer_ready(false), enc(Rf_getCharCE(charsxp)) {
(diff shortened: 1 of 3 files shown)


eddelbuettel:
I see. By setting a few of the member 'later' after having passed some sanity checks, we survive the corner case.  Nice.

dcdillon:
I would, however, suggest an rdepends check on this one as it messes with some fundamental constructors.

thirdwing:
LGTM. Just a remainder, please close the issue when merging.

eddelbuettel:
I'll merge this now and turn on a rev dep check.  If it barks we can still revert.

eddelbuettel:
And just for the record the reverse depends worked out [just fine](https://github.com/RcppCore/rcpp-logs/commit/92a38c97cb55eb3d0c318f0c2923f15d245574bd).  

Thanks again for the fix.

PR #464, opened 2016-04-15
reverting PR 454 for copyObject and underlying copy constructor

eddelbuettel:


jjallaire:
@romainfrancois We've had a bunch of problems with R CMD check and the new Windows C++11 toolchain w/ the original PR. Since CRAN is right in the window where they are deciding whether to take the C++11 toolchain or wait until next year we thought it best to roll this back to smooth the way for CRAN, then take it up again after the R release window.

jjallaire:
@kevinushey You are more familiar with the mechanics of this than I am. Could you do a quick additional review then merge?

PR #474, opened 2016-05-14
first crack at checking compiler versions from R

eddelbuettel:
currently limited to g++

works reasonably well -- log from my Ubuntu 15.10 box below

```
edd@max:~$ g++ --version | head -1                                  # default
g++ (Ubuntu 5.2.1-22ubuntu2) 5.2.1 20151010
edd@max:~$ Rscript -e 'Rcpp:::compilerCheck("5.2.1")' 
[1] TRUE
edd@max:~$ Rscript -e 'Rcpp:::compilerCheck("4.6.0")' 
[1] TRUE
edd@max:~$ Rscript -e 'Rcpp:::compilerCheck("5.3.1")'               # what 16.04 has
[1] FALSE
edd@max:~$ CXX=g++-4.5 Rscript -e 'Rcpp:::compilerCheck("4.6.0")'   # old version
[1] FALSE
edd@max:~$ 
```


JanMarvin:
```c++
int gccver(){
  return(  __GNUC__ * 10000 +  \
    __GNUC_MINOR__ * 100     +  \
    __GNUC_PATCHLEVEL__);
}
```

Isn't this what you want?

eddelbuettel:
@JanMarvin:  Thanks. In fact, @dcdillon made a similar point ("just compile `g++ -std=c+11 hello.cpp`").  But I wanted something a little simpler and quicker.  

We do actually have some tests in Rcpp already:

- `inst/include/Rcpp/platform/compiler.h` tests for a variety of features needed around/before TR1
- `inst/discovery/cxx0x.R` is an old (and outdated) check for minimal standards version

so I don't think the last word has been said.   Right now it is more about helping users _right at the start_ of a build process, ie in `configure`.  We had a number of unfortunate error reports when folks on older RHEL/CentOS attempted to build RcppArmadillo, and of course flamed out fabulously as `g++-4.4` doesn't even buy you half a cup of joe anymore.

eddelbuettel:
I'm going to merge this in, it is unlikely to cause harm as it isn't even exported yet.

jjallaire:
It would be cool to have a variation of this which could be called from Makevars and which throws an error when the minimum compiler version isn't met. The error would have a nice clear message indicating that the issue is the compiler version.

eddelbuettel:
Certainly.  The bigger question really is whether we want

- full-blown compiler-based tests which actually test particular compiler features (a la `platform/compiler.h` in the sources)
- something lightweight to throw decent error messages either via Makevars (but then everybody has to modify `Makevars` again)
- or just build on this and call via `Rscript` from a shell script `configure` which people call?

Making the error message nicer/more informative is easy.  I was just happy when I realized that I could test `g++` as well as the common environment variables _from R_ and use R to compare versions strings correctly.   What we have here is a start, and possibly 'good enough' to replace what I just did in RcppArmadillo.

jjallaire:
I think that #2 (have folks add a line to Makevars) strikes the best
balance, especially since everyone already has a Makevars and this is
pretty much a boilderplate one-liner.

eddelbuettel:
How do you suggest to hook it in?  It is not just an expansion as `PKG_CXXFLAGS` ?

Or expand to something which gets called and remains silent on success but emits loud and clear error messages on failure?

jjallaire:
Yeah, I'd say silent on success and loud and clear error messages on
failure.

eddelbuettel:
I thought about that while going for a run and I am not so sure we can 'break' cleanly there.  It is R calling make calling our (Rscript-based ?) snippet.  I still think `configure` is cleaner.  To be seen...

PR #476, opened 2016-05-14
Fixing ambiguous arithmetic operators for Matrix/Vector on 32-bit

dcdillon:
So the crux of this is it makes templates that match exactly the function calls that happened.  I achieved this by adding another template parameter to each of the arithmetic operators for for `Vector` and `Matrix` and then restricting the values allowed for this template parameter using `enable_if`.

Additionally, the same ambiguities can be exposed with the following C++ code on 64-bit (though it's unlikely anyone would have done this):

```c++
#include <Rcpp.h>

void testFunction(Rcpp::NumericVector x, long l)
{
        Rcpp::NumericVector y = x + l;
}

void testFunction2(Rcpp::NumericMatrix x, long l)
{
        Rcpp::NumericMatrix y = x + l;
}
```
@eddelbuettel, @kevinushey, @thirdwing please have a look.

### DIFF
--- a/inst/include/Rcpp/sugar/operators/divides.h
+++ b/inst/include/Rcpp/sugar/operators/divides.h
@@ -409,19 +409,19 @@ namespace sugar{
 }
 
 template <int RTYPE,bool NA, typename T, typename U>
-inline typename traits::enable_if< traits::is_primitive< typename traits::remove_const_and_reference<U>::type >::value, sugar::Divides_Vector_Primitive< RTYPE , NA, T > >::type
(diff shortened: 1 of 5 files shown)


dcdillon:
Hmm...perhaps it should be `traits::is_convertible<T, typename traits::storage_type<RTYPE>::type>` instead of `traits::is_primitive`.

Will do that if people agree.

thirdwing:
Just confirmed this fixed 64-bit and 32-bit situations.

Let's merge this if it doesn't break any other packages.

eddelbuettel:
We just got Rcpp 0.12.5 onto CRAN so we have a moment -- but yes, will rev.dep this.

eddelbuettel:
With the various 'LGTM' having come in I'll just merge, turn on a rev.dep run -- and have the option to revert later today should it be necessary.

Issue #477, opened 2016-05-15
const'ness (constivity?) in sugar::min/max


dcdillon:
The `sugar::min` and `sugar::max` classes (which to my mind should just be functions anyhow, but leaving that) have a cast operator to the typedef'ed `STORAGE` type.  This operator is not defined as `const` because it actually modifies some `private` member variables.  This is causing a problem with the latest fix for operator ambiguity in `Matrix` or `Vector` arithmetic operators.

I see no problem in moving these `private` member variables simply into the cast operator function and declaring the function as `const`.  Does anyone else see any pitfalls?

I would propose changing:

```c++
template <int RTYPE, bool NA, typename T>
class Max {
public:
    typedef typename Rcpp::traits::storage_type<RTYPE>::type STORAGE ;

    Max( const T& obj_) : obj(obj_) {}

    operator STORAGE() {
        max_ = obj[0] ;
        if( Rcpp::traits::is_na<RTYPE>( max_ ) ) return max_ ;

        R_xlen_t n = obj.size() ;
        for( R_xlen_t i=1; i<n; i++){
            current = obj[i] ;
            if( Rcpp::traits::is_na<RTYPE>( current ) ) return current;
            if( current > max_ ) max_ = current ;
        }
        return max_ ;
    }

private:
    const T& obj ;
    STORAGE min_, max_, current ;
} ;
```

to

```c++
template <int RTYPE, bool NA, typename T>
class Max {
public:
    typedef typename Rcpp::traits::storage_type<RTYPE>::type STORAGE ;

    Max( const T& obj_) : obj(obj_) {}

    operator STORAGE() const {
        STORAGE min_, max_, current ;
        max_ = obj[0] ;
        if( Rcpp::traits::is_na<RTYPE>( max_ ) ) return max_ ;

        R_xlen_t n = obj.size() ;
        for( R_xlen_t i=1; i<n; i++){
            current = obj[i] ;
            if( Rcpp::traits::is_na<RTYPE>( current ) ) return current;
            if( current > max_ ) max_ = current ;
        }
        return max_ ;
    }

private:
    const T& obj ;
} ;
```

Alternatively, we could greatly simplify `min/max` by simply defining:

```c++
namespace Rcpp
{

template< int RTYPE, typename T >
typename traits::storage_type< RTYPE >::type max(const VectorBase< RTYPE, true, T > &x) {
    typedef typename Rcpp::traits::storage_type<RTYPE>::type STORAGE;

[...]

PR #478, opened 2016-05-16
Fixing const'ness of min and max

dcdillon:
This is a middle of the road in intrusiveness fix for the const'ness of `Rcpp::min` and `Rcpp::max` to address #477.


eddelbuettel:
Very nice, thank you.

kevinushey:
Looks good to me as well -- thanks!

eddelbuettel:
Ran another rev.dep overnight with it, all good.  Failures still at seven with 655 good, but the seven are all in tests with testthat throwing up a number of them.  

I guess we should glance at the other sugar functions to see if any need `const` variants.

PR #479, opened 2016-05-16
Fix the string replace

thirdwing:
The CRAN version:

```cpp
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
String replace() {
  String foo("abba");
  return foo.replace_first("ba", "BA");
}

/*
> Rcpp::sourceCpp("test.cpp")
> replace()
[1] "BAba"
*/
```

After fix:

```r
> Rcpp::sourceCpp("test.cpp")
> replace()
[1] "abBA"
```

@kevinushey

### DIFF
--- a/inst/include/Rcpp/String.h
+++ b/inst/include/Rcpp/String.h
@@ -262,7 +262,8 @@ namespace Rcpp {
             RCPP_STRING_DEBUG_2("String::replace_first(const char* = '%s' , const char* = '%s')", s, news);
             if (is_na()) return *this;
             setBuffer();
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Looks good first glance.

kevinushey:
Looks good to me as well -- I kind of wish we could get away without copying the C strings into `std::string`s, but I'm fine with the PR as-is.

kevinushey:
Thanks for doing the legwork @thirdwing!

PR #485, opened 2016-05-31
Trivial: "sire" to "size"

wibeasley:
Also, Would "length" be less ambiguous, in case a programmer is thinking in bytes for "size"?


eddelbuettel:
Thanks.  And for this sentence length would have been fine too,  but we may as well leave this now.

PR #487, opened 2016-06-01
Annotate stop() with NORET

### LINKED ISSUE #486: How to teach the compiler that stop() doesn't return?
So that we can write

```
bool foo(int x) {
  switch (x) {
    case 1: return true;
    case 2: return false;
    default: stop("Don't know this x");
  }
}
```

without issuing a warning.

There is a "noreturn" attribute for [GCC](https://gcc.gnu.org/onlinedocs/gcc-3.4.3/gcc/Function-Attributes.html) and [clang](http://clang.llvm.org/docs/LanguageExtensions.html), but there might be other solutions. **EDIT**: http://stackoverflow.com/a/25408952/946850 is a nice SO answer that features `__builtin_unreachable()`.

Re https://github.com/hadley/dplyr/pull/1878#issuecomment-223140346, CC @hadley.

krlmlr:
per @kevinushey's suggestion. Fixes #486.

Closes #488 (=includes it).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2016-06-02  Kirill Müller <krlmlr@mailbox.org>
+
+        * inst/include/Rcpp/algorithm.h: Use "long long" only if available
+        * inst/include/Rcpp/exceptions.h: Annotate stop() with NORET
+
 2016-05-18  Daniel C. Dillon  <dcdillon@gmail.com>
 
         * inst/include/Rcpp/algorithm.h: New approach for sugar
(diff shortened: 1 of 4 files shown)


eddelbuettel:
I believe this is (once again) an auto-generated file but with the rushed way in which this repo got moved from svn we may have lost the connection to the generator script.  It may still be in the rcpp-logs repo.

Can you please add a ChangeLog entry and possibly a line for NEWS.Rd and then I'd be happy to merge.

krlmlr:
Done. I merged #488 here and added both NEWS entries.

eddelbuettel:
Thanks so much, looks great.

PR #488, opened 2016-06-01
Use long long only if available

krlmlr:
Compiling with -pedantic gives me warnings about "long long" otherwise.


eddelbuettel:
I was fearful when I saw the subject/title but it looks very much like this does the right thing.  Will merge later.

So thanks!

krlmlr:
I wonder if we want to minimize the number of `#ifdef`-s by regrouping.

eddelbuettel:
I'd say minimize changes so leave it.  But ChangeLog and NEWS.Rd would be nice.

PR #492, opened 2016-06-13
Improve na_omit for vectors without NA

artemklevtsov:
If lengths of the input and output vectors are equal simply return input vector. Also create out vector without initialization (`no_init`).

Comparison old and new version of the `na_omit`: https://gist.github.com/artemklevtsov/38ce7eda7d6eecead5582fae997691b4


eddelbuettel:
Thanks for this.  It looks reasonable.  But as you mention the testing gist, would you mind adding one, tow, or more unit tests to Rcpp to test this?  

An addition to the `ChangeLog` file would also be welcome.

eddelbuettel:
Looks good, will try to examine more closely this eve.

PR #493, opened 2016-06-17
Add RCPP_USING_UTF8_ERROR_STRING macro to use UTF-8 encoding exception string in R.

qinwf:
Hi, I posted on Rcpp mailing list last week about [Rcpp exception with UTF-8 strings on Windows](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2016-June/009291.html).

This PR adds RCPP_USING_UTF8_ERROR_STRING to use UTF-8 encoding exception string in R.

Here is an example for this on Windows.

By default, not using RCPP_USING_UTF8_ERROR_STRING,

```r
sourceCpp(code = "
#include <Rcpp.h>
          
// [[Rcpp::export]]
void error_with_unicode(std::string input){
    Rcpp::stop(input);
}
")
res = try(error_with_unicode(enc2utf8("测试")))
#> Error : 娴嬭瘯
Encoding(res)
#> [1] "unknown"
```

Using RCPP_USING_UTF8_ERROR_STRING,

```r
sourceCpp(code = "
#define RCPP_USING_UTF8_ERROR_STRING
#include <Rcpp.h>

// [[Rcpp::export]]
void error_with_unicode(std::string input){
    Rcpp::stop(input);
}
")
res = try(error_with_unicode(enc2utf8("测试")))
#> Error : 测试
Encoding(res)
#> [1] "UTF-8"
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2016-06-20  Qin Wenfeng <mail@qinwenfeng.com>
+
+        * inst/include/Rcpp/exceptions.h: add RCPP_USING_UTF8_ERROR_STRING macro to use UTF-8 encoding exception string in R
+
 2016-06-14  Artem Klevtsov <a.a.klevtsov@gmail.com>
 
         * inst/include/Rcpp/sugar/functions/na_omit.h: Improve na_omit for
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -14,6 +14,7 @@
       \ghit{486}).
       \item String replacement was corrected (Qiang in \ghpr{479} following
       mailing list bug report by Masaki Tsuda)
+      \item Add RCPP_USING_UTF8_ERROR_STRING macro to use UTF-8 encoding exception string in R (Qin Wenfeng in \ghpr{493})
     }
     \item Changes in Rcpp Sugar:
     \itemize{
--- a/inst/include/Rcpp/exceptions.h
+++ b/inst/include/Rcpp/exceptions.h
@@ -137,7 +137,12 @@ inline SEXP get_last_call(){
 
 inline SEXP get_exception_classes( const std::string& ex_class) {
     Rcpp::Shield<SEXP> res( Rf_allocVector( STRSXP, 4 ) );
+    
+    #ifndef RCPP_USING_UTF8_ERROR_STRING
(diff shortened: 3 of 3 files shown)


eddelbuettel:
Thanks.  The PR is nicely minimal, and even contains ChangeLog and NEWS.  Tests would be nice too.

As the [Contributing.md](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) file suggests, we generally prefer discussion _prior_ to submission but we'll try to review in due course.

kevinushey:
I feel like we shouldn't even use this macro define and _always_ create our strings with UTF-8 encoding. (Ie, we shouldn't use `mkChar()`; rather, always `mkCharLenCE()` with UTF-8 encoding as in this PR.

eddelbuettel:
Not a bad idea.  Any possible side effects?  Sounds like I may need a rev.deps run for this.

thirdwing:
I agree with Kevin. I will try to replace `mkChar` this weekend.

BTW, it seems the error fixed by this PR can only be reproduced on Windows.

eddelbuettel:
Right on.  

But ... didn't we once try this and failed with a different / older PR?  

[ Goes off looking ... ]  

What about #280 by @thirdwing ?

thirdwing:
In the old PR, I try to convert strings into UTF-8 when passing strings from R to C++.

Now, this error happens when we throw strings from C++ to R...

A good solution is needed for encoding problems...

kevinushey:
FWIW, this PR is a bit more targeted in that we only change the representation of strings returned to R, not _all_ Rcpp `String` objects (which may or may not be UTF-8 encoded). Because of that, I think this is safer than #280.

eddelbuettel:
Yep, yep. And that we could indeed make unconditionally UTF-8.

qinwf:
We can remove the macro and make it unconditionally UTF-8, and this may change error messages of some existing packages.

Or leave the macro as now, and developers will only be able to change the result with the macro in the new code.

eddelbuettel:
That is a good point.

eddelbuettel:
I think I'll merge as-is.  We can still discuss

- actually setting the `#define` ourselves (either now or later) to make it a default
- revert the logic, ie have UTF-8 the default with an opt-out for packages that cannot/will not change

but the current suggestion is good as it is pretty minimal.

kevinushey:
Sounds good to me.

Issue #498, opened 2016-06-29
eliminate unsafe usages of 'Rf_eval'


kevinushey:
There are a number of places in Rcpp where we use `Rf_eval` rather than the more appropriate / safer `Rcpp_eval`:

```
$ ag -Q "Rf_eval"
inst/include/Rcpp/api/meat/Rcpp_eval.h
57:    Shield<SEXP> res(::Rf_eval(call, R_GlobalEnv));
69:            Shield<SEXP> conditionMessage(::Rf_eval(conditionMessageCall, R_GlobalEnv));

inst/include/Rcpp/Environment.h
114:                res = Rf_eval( res, env ) ;
135:                res = Rf_eval( res, env ) ;
206:                    Rf_eval( call, R_GlobalEnv ) ;

inst/include/Rcpp/exceptions.h
132:    Rcpp::Shield<SEXP> calls( Rf_eval( sys_calls_expr, R_GlobalEnv ) );
197:    Rcpp::Shield<SEXP> simpleError( Rf_eval(simpleErrorExpr, R_GlobalEnv) );
326:    Rf_eval( expr, R_GlobalEnv ) ;

inst/include/Rcpp/Language.h
148:            return Rf_eval( Storage::get__(), R_GlobalEnv) ;
151:            return Rf_eval( Storage::get__(), env) ;

inst/include/Rcpp/macros/macros.h
55:       Rf_eval( expr, R_GlobalEnv ) ;                                                            \

inst/include/Rcpp/Module.h
457:    Rf_eval(__load_module_call__), R_GlobalEnv );

inst/include/Rcpp/proxy/FieldProxy.h
45:            return Rf_eval( call, R_GlobalEnv ) ;
50:            parent.set__( Rf_eval( call, R_GlobalEnv ) );
70:            return Rf_eval( call, R_GlobalEnv ) ;

inst/include/Rcpp/r_cast.h
98:                    Shield<SEXP> res( Rf_eval( call, R_GlobalEnv ) ) ;

inst/unitTests/cpp/language.cpp
71:    return Rf_eval( call, R_GlobalEnv ) ;

src/barrier.cpp
98:        Rcpp::Shield<SEXP> RCPP(Rf_eval(call, R_GlobalEnv));
140:    Rcpp::Shield<SEXP> RCPP(Rf_eval(call, R_GlobalEnv));
```

Most of these (barring ones used directly for R error / exception handling) should likely be `Rcpp_eval`.

PR #499, opened 2016-06-29
remove many usages of 'Rf_eval' (#498)

kevinushey:
As `Rf_eval` may blow up in our face on Windows.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,15 @@
+2016-06-29  Kevin Ushey  <kevinushey@gmail.com>
+
+        * inst/examples/SugarPerformance/sugarBenchmarks.R: remove Rf_eval usages
+        * inst/include/Rcpp/Environment.h: idem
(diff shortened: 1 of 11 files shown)


kevinushey:
No idea what the Travis error is yet:

```
Error executing .setUp before test.wrap.unordered.map.string.int : Error in eval(substitute(expr), envir, enclos) : 
  object 'datafile' not found
```

eddelbuettel:
@kevinushey What shall we do with this one?  As I recall the underlying issue was actually due to the toolchain as opposed to the code, so maybe time to close this?

kevinushey:
Yes, I think we can close this although I think at least some of the existing usages of `Rf_eval` are code smells,

eddelbuettel:
Shall we retreat for that and start smaller?  I am a little worried that this changeset mostly failed...

kevinushey:
Sounds good.

Issue #502, opened 2016-07-10
long long warnings in algorithm.h


nathan-russell:
Per @dcdillon's comment in #501, 

> The long long warnings are puzzling as all references to long long seem to be wrapped in ifdefs appropriately 

in reference to @thirdwing's compiler output, 

```
* checking whether package ‘with.dot’ can be installed ... WARNING
Found the following significant warnings:
/Library/Frameworks/R.framework/Versions/3.2/Resources/library/Rcpp/include/Rcpp/algorithm.h:48:62: warning: 'long long' is a C++11 extension [-Wc++11-long-long] 
/Library/Frameworks/R.framework/Versions/3.2/Resources/library/Rcpp/include/Rcpp/algorithm.h:77:80: warning: 'long long' is a C++11 extension [-Wc++11-long-long]   
/Library/Frameworks/R.framework/Versions/3.2/Resources/library/Rcpp/include/Rcpp/algorithm.h:89:43: warning: 'long long' is a C++11 extension [-Wc++11-long-long] 
/Library/Frameworks/R.framework/Versions/3.2/Resources/library/Rcpp/include/Rcpp/algorithm.h:100:61: warning: 'long long' is a C++11 extension [-Wc++11-long-long] 
/Library/Frameworks/R.framework/Versions/3.2/Resources/library/Rcpp/include/Rcpp/algorithm.h:117:43: warning: 'long long' is a C++11 extension [-Wc++11-long-long]   
/Library/Frameworks/R.framework/Versions/3.2/Resources/library/Rcpp/include/Rcpp/algorithm.h:128:61: warning: 'long long' is a C++11 extension [-Wc++11-long-long] 
``` 

I have also been experiencing this with more recent versions of Rcpp (since the addition of [algorithm.h](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/algorithm.h)). As far as I can tell, this only happens when compiling with `clang` **and** passing the `-pedantic` flag: 

```
devtools::install_github("RcppCore/Rcpp") 

clang++ -I/usr/share/R/include -DNDEBUG -I../inst/include/ \   
-fpic  -g -O2 -fstack-protector --param=ssp-buffer-size=4 \
-Wformat -Wall -pedantic -Werror=format-security -D_FORTIFY_SOURCE=2 \
-g  -c Date.cpp -o Date.o

In file included from Date.cpp:31:
In file included from ../inst/include/Rcpp.h:84:
[...]

PR #503, opened 2016-07-10
Fix for #502

nathan-russell:
As discussed in #502, use `rcpp_long_long_type` and `rcpp_ulong_long_type` in `algorithm.h` to avoid warnings when compiling with clang in pedantic mode.

### DIFF
--- a/inst/include/Rcpp/algorithm.h
+++ b/inst/include/Rcpp/algorithm.h
@@ -45,7 +45,7 @@ namespace helpers {
 
 #ifdef RCPP_HAS_LONG_LONG_TYPES
     template<>
-    struct ctype_helper< sizeof(CTYPE_LONG_LONG) > { typedef long long type; static const bool value = true; };
+    struct ctype_helper< sizeof(CTYPE_LONG_LONG) > { typedef rcpp_long_long_type type; static const bool value = true; };
 #endif
 
     template<>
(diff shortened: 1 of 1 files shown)


eddelbuettel:
@dcdillon Any comment on this and/or #502?  Should be smooth sailing...

dcdillon:
I would actually suggest commenting out all the long long stuff for now.  It doesn't need to be there for now for algorithm to work

eddelbuettel:
I appears to work just fine with the patch by @nathan-russell  -- same behaviour in rev.dep checks as the test late last week, ie zero new failures.  

So I leaning towards a) keeping it and b) folding the PR in.

dcdillon:
Suits me.  Basically if the long long code in ctype ever gets called by algorithm it will break somewhere else anyhow.  It's just there for completeness.  

Additionally the warning either points to an error in clang or in the Rcpp long long protection macro.  We should figure out which and report it.

dcdillon:
I'd offer to do it but all I have is a 5 inch screen until the 23rd.

eddelbuettel:
This looks good and passed a full rev.dep for which I just [committed](https://github.com/RcppCore/rcpp-logs/commit/3d3ebf60290cf3d6c1d00ec563c06c6b0a8ea3c3) the logs / results.

PR #509, opened 2016-07-17
Rcpp-quickref Vignette Updates (Closes #484)

### LINKED ISSUE #484: Rcpp-quickrefs issue
In the Inline section of the quickrefs sheet, the single quotes in the body aren't single quotes. a copy/paste of the code into R (3.3.0) from the 0.12.5 vignette doesn't work without replacing the quotes.
Additionally, since this is a minor point, I'll add another. In the Important Notes section, it states that many of the examples imply
`using namespace Rcpp;`

If this is mentioned, shouldn't it also mention that they also imply
`#include <Rcpp.h>`
?

coatless:
Preview Update: [Rcpp-quickref.pdf](https://github.com/RcppCore/Rcpp/files/367705/Rcpp-quickref.pdf)

Updated the Important Notes with #include <Rcpp.h>

Updated Inline section to use cppFunction() and mentioned evalCpp()

Added section on Rcpp Attributes and Plugins

Updated Environment with comments

Updated Rcpp function section to include an example pass of an R
function & emphasized loss of performance

Moved Module, Environment, and Function section to the end.

### DIFF
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -41,6 +41,12 @@
     }
     \item Changes in Rcpp Documentation:
     \itemize{
+      \item The Rcpp Quick Reference vignette received a facelift with 
+        new sections on Rcpp attributes and plugins begin added.
+        Furthermore, content was improved in the following
+        sections: important notes, inline compiling, environments,
+        and calling R functions.
+      (James Balamuta in \ghpr{509} fixing \ghit{484}).
       \item A section on default parameters was added to the Rcpp FAQ vignette
       (James Balamuta in \ghpr{505} fixing \ghit{418}).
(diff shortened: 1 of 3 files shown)


inst/NEWS.Rd
```diff
@@ -41,6 +41,12 @@
     }
     \item Changes in Rcpp Documentation:
     \itemize{
+      \item The Rcpp Quick Reference vignette received a facelift with 
+        new sections on Rcpp attributes and plugins begin added.
+        Furthermore, content was improved in the following
+        sections: important notes, inline compiling, environments,
+        and calling R functions.
```
eddelbuettel: Thanks for adding this.  Four rows is a little longer than what we usually do, that length is better for the Changelog.  So I may trim this.

PR #511, opened 2016-07-20
fix minor typo in FAQ

achubaty:


eddelbuettel:
Thanks. It is also missing a NOT in the first half of the sentence.  I'll add that.

PR #514, opened 2016-07-22
Added random number generator unit tests (closes #28)

### LINKED ISSUE #28: tests for random generators
all r* functions : rnorm, etc ...

coatless:
Adds unit test for random generation for the R:: namespace where Rmath
is kept.

Added section breaks in C++ code for each distribution

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2016-07-22  James J Balamuta  <balamut2@illinois.edu>
+
+        * inst/unitTests/cpp/rmath.cpp: Added unit test functions for Rmath
+        random number generators.
+        * inst/unitTests/runit.rmath.R: idem
+
 2016-07-21  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version): Roll minor release
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -10,6 +10,12 @@
       \item The \code{NORET} macro is now defined if it was not already defined
       by R (Kevin fixing issue \ghit{512}). 
     }
+    \item Changes in Rcpp unit tests
+    \itemize{
+      \item New unit tests for random number generators the R namespace which 
+      call the standalone Rmath library. (James Balamuta in \ghpr{514} 
+      addressing issue \ghit{28}).
+    }
   }
 }
 
--- a/inst/unitTests/cpp/rmath.cpp
+++ b/inst/unitTests/cpp/rmath.cpp
@@ -22,6 +22,9 @@
 #include <Rcpp.h>
 using namespace Rcpp ;
 
+
+// ------------------- Normal Distribution
+
 // [[Rcpp::export]]
 NumericVector runit_dnorm( double x, double a, double b ){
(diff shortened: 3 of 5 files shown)


eddelbuettel:
That's nice.  Setting the seed 'on both sides' makes sure we didn't spill the milk.

eddelbuettel:
When I merge #513 this one will conflict as both alter [Changelog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog).

coatless:
I'll resync with main branch and then reload this commit.

coatless:
Weird, works fine under the local check. I'll look into the failure. 

I have a funny feeling it is due to the RNG seed generation difference. Any way to grab the `Rcpp.Rcheck/tests/doRUnit.Rout.fail` off of Travis?

eddelbuettel:
But Travis 'just' running Ubuntu Linux too.  

I think we can mod this part of `.travis.yml`: 

```
after_failure:
  - ./travis-tool.sh dump_logs
```

and look into what the shell script for dump_logs and add to that.

eddelbuettel:
@coatless and I took this off-line and found that the concatenation of two subvectors, ie `c(a1, a2)` appears to results in `c(a2, a1)` making the comparison fail.

coatless:
@eddelbuettel : Modified the test bit to avoid having to `rev()` the vector. 

Cpp now generates under a `for` instead of `::create()`

R now generates under the vectorized `r*(n,...)` functions.

Though, this exercise did reveal something interesting: The concatenation order differs between R & Cpp. I'm still trying to repo this on my end. Might be a bigger bug.

eddelbuettel:
Not sure it is a bug or a just a, well, hesitating, err, _feature_.  The way you had lined up your expression, these were actually not yet evaluated. So we are racing against both concatenation and whether evaluation happens a certain way.  Could be a tricky corner to play in.

eddelbuettel:
There is one thing that this now lacks (but which we can add):  vectorised, ie sugar, accessors.  Ie when you do 

```c++
NumericVector o(5);

for(int i = 0; i < o.size(); i++)
        o[i] = R::rnorm(a, b);
```

you are testing the (scalar) accessor from Rmath via the `R::` namespace.  We also need 

```
RcppNumericVector o = Rcpp::rnorm(5, a, b);
```

but that can be added.

eddelbuettel:
I started to add some more in [this new branch](https://github.com/RcppCore/Rcpp/tree/feature/more-rng-tests) via this [first commit](https://github.com/RcppCore/Rcpp/commit/4517da6ab9f7214606a60d0d216760a73ee543af).

coatless:
@eddelbuettel I have no write access on that branch =(

eddelbuettel:
It's ok -- I'm already done.

coatless:
@eddelbuettel 

Need to also do sugar testing for p/d/q * functions as well.

eddelbuettel:
Maybe, maybe not. The the very highest priority but we should keep it in mind.   It is really just looping over scalar version.  The 'r' variants are equally close but do have the RNG draw business so are potentially more easily breakable.

Issue #506, opened 2016-07-16
Documentation Update / Issue Tracker Cleaning


coatless:
I'm aiming to update some documentation hopefully in time for the 0.12.6 push and subsequently when I'll be covering Rcpp in class.
# Documentation Specific Issues

The issues that I am looking to resolve are:
- [x] #484 opened on May 29 _Rcpp-quickrefs issue_
- [x] #460 opened on Apr 12 _Long vector Rcpp function fails on Windows server, C function works_
- [x] #419 opened on Jan 14 _Strange behavior with CharacterVector and std::sort_
- [x] #418 opened on Jan 13 _Document default arguments in Rcpp FAQ_
- [x] #416 opened on Jan 13 _Add documentation for RCPP_EXPOSED_AS and similar macros in the vignette_
- [ ] #317 opened on Jun 30, 2015 _Coding Standards note ?_
- [ ] #268 opened on Feb 26, 2015 _Problem with automatic RNG scoping_
- [ ] #56 opened on Nov 15, 2013 _Convert uses of inline::cxxfunction to attributes._

Outside of the above issues I will also try to do the following:
- [x] 1. #524 Introduce an "Rcpp beginners" vignette to serve as an "Rcpp Introductory" vignette
- [x] ~~2. Move the current "Rcpp Introduction" to "Rcpp JSS Paper" as the present title is a bit misleading [per SO](http://stackoverflow.com/questions/38399935/how-to-use-a-c-wrapper-in-r-via-rcpp).~~
- [x] ~~3. #604 Try to switch the present vignettes over to use `knitr` or `rmarkdown` if there is no objection. Principal motivation for this change is due to the present way being a bit adhoc and non-standard with respect to `highlight`.~~

# Closing Issues

Lastly, depending on time, I'm going to try to bring the issue tracker up to date by finding the resolution to associated issues via PRs or git commits and asking that those issues are closed. 

## Items completed, but not closed
- [x] #426 opened on Jan 18 _Rcpp::algorithm proposal_
- [x] #321 opened on Jul 6, 2015 _Underlying CHARSXP in String class needs gc protection_
- [x] #277 opened on Mar 15, 2015 _Rcpp 0.11.5 test failures with new Rtools toolchain_
- [x] #272 opened on Mar 9, 2015  _Macros like 'major' still leak, or are protected too aggressively_
- [x] #223 opened on Jan 2, 2015 _Sugar var() only knows numeric_
[...]

PR #515, opened 2016-07-23
Adds compiler warning message (closes #308)

### LINKED ISSUE #308: Error messaging suggestion
This is a nice-to-have suggestion for those of us having less C++ experience and currently using older environments (like RHEL6).  When there is a problem compiling a package, could the error messaging be enhanced to check the compiler version and let the user know when he/she may be using an older, unsupported version of the compiler?  This info could save a lot of debugging/Google time.

This suggestion was inspired by [this issue](https://github.com/Rblp/Rblpapi/issues/39).

coatless:
Modifies /Rcpp/platform/compiler.h to trigger a pragma message
indicating if gcc <= 4.6.2 to satisfy #308 

Part of #506


eddelbuettel:
We will get that `#pragma` shown for each compilation unit, correct?   That may be a little noisy.   Should we maybe just warn once on installation?  I once went towards this route in #474.

@jjallaire What is the _informal census_ on proportion of systems out there still running old RHEL/CentOS?  How many of these are likely to be using Rcpp (maybe via tidyverse) ?

coatless:
@eddelbuettel 

I'm thinking that the only way to do this is going to be via `configure`.

Though, it may be a bit more appropriate to alert the user at the time of compilation that the compiler is problematic. What if the alert just have only *one* line?

```cpp
#pragma message ("WARNING: GCC Compiler Version <= 4.6.2")
```

I say this because sometimes it is just the sysadmin or an inexperience user who might not **catch** the warning in the sea of compile messages if it is only on package install.

eddelbuettel:
May be converging:

- I also use `configure` [here for RcppArmadillo](https://github.com/metacran/r-builder)
- I was thinking earlier today that a one-liner may work better. It might still be bloody noisy though.

coatless:
@eddelbuettel Bad url for the RcppArmadillo line unless it was meant to be generic.

I think you are referring to:

```bash
## If it is g++, we have GXX set so let's examine it
if test "${GXX}" = yes; then
    { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether g++ version is sufficient" >&5
$as_echo_n "checking whether g++ version is sufficient... " >&6; }
    gxx_version=$(${CXX} -v 2>&1 | awk '/^.*g.. version/ {print $3}')
    case ${gxx_version} in
        1.*|2.*|3.*|4.0.*|4.1.*|4.2.*|4.3.*|4.4.*|4.5.*)
             { $as_echo "$as_me:${as_lineno-$LINENO}: result: no" >&5
$as_echo "no" >&6; }
	     { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: Only g++ version 4.6 or greater can be used with RcppArmadillo." >&5
$as_echo "$as_me: WARNING: Only g++ version 4.6 or greater can be used with RcppArmadillo." >&2;}
	     as_fn_error $? "Please use a different compiler." "$LINENO" 5
        ;;
	4.6.*|4.7.*|4.8.*|4.9.*|5.*|6.*)
	     gxx_newer_than_45="-fpermissive"
             { $as_echo "$as_me:${as_lineno-$LINENO}: result: (${gxx_version}) yes" >&5
$as_echo "(${gxx_version}) yes" >&6; }
	;;
    esac
fi
```

https://github.com/RcppCore/RcppArmadillo/blob/master/configure#L2700-L2719

That is a heck of a lot of code that is powering the `configure`. I'm sure about 9/10ths of it isn't needed per https://cran.r-project.org/doc/manuals/R-exts.html#Configure-and-cleanup

eddelbuettel:
@coatless My bad. Updated the URL. Still had the region highlighted in another tab.  In general we always look at `configure.ac` as the generated `configure` is mostly unreadable...

coatless:
@eddelbuettel 

To confirm, the plan is to write a `configure.ac` file with GCC detection and include only one line of the pre-existing message:

```cpp
#pragma message ("WARNING: GCC Compiler Version <= 4.6.2")
```

Sound about right? If so, I'll update this PR on Friday.

eddelbuettel:
Yes, let's refine.  I see two things.

#### First: `configure.ac`

Which we already have, with refinements, in RQuantLib and RcppArmadillo:

```bash
## Set R_HOME, respecting an environment variable if one is set 
: ${R_HOME=$(R RHOME)}
if test -z "${R_HOME}"; then
    AC_MSG_ERROR([Could not determine R_HOME.])   
fi
## Use R to set CXX and CXXFLAGS
CXX=$(${R_HOME}/bin/R CMD config CXX)
CXXFLAGS=$("${R_HOME}/bin/R" CMD config CXXFLAGS)

## We are using C++
AC_LANG(C++)
AC_REQUIRE_CPP

## Check the C++ compiler using the CXX value set
AC_PROG_CXX
## If it is g++, we have GXX set so let's examine it
if test "${GXX}" = yes; then
    AC_MSG_CHECKING([whether g++ version is sufficient])
    gxx_version=$(${CXX} -v 2>&1 | awk '/^.*g.. version/ {print $3}')
    case ${gxx_version} in
        1.*|2.*|3.*|4.0.*|4.1.*|4.2.*|4.3.*|4.4.*|4.5.*)
             AC_MSG_RESULT([no])
	     AC_MSG_WARN([Only g++ version 4.6 or greater can be used with RcppArmadillo.])
	     AC_MSG_ERROR([Please use a different compiler.])   
        ;;
	4.6.*|4.7.*|4.8.*|4.9.*|5.*|6.*)
	     gxx_newer_than_45="-fpermissive"
             AC_MSG_RESULT([(${gxx_version}) yes])
	;;
    esac
fi
```

The ordering is somewhat important, ie need to set the `PATH` so that we get `CXX` we want (Kurt Hornik has been after me a few times on this).  We now need to basically a variable `YesItIsThisBad` which will then used in 2. below.

#### Second the `pragma`

I was hesitating a little earlier and waiting for a hint from @jjallaire about what the 'census proportion' of really old machines is.  He knows because the woes with making RStudio work there etc pp.  Hitting them with the `pragma` on each compilation unit is a little loud.

But I am starting to come around that it may be the right thing.  _We all_ will not see it as we use reasonable compilers.  And those who see it ... really ought to upgrade or stick with older code.

So PR away (provided JJ does not convince us to be more quiet).  Looking forward to it.

thirdwing:
On a CentOS 6.8 VM:

```
> g++ --version
g++ (GCC) 4.4.7 20120313 (Red Hat 4.4.7-17)
Copyright (C) 2010 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
> sessionInfo()
R version 3.3.1 (2016-06-21)
Platform: x86_64-redhat-linux-gnu (64-bit)
Running under: CentOS release 6.8 (Final)

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
```

```
> R CMD INSTALL Rcpp/
* installing to library ‘/usr/lib64/R/library’
* installing *source* package ‘Rcpp’ ...
** libs
g++ -m64 -I/usr/include/R -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic  -c Date.cpp -o Date.o
In file included from ../inst/include/Rcpp/r/headers.h:48,
                 from ../inst/include/RcppCommon.h:29,
                 from ../inst/include/Rcpp.h:27,
                 from Date.cpp:31:
../inst/include/Rcpp/platform/compiler.h:60: note: #pragma message: WARNING: GCC Compiler Version <= 4.6.2
../inst/include/Rcpp/platform/compiler.h:61: note: #pragma message: WARNING: The compiler being used is both OUTDATED and INCOMPLETE in terms of C++ standards;
../inst/include/Rcpp/platform/compiler.h:62: note: #pragma message: WARNING: if something does not work, you should upgrade your compiler!
g++ -m64 -I/usr/include/R -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic  -c Module.cpp -o Module.o
In file included from ../inst/include/Rcpp/r/headers.h:48,
                 from ../inst/include/RcppCommon.h:29,
                 from ../inst/include/Rcpp.h:27,
                 from Module.cpp:24:
../inst/include/Rcpp/platform/compiler.h:60: note: #pragma message: WARNING: GCC Compiler Version <= 4.6.2
../inst/include/Rcpp/platform/compiler.h:61: note: #pragma message: WARNING: The compiler being used is both OUTDATED and INCOMPLETE in terms of C++ standards;
../inst/include/Rcpp/platform/compiler.h:62: note: #pragma message: WARNING: if something does not work, you should upgrade your compiler!
g++ -m64 -I/usr/include/R -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic  -c Rcpp_init.cpp -o Rcpp_init.o
In file included from ../inst/include/Rcpp/r/headers.h:48,
                 from ../inst/include/RcppCommon.h:29,
                 from ../inst/include/Rcpp.h:27,
                 from Rcpp_init.cpp:24:
../inst/include/Rcpp/platform/compiler.h:60: note: #pragma message: WARNING: GCC Compiler Version <= 4.6.2
../inst/include/Rcpp/platform/compiler.h:61: note: #pragma message: WARNING: The compiler being used is both OUTDATED and INCOMPLETE in terms of C++ standards;
../inst/include/Rcpp/platform/compiler.h:62: note: #pragma message: WARNING: if something does not work, you should upgrade your compiler!
g++ -m64 -I/usr/include/R -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic  -c api.cpp -o api.o
In file included from ../inst/include/Rcpp/r/headers.h:48,
                 from ../inst/include/RcppCommon.h:29,
                 from ../inst/include/Rcpp.h:27,
                 from api.cpp:25:
../inst/include/Rcpp/platform/compiler.h:60: note: #pragma message: WARNING: GCC Compiler Version <= 4.6.2
../inst/include/Rcpp/platform/compiler.h:61: note: #pragma message: WARNING: The compiler being used is both OUTDATED and INCOMPLETE in terms of C++ standards;
../inst/include/Rcpp/platform/compiler.h:62: note: #pragma message: WARNING: if something does not work, you should upgrade your compiler!
g++ -m64 -I/usr/include/R -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic  -c attributes.cpp -o attributes.o
In file included from ../inst/include/Rcpp/r/headers.h:48,
                 from ../inst/include/RcppCommon.h:29,
                 from ../inst/include/Rcpp.h:27,
                 from attributes.cpp:39:
../inst/include/Rcpp/platform/compiler.h:60: note: #pragma message: WARNING: GCC Compiler Version <= 4.6.2
../inst/include/Rcpp/platform/compiler.h:61: note: #pragma message: WARNING: The compiler being used is both OUTDATED and INCOMPLETE in terms of C++ standards;
../inst/include/Rcpp/platform/compiler.h:62: note: #pragma message: WARNING: if something does not work, you should upgrade your compiler!
g++ -m64 -I/usr/include/R -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic  -c barrier.cpp -o barrier.o
g++ -m64 -shared -L/usr/lib64/R/lib -o Rcpp.so Date.o Module.o Rcpp_init.o api.o attributes.o barrier.o -L/usr/lib64/R/lib -lR
installing to /usr/lib64/R/library/Rcpp/libs
```

thirdwing:
Maybe we can guess what people are using according to release dates (https://access.redhat.com/articles/3078).

Default compilers on different version are list below:

```
RHEL7 : gcc 4.8.x
RHEL6 : gcc 4.4.x
RHEL5 : gcc 4.1.x
RHEL4 : gcc 3.4.x
RHEL3 : gcc 3.2.x
```

eddelbuettel:
Good list.  We should essentially just tell anybody who is on RHEL 6 or older to bag it and go home.  Sadly, we can't quite ...

coatless:
Okay, per a little bit of discovery... A config file will need to be written as `clang` adds in `gcc 4.2.1` headers into the mix.  

```bash
smac-group-imac:Rcpp james$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 7.3.0 (clang-703.0.31)
```

This throws off the compiler specific detects in `platform/compiler.h` and the `configure` scripts. Note that within the `RcppArmadillo` script the `yes/no` does not trigger on the "g++ version is sufficient" check and, therefore, the line is not ended. So, there needs to be a `default` case that checks if `gxx_version` is empty (the case when compiled against `clang` or `clang-omp`. 

```bash
checking whether g++ version is sufficient... checking LAPACK_LIBS... fallback LAPACK from R 3.3.0 or later used
configure: creating ./config.status
config.status: creating inst/include/RcppArmadilloLapack.h
```

Also, to ease worries about noise, I've added in a`RCPP_GCC_COMPILER_WARNING_SILENT` option to quiet compiles on platforms with <= GCC 4.6.2

inst/include/Rcpp/platform/compiler.h
```diff
@@ -57,10 +44,11 @@
         #define IS_GCC_460_OR_LATER
     #endif
     // Inform users of dated compiler (e.g. <= RHEL6 )
-    #if GCC_VERSION <= 40602
-        #pragma message ("WARNING: GCC Compiler Version <= 4.6.2")
-        #pragma message ("WARNING: The compiler being used is both OUTDATED and INCOMPLETE in terms of C++ standards;")
-        #pragma message ("WARNING: if something does not work, you should upgrade your compiler!")
+    #if GCC_VERSION <= 40602 && !defined(RCPP_GCC_COMPILER_WARNING_SILENT)
+         // Ensure not fake GCC flag set by clang or intel.
+         #if !(defined(__clang__) || defined(__INTEL_COMPILER))
```
eddelbuettel: I guess these do not hurt. I would have hope the earlier `#ifdef __GNUC__` would be enough. Maybe not.

coatless: The reason why I added in the detect genuine GCC is due to the warning being triggered on macOS due to GCC headers being identified as 4.2.1

eddelbuettel:
This now has too.  See discussion in #528.  Can merge manually if you want me to.

eddelbuettel:
Let me call out to @jjallaire one last time ...   JJ: any view as to with all the RHEL systems out there we are too aggressive?   A one-liner message seems fair, particularly with a `#define` toggle to suppress it.

inst/include/Rcpp/platform/compiler.h
jjallaire: Are we saying that anyone running RHEL 6 is going to see this warning when
compiling Rcpp? That I think is far too aggressive as this describes the
vast majority of Linux systems deployed in larger organizations. Worse yet
is that the end users who see this message during compilation have no way
to update the compiler (that's handled by the sysadmin on a multi-user
system, and compiler upgrades are considered extremely heavyweight/risky so
are basically not done).



On Sat, Jul 30, 2016 at 10:13 PM, James J Balamuta <notifications@github.com

eddelbuettel: Yes that is the current discussion but I held off on merging as I wanted to hear from you.

inst/include/Rcpp/platform/compiler.h
```diff
@@ -56,6 +43,13 @@
     #if GCC_VERSION >= 40600
         #define IS_GCC_460_OR_LATER
     #endif
+    // Inform users of dated compiler (e.g. <= RHEL6 )
```
jjallaire: Per my other comment, I don't think we can do this. We have 700+ packages on CRAN all of whom will be subjected to this warning (and all of whom will have to field enquiries about the warning from users and sysadmins worried about it's implications). If we believe that GCC <= 4.6.2 truly warrants a warning message then we need to *change our code* so it doesn't.

eddelbuettel: (I should add that RcppArmadillo now actually fails when this happens, but then again Conrad left me no choice. It has 246 rev.deps and mayhem may well ensue ...)

jjallaire: Right, your hand was forced there.

I should note that I am not at all happy about the state of affairs with
RHEL 6 and we are looking at creating a new binary distribution of R for
RHEL 6 that uses clang. Until then though I don't think we can cry wolf
when being compiled on RHEL 6, it's just too mainstream a platform.

coatless: Throw it out the window... The window... We'll throw it out the window... Err, I'm removing it. 

Do we still want to proceed in throwing a warning via: `configure.ac`? @eddelbuettel @jjallaire

eddelbuettel: I think we could, and should -- but if @jjallaire vetos it I will not object.

jjallaire:
I don't think we should throw a warning of any kind right now because we are in fact fully compatible with GCC 4.6 (and if we aren't compatible we should be). Again, I would like us to get beyond this by having a RHEL 6 compatible R distribution with a more modern compiler but we aren't there yet.

coatless:
@eddelbuettel verdict decreed. Please close out issue #308.

PR #518, opened 2016-07-25
Feature/more rng tests

eddelbuettel:
This adds vectorised sugar tests to the scalar tests based on `R::r...`


eddelbuettel:
Plan to pull this in if nobody objects.

PR #523, opened 2016-07-31
Remove unsafe usage of Rf_eval (close #498)

### LINKED ISSUE #498: eliminate unsafe usages of 'Rf_eval'

thirdwing:
@kevinushey

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,14 @@
+2016-07-31 Qiang Kou <qkou@umail.iu.edu>
+
+        * inst/examples/SugarPerformance/sugarBenchmarks.R: Remove usage of Rf_eval
+        * inst/include/Rcpp/Environment.h: Idem
+        * inst/include/Rcpp/Module.h: Idem
+        * inst/include/Rcpp/exceptions.h: Idem
+        * inst/include/Rcpp/proxy/FieldProxy.h: Idem
+        * inst/include/Rcpp/r_cast.h: Idem
+        * inst/unitTests/cpp/language.cpp: Idem
+        * src/barrier.cpp: Idem
+
 2016-07-24  Dirk Eddelbuettel  <edd@debian.org>
 
         * inst/unitTests/cpp/rmath.cpp: Added RNG unit tests for sugar variants
--- a/inst/examples/SugarPerformance/sugarBenchmarks.R
+++ b/inst/examples/SugarPerformance/sugarBenchmarks.R
@@ -53,7 +53,7 @@ src <- sprintf( '
 
     timer.Reset(); timer.Start();
     for (unsigned int i=0; i<runs; i++) {
-        NumericVector res2 = Rf_eval( call, e ) ;
+        NumericVector res2 = Rcpp_eval( call, e ) ;
     }
     timer.Stop();
     double t3 = timer.ElapsedTime();
--- a/inst/include/Rcpp/Environment.h
+++ b/inst/include/Rcpp/Environment.h
@@ -247,7 +247,7 @@ namespace Rcpp{
                     Shield<SEXP> call( Rf_lang2(internalSym,
                             Rf_lang4(removeSym, Rf_mkString(name.c_str()), Storage::get__(), Rf_ScalarLogical( FALSE ))
                         ) );
-                    Rf_eval( call, R_GlobalEnv ) ;
+                    Rcpp_eval( call, R_GlobalEnv ) ;
                 }
             } else{
                 throw no_such_binding(name) ;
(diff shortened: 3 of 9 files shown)


inst/include/Rcpp/api/meat/Rcpp_eval.h
```diff
@@ -54,7 +54,7 @@ inline SEXP Rcpp_eval(SEXP expr, SEXP env) {
     SET_TAG(CDDR(CDR(call)), ::Rf_install("interrupt"));
 
     // execute the call
-    Shield<SEXP> res(::Rf_eval(call, R_GlobalEnv));
+    Shield<SEXP> res(::Rf_eval(call, env));
```
coatless: Why the switch from `R_GlobalEnv` to `env` here but not on [line 69](https://github.com/thirdwing/Rcpp/blob/20d02de09b155b10cb463b7a3adea35dd8267fab/inst/include/Rcpp/api/meat/Rcpp_eval.h)? 

```cpp
Shield<SEXP> conditionMessage(::Rf_eval(conditionMessageCall, R_GlobalEnv));
```

thirdwing: Actually this switch is not necessary, since we already have the `env` from `evalq`. I involve this when trying to simplify `Rcpp_eval`. I will update.

eddelbuettel:
Thank you -- nice work.  Is that "all of them" or just the ones you thought we could safely change?  I seem to recall @kevinushey mentioning that maybe some needed to remain or something to that extend.  Or am I misremembering?   Anybody to second the thumbs-up?

kevinushey:
Looks great, thanks! @eddelbuettel I was thinking that the `Rf_eval`s e.g. here:

https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/exceptions.h#L197
https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/exceptions.h#L326

probably don't want to be touched, as that's code that will get called in our try-catch macros, e.g. at

https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/macros/macros.h#L37-L57

and so `Rcpp_eval`s within there would be executed outside the 'safety' of a try-catch block, IIUC.

inst/include/Rcpp/api/meat/Rcpp_eval.h
kevinushey: The `evalq` call itself might be unneeded, really -- we could probably omit constructing of the `evalq` bit and just evaluate the `tryCatch` expression in `env`.

We want to ensure that `expr` is evaluated in `env`, but the other usages likely want to just be evaluated in the global environment.

eddelbuettel:
Thank you both -- in it goes!  And I resolve to do a full rev.dep check tomorrow. Worst case we revert.

eddelbuettel:
Just for the record it passed the rev.dep with flying colours as expected.

Issue #525, opened 2016-08-01
Decreasing option for Vector's sort member function 


nathan-russell:
Currently [`sort`](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Vector.h#L385-L400) does not have a parameter controlling the sort direction (ascending vs. descending). I have an update which would provide this option, changing the signature to `sort(bool decreasing = false)`, as is the default in `base::sort`. 

Although it is trivial to do this from client code for `NumericVector` and `IntegerVector` with `std::sort(..., std::greater<T>())`, that is not the case for `CharacterVector` (#419), so I think an internal implementation is warranted. If there are no objections, I can submit a PR for further review.

coatless:
This would be great as the documentation for #419 in the FAQ can be switched from noting a flaw to providing an actionable solution. (Though, it might be also helpful to highlight that the use must be via the internal helper instead of stl.)

nathan-russell:
@coatless Agreed -- I think this will provide a suitable workaround for a fair number of situations where #419 emerges as a problem, but the STL / `string_proxy` incompatibility is definitely something that should be documented.

PR #527, opened 2016-08-01
Added decreasing option for Vector sort (#525)

nathan-russell:
As discussed in #525, this modifies `Vector::sort` such that 

- `x.sort()` (and `x.sort(false)`) will sort `x` ascendingly 
- `x.sort(true)` will sort `x` descendingly


thirdwing:
LGTM.

Issue #387, opened 2015-10-30
Fix invalid C++ identifiers


klmr:
Rcpp uses `__` as a prefix for identifiers in auto-generated code (e.g. [here](../blob/9642541239f0985d0f211d106f2cdc3169806d23/src/attributes.cpp#L2013) and [here](../blob/9642541239f0985d0f211d106f2cdc3169806d23/src/attributes.cpp#L2116)). This is problematic, since [`__` is not not valid in C++ identifiers](http://stackoverflow.com/a/228797/1968) (it’s reserved for use by the compiler).

Such identifiers are for instance used by standard library implementations. And although most modern implementations try to avoid aggravating users, and so use more explicit conventions in their macro names, a simple grep through the libc++ implementation shows a few macros with names such as `__need_size_t` (and more in libstdc++), so it’s not unthinkable that there might be conflicts with future versions of some compilers.

It would be better to replace these instances with a legal identifier token, e.g. `Rcpp_`.

eddelbuettel:
Thanks for the note, Konrad.

That is attributes.cpp only and hence well localised.  I would encourage you (as the one with an itch to scratch) to work on a pull request if you want to see change "soon" as we all aer  atad to budy to write code on demand for what I'd call non-bugs.

We effectively live in a g++/clang++ world only even though the occasional user may edploy the Intel compiler or work on some non-standard OS besides Windoof, OS X and Linux.  And in that world this just doesn't seem to have bitten so _shrugs_ ...

jjallaire:
We used those prefixes specifically because they are reserved for the
compiler (and thus extremely unlikely to conflict with user variable
names). Point well taken though that we might end up conflicting with the
compiler as a result! You can in some ways view compileAttributes
pre-processing as effectively part of the (extended) compilation and thus
not at all an inappropriate place to use __ prefixes.

klmr:
@jjallaire I agree that *a* prefix is entirely appropriate (and arguably even necessary) here.

I can try my hand at a PR.

eddelbuettel:
I read the comment differently from you, @klmr :

> You can [...] view compileAttributes pre-processing as [...] part of the [...] compilation and thus
not at all an inappropriate place to use __ prefixes.

Looks like we're not yet convinced we need it but heck, if you submit something we will look at it.

klmr:
[...]

PR #528, opened 2016-08-02
Auto generation Warnings & Invalid C++ Identifiers (Closes #526 and #387)

### LINKED ISSUE #526: Autogeneration Warning for RcppExports.*
As shown in [Rcpp function with no argument](http://stackoverflow.com/questions/38675328/rcpp-function-with-no-argument) on StackOverflow, there may need to be some slight changes associated with:

1.  Code generation headers in **RcppExports.cpp** and **RcppExports.R**
2. Documentation surrounding `Rcpp.package.skeleton()` and `compileAttributes()`

## The current headers:

**RcppExports.R**
```r
# This file was generated by Rcpp::compileAttributes
# Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
```
**RcppExports.cpp**
```rcpp
// This file was generated by Rcpp::compileAttributes
// Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
```

## The proposed new headers:

**RcppExports.R**

```r
# Generated by using Rcpp::compileAttributes() - do not edit by hand
# Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
```
**RcppExports.cpp**
```rcpp
// Generated by using Rcpp::compileAttributes() - do not edit by hand
// Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
```

Mimics the descriptor used by `roxygen2`:
```r
% Generated by roxygen2: do not edit by hand
```

### Rcpp.package.skeleton()

Modify the details section to emphasize autocreation and not to modify by hand

Current:
> If the attributes argument is TRUE, then rather than generate the example files as described above, a single rcpp_hello_world.cpp file is created in the src directory and it's attributes are compiled using the compileAttributes function, so files RcppExports.R and RcppExports.cpp are generated as well.

Proposed:
> If the attributes argument is TRUE, then rather than generate the example files as described above, a single rcpp_hello_world.cpp file is created in the src directory and it's attributes are compiled using the compileAttributes function. **This leads to the files RcppExports.R and RcppExports.cpp being generated. They are automatically regenerated from \emph{scratch} each time compileAttributes is called. Therefore, one should not modify by hand either RcppExports file.**

### compileAttributes

Current
> For C++ functions adorned with the Rcpp::export attribute, the C++ and R source code required to bind to the function from R is generated and added (respectively) to src/RcppExports.cpp or R/RcppExports.R.

Proposed:
> For C++ functions adorned with the Rcpp::export attribute, the C++ and R source code required to bind to the function from R is generated and added (respectively) to src/RcppExports.cpp or R/RcppExports.R. **Both of these files are automatically generated from \emph{scratch} each time compiledAttributes is run.**

Tagging as a part of #506

coatless:
Added Autogeneration Warnings & Invalid C++ Identifiers (Closes #526 and #387)


src/attributes.cpp
```diff
@@ -2061,18 +2062,18 @@ namespace attributes {
...
+                ostr() << "        if (RCPP_result_gen_.inherits(\"try-error\"))"
                        << std::endl
                        << "            throw Rcpp::exception(as<std::string>("
-                       << "__result).c_str());"
+                       << "RCPP_result_gen_).c_str());"
                        << std::endl;
                 if (!function.type().isVoid()) {
                     ostr() << "        return Rcpp::as<" << function.type() << " >"
-                           << "(__result);" << std::endl;
+                           << "(RCPP_result_gen_);" << std::endl;
```
jjallaire: I'd say drop the trailing underscore as this convention is used elsewhere (e.g. in the attributes implementation) to indicate a C++ member variable.

src/attributes.cpp
```diff
@@ -2196,7 +2197,7 @@ namespace attributes {
     }
 
     std::string CppPackageIncludeGenerator::getHeaderGuard() const {
-        return "__" + packageCpp() + "_h__";
+        return "RCPP_" + packageCpp() + "_h_gen_";
```
jjallaire: Here I think we want to add a leading underscore (and keep the trailing underscore) to be consistent with the use of header guards elsewhere.

eddelbuettel:
We now have a merge conflict.  Do you want to / have time to work through it?   Else I can merge it manually later bu then you don't get the glory ...

src/attributes.cpp
coatless: To confirm, does this mean:

`"_rcpp_" + packageCpp() + "_h_gen_`

or

`"_RCPP_" + packageCpp() + "_h_gen_`?

jjallaire: @eddelbuettel had suggested capitals (which is generally what I've seen in header guards as well). I don't have a super strong preference but I think we should be consistent, e.g. either:

```cpp
"_rcpp_" + packageCpp() + "_h_gen_"
```

or

```cpp
"_RCPP_" + packageCpp() + "_H_GEN_"
```

eddelbuettel:
Will merge this as nobody cried foul ...

rafstraumur:
Hi, might I suggest to drop the leading underscore at all? So instead of writing

        return "_RCPP_" + packageCpp() + "_RCPPEXPORTS_H_GEN_";
        ...
        return "_RCPP_" + packageCpp() + "_H_GEN_";

just write:

      return "RCPP_" + packageCpp() + "_RCPPEXPORTS_H_GEN_";
      ...
      return "RCPP_" + packageCpp() + "_H_GEN_";

See our discussion on stackoverflow: http://stackoverflow.com/questions/39291133/how-to-change-header-include-guards-in-rcpp-interfaces/39291326#39291326 and http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier/228797#228797 

> All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.

Best,
Simon

PR #529, opened 2016-08-02
Rcpp::String ctor update(close #263)

### LINKED ISSUE #263: String class needs someway to declare encoding of string
e.g. instead of `Rf_mkCharCE("abc", CE_UTF8)`

thirdwing:
As discussed in #263 

(1) remove the string representation of encoding (I am not sure if any pkgs really use this, we might need a rev.dep);

(2) use `CE_UTF8` as default encoding (not sure if this will cause problems with no-English string).

@kevinushey @hadley @jjallaire

### DIFF
--- a/inst/include/Rcpp/String.h
+++ b/inst/include/Rcpp/String.h
@@ -40,7 +40,6 @@
     #define RCPP_STRING_DEBUG_3(fmt, M1, M2, M3)
 #endif
 
-
 namespace Rcpp {
 
     /**
@@ -53,7 +52,7 @@ namespace Rcpp {
         typedef internal::const_string_proxy<STRSXP> const_StringProxy;
 
         /** default constructor */
-        String(): data(Rf_mkChar("")), buffer(), valid(true), buffer_ready(true), enc(CE_NATIVE) {
+        String(): data(Rf_mkCharCE("", CE_UTF8)), buffer(), valid(true), buffer_ready(true), enc(CE_UTF8) {
             Rcpp_PreserveObject(data);
             RCPP_STRING_DEBUG("String()");
         }
@@ -61,12 +60,6 @@ namespace Rcpp {
         /** copy constructor */
         String(const String& other) : data(other.get_sexp()), valid(true), buffer_ready(false), enc(Rf_getCharCE(other.get_sexp())) {
             Rcpp_PreserveObject(data);
-            RCPP_STRING_DEBUG("String(const String&)");
-        }
-
-        String(const String& other, const std::string& enc) : data(other.get_sexp()), valid(true), buffer_ready(false) {
-            Rcpp_PreserveObject(data);
-            set_encoding(enc);
             RCPP_STRING_DEBUG("String(const String&)");
         }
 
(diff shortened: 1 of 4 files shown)


inst/include/Rcpp/String.h
```diff
@@ -53,7 +52,7 @@ namespace Rcpp {
         typedef internal::const_string_proxy<STRSXP> const_StringProxy;
 
         /** default constructor */
-        String(): data(Rf_mkChar("")), buffer(), valid(true), buffer_ready(true), enc(CE_NATIVE) {
+        String(): data(Rf_mkChar("")), buffer(), valid(true), buffer_ready(true), enc(CE_UTF8) {
```
kevinushey: Should the data element be initialized with `Rf_mkCharCE()`?

thirdwing: Agree.

inst/include/Rcpp/String.h
```diff
@@ -64,10 +63,10 @@ namespace Rcpp {
             RCPP_STRING_DEBUG("String(const String&)");
         }
 
-        String(const String& other, const std::string& enc) : data(other.get_sexp()), valid(true), buffer_ready(false) {
+        String(const String& other, cetype_t enc) : data(other.get_sexp()), valid(true), buffer_ready(false) {
```
kevinushey: This method feels a little weird:

1. It accepts a `String` object, which should have its own encoding; but
2. It _overrides_ that encoding with the encoding passed in using `set_encoding`.

I'm curious where / how this is used; it feels a little strange.

inst/include/Rcpp/String.h
```diff
@@ -88,7 +87,7 @@ namespace Rcpp {
             RCPP_STRING_DEBUG("String(SEXP)");
         }
 
-        String(SEXP charsxp, const std::string& enc) : data(R_NilValue) {
+        String(SEXP charsxp, cetype_t enc) : data(R_NilValue) {
```
kevinushey: Same as above -- overriding encoding on a passed-in charsxp.

inst/include/Rcpp/String.h
thirdwing: Actually this is mostly to make our ctor matrix complete. These two ctors can be removed. If we want to change the encoding of `String s`, we can just `s.set_encoding()`. `set_encoding` is public.

thirdwing: So remove or keep them? @kevinushey

kevinushey: I would be in favor of removing them -- I think it's better for users to be explicit with `set_encoding` if they wanted to force a different encoding.

thirdwing: Removed.

kevinushey:
PR looks good for me; let's see if anyone else has thoughts.

eddelbuettel:
Same here.  Lemme sleep it over and give a glance over morning coffee in 6 to 7 hours and it'll go in.

eddelbuettel:
Merged by hand

PR #531, opened 2016-08-04
Fix for redeclaration of 'Rboolean rcpp_isError_gen'

coatless:
Adds a fix for the bad redeclaration introduced in: #528 

`rcpp_isInterrupt_gen` => `rcpp_isError_gen` by accident.


eddelbuettel:
All good now.  I roll the minor so that we call this 0.12.6.4.

PR #532, opened 2016-08-05
Modernize rcppgibbs

coatless:
Part of #56's push to switch examples away from the cxxfunction to Rcpp attributes.

From the discussion of #56, it seems like there will be multiple PRs to complete the job. Specifically, 1 PR per example directory / vignette.


eddelbuettel:
Looks beautiful and is in a size that is easy to chew and digest.  

Plan to merge, but will leave it here for the remainder of the day for others to chime in.

kevinushey:
Looks good to me as well!

PR #533, opened 2016-08-05
Fixed bug relating to copying of FormatList object.

rohan-shah:
On the intel compiler, version 15.0.5.223 there appears to be a bug in tinyformat.h, which in my case is triggered by use of the readxl package.

Due to the ridiculous amount of templating it's hard for me to be sure, but I believe that somehow an object of class detail::FormatListN<N> is being copied as an object of class FormatList. The original FormatListN<N> object is then deleted, and the copied FormatList now contains an invalid pointer.

This may be compiler specific.

### DIFF
--- a/inst/include/Rcpp/utils/tinyformat.h
+++ b/inst/include/Rcpp/utils/tinyformat.h
@@ -850,6 +850,7 @@ class FormatList
                             const FormatList& list);
 
     private:
+        FormatList(const FormatList& other);
         const detail::FormatArg* m_formatters;
         int m_N;
 };


rohan-shah:
So according to CI the patch fails to compile on gcc 4.6.3, but I can confirm it works with intel. Regardless, I don't think those objects should be copyable.

eddelbuettel:
If you want us to take a look at an issue, not breaking the deployment variant used by the vast majority of our users (lumping g++ and clang++ together here) is a good first step.

Now suppose something were wrong with tinyformat -- which is entirely possible.  Even then this repo would _not_ be the place to file it.  We just use it.  Please go upstream, sort it out there and then maybe alert us to an updated version.

So in sum there is currently nothing really for us here.  I will close this, but am of course more than ready to reopen this once material discussions related to Rcpp are to be had here.

Thanks for your understanding.

PR #535, opened 2016-08-06
Modernizing fastlm and functionCallback

coatless:
Part of #56's push to switch examples away from the cxxfunction to Rcpp
attributes.


eddelbuettel:
Thanks for bringing the current RcppArmadillo version in.  With the work I did for RcppGSL some, oh, 14 or 15 months ago we can make that one better too (no more `alloc()` and `free()` !!) and I'll get to that.

coatless:
@eddelbuettel: I'll rewrite and submit a PR using those model types tonight.

PR #541, opened 2016-08-15
Modernizing GSL's FastLM function to use new RcppGSL  templates.

coatless:
Update of #535 ’s attempt to modernize GSL per Dirk’s comment.

Added Travis-CI News entry as well.


eddelbuettel:
Looks good, no objections so far -- merging ahead.

Issue #549, opened 2016-09-03
Sugar functions - rowSums, colSums, rowMeans, colMeans


nathan-russell:
I put together sugar functions for `rowSums`, `colSums`, `rowMeans`, and `colMeans` (for use with matrices, not `data.frame`s) which all seem to be working as expected. However, early on in the process I switched from using `Rcpp::traits::is_na<>` to the R macro `ISNAN` to check for `NA`s / `NaN`s in numeric matrices because I noticed a large difference in performance. As an example, 

```cpp
#include <Rcpp.h>

// [[Rcpp::export]]
Rcpp::LogicalVector macro_na(Rcpp::NumericVector x) {
    R_xlen_t i = 0, sz = x.size();
    Rcpp::LogicalVector res(sz);

    for ( ; i < sz; i++) {
        res[i] = ISNAN(x[i]);
    }

    return res;
}

// [[Rcpp::export]]
Rcpp::LogicalVector traits_na(Rcpp::NumericVector x) {
    R_xlen_t i = 0, sz = x.size();
    Rcpp::LogicalVector res(sz);

    for ( ; i < sz; i++) {
        res[i] = Rcpp::traits::is_na<REALSXP>(x[i]);
    }

    return res;
}

/*** R

set.seed(123); x <- sample(c(rnorm(4), NA), 1e6, TRUE)

microbenchmark::microbenchmark(
    "Macro" = macro_na(x),
    "Traits" = traits_na(x),
    times = 200L
)
# Unit: milliseconds
#    expr       min        lq      mean    median        uq      max neval
#   Macro  6.075743  6.158497  7.524288  6.259146  8.587615 26.61165   200
#  Traits 13.508780 13.621339 16.530404 13.753101 15.818473 92.80678   200
 
*/ 
``` 
------ 

I believe the difference is due to [the `memcmp` call here](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/internal/na.h#L39-L64). @kevinushey It looks like you authored this, so can you comment on whether or not it is safe to be using `ISNAN` in place of `traits::is_na`? I read the explanatory comment, 

> motivation: on 32bit architectures, we only see 'LargeNA'
> as defined ahead; on 64bit architectures, R defaults to
> 'SmallNA' for R_NaReal, but this can get promoted to 'LargeNA'
> if a certain operation can create a 'signalling' NA, e.g. NA_real_+1

[...]

PR #551, opened 2016-09-04
row/col-Sums/Means and unit tests - for #549

nathan-russell:
As discussed in #549, this commit adds Sugar functions for `rowSums`, `colSums`, `rowMeans`, and `colMeans`, along with unit tests.

### DIFF
--- a/inst/include/Rcpp/sugar/functions/rowSums.h
+++ b/inst/include/Rcpp/sugar/functions/rowSums.h
@@ -48,37 +48,37 @@ inline bool check_na(Rcomplex x) {
 }
 
 
-inline void incr(double& lhs, double rhs) {
-    lhs += rhs;
+inline void incr(double* lhs, double rhs) {
+    *lhs += rhs;
 }
 
-inline void incr(int& lhs, int rhs) {
-    lhs += rhs;
+inline void incr(int* lhs, int rhs) {
+    *lhs += rhs;
 }
 
-inline void incr(Rcomplex& lhs, const Rcomplex& rhs) {
-    lhs.r += rhs.r;
-    lhs.i += rhs.i;
-}
-
-
-inline void div(double& lhs, R_xlen_t rhs) {
-    lhs /= rhs;
-}
-
-inline void div(Rcomplex& lhs, R_xlen_t rhs) {
-    lhs.r /= rhs;
-    lhs.i /= rhs;
+inline void incr(Rcomplex* lhs, const Rcomplex& rhs) {
+    lhs->r += rhs.r;
+    lhs->i += rhs.i;
 }
 
 
-inline void set_nan(double& x) {
-    x = R_NaN;
+inline void div(double* lhs, R_xlen_t rhs) {
+    *lhs /= rhs;
 }
 
-inline void set_nan(Rcomplex& x) {
-    x.r = R_NaN;
-    x.i = R_NaN;
+inline void div(Rcomplex* lhs, R_xlen_t rhs) {
+    lhs->r /= rhs;
+    lhs->i /= rhs;
+}
+
+
+inline void set_nan(double* x) {
+    *x = R_NaN;
+}
+
+inline void set_nan(Rcomplex* x) {
+    x->r = R_NaN;
+    x->i = R_NaN;
 }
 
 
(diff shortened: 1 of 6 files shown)


eddelbuettel:
That looks like it is really nice work, once again.  Thank so much!

inst/include/Rcpp/sugar/functions/rowSums.h
```diff
@@ -0,0 +1,963 @@
...
+inline bool check_na(SEXP x) {
+    return x == NA_STRING;
+}
+
+inline bool check_na(Rcomplex x) {
+    return ISNAN(x.r) || ISNAN(x.i);
+}
+
+
+inline void incr(double& lhs, double rhs) {
```
kevinushey: Minor style issue, but I think it's a preferable to receive arguments that will be mutated by pointer, rather than by reference -- it makes it more obvious at the call site whether a passed-in parameter might be mutated. For example,

```
do_something(x);
```

versus

```
do_something(&x);
```

In the first case, I think the expectation most would have is that `do_something(x)` will not mutate `x`, whereby it's more obvious that mutation might happen with `do_something(&x)`.

The main downside is the fact that pointers can be `NULL` while references can't, but I think the clarity in code makes up for that. I'll let others weigh on whether that change is worth making.

inst/include/Rcpp/sugar/functions/rowSums.h
```diff
@@ -0,0 +1,963 @@
...
+                }                                                                                           \
+            }                                                                                               \
+        }                                                                                                   \
+                                                                                                            \
+        return res;                                                                                         \
+    }                                                                                                       \
+}; 
+
+ROW_SUMS_IMPL_RMNA(LGLSXP) 
+ROW_SUMS_IMPL_RMNA(INTSXP)
```
kevinushey: Let's `#undef` the macro after we're done with it.

eddelbuettel:
No strong feelings on pointers (clearer as @kevinushey says) versus references (bettern 'modern' C++ ?).

inst/include/Rcpp/sugar/functions/rowSums.h
jjallaire: I agree with Kevin, especially in the case where by local idiom there is no
possibility of the pointer being NULL (which BTW would already be true if
you can pass a reference there).

inst/include/Rcpp/sugar/functions/rowSums.h
```diff
@@ -0,0 +1,963 @@
...
+                res[i] = NA_INTEGER;                                                                        \
+            }                                                                                               \
+        }                                                                                                   \
+                                                                                                            \
+        return res;                                                                                         \
+    }                                                                                                       \
+}; 
+
+ROW_SUMS_IMPL_KEEPNA(LGLSXP) 
+ROW_SUMS_IMPL_KEEPNA(INTSXP)
```
kevinushey: Same `#undef` here?

inst/include/Rcpp/sugar/functions/rowSums.h
```diff
@@ -0,0 +1,963 @@
...
+
+
+#undef ROW_SUMS_IMPL_KEEPNA
+#undef ROW_SUMS_IMPL_RMNA
+#undef COL_SUMS_IMPL_KEEPNA
+#undef COL_SUMS_IMPL_RMNA
+#undef ROW_MEANS_IMPL_KEEPNA
+#undef ROW_MEANS_IMPL_RMNA
+#undef COL_MEANS_IMPL_KEEPNA
+#undef COL_MEANS_IMPL_RMNA
```
kevinushey: Oh, I see now that everything is `#undef`ed at the bottom here. Marginal preference to `#undef` immediately after you know you don't need it anymore, but I think this is fine too.

kevinushey:
Barring some very minor stylistic suggestions, LGTM!

eddelbuettel:
Thanks @kevinushey .  We'll see if @nathan-russell has time for a revision, else we can probably take it as is.

nathan-russell:
Those are reasonable suggestions; thanks for reviewing everyone.

PR #560, opened 2016-10-21
Change string_elt/vector_elt to accept R_xlen_t

thirdwing:
As @eddelbuettel suggest, break #556 into some small PRs.

This also makes the Changelog looks better.


eddelbuettel:
Restarted, and passed as expecting.   So ... merging.

hrbrmstr:
Aye, that'd be the reason. We've been tracking the outage where I work and
watching the various providers try to compensate. Such a fragile internet
we have.

Issue #564, opened 2016-10-22
Matrix utility function ones, zeros, eye, ...


eddelbuettel:
There are a bunch of matrix construction helpers in typical Matrix packages such as Armadillo. We could add some:
- `ones(n)` to fill a matrix of size n with ones (not sure if we need non-square size here)
- `zeros(n)` similarly to fill with zeros
- `eye(n)` to file a matrix of zeros along the diagonal.

We do have `diag()` as an extractor though.

nathan-russell:
I'm not sure if you wanted to add member and non-member versions of these functions, as is done with Armadillo, but I put something together to tackle (just) the latter case, and can submit a PR if you'd like to review.

eddelbuettel:
Simpler may well be better -- was thinking 'sugar' so non-member functions is my leaning too.

thirdwing:
I think this has been fixed in #566 and #569

Issue #562, opened 2016-10-22
New matrix ctor to allow for fill on dimension creation


coatless:
Going through the matrix options right now, there seems to be only two options to fill a matrix:

``` cpp
// by vector
NumericVector x = NumericVector(100, 2.0); 
y = NumericMatrix(10, 10, v.begin());


// using stl
NumericMatrix z(10,10);
std::fill(z.begin, z.end(), 2.0);
```

However, as indicated by option using a vector, it would be nice to be able to immediately specify the fill value within the constructor:

``` cpp
// proposed addition
NumericMatrix z2(10, 10, 2.0);
```

eddelbuettel:
Yep, could do.

coatless:
The ctor would need to be:

```cpp
template <typename U>
Matrix( const int& nrows_, const int& ncols, const U& u) : 
        VECTOR( Dimension( nrows_, ncols ), u ), nrows(nrows_) { } // Vector ctor already exists.
```
    
The only issue with this approach is a ctor of `Matrix(int, int, typename)` already exists for handling iterators

```cpp
template <typename Iterator>
Matrix( const int& nrows_, const int& ncols, Iterator start ) :
        VECTOR( start, start + (static_cast<R_xlen_t>(nrows_)*ncols) ),
        nrows(nrows_)
{
        VECTOR::attr( "dim" ) = Dimension( nrows, ncols ) ;
}
```

thirdwing:
If we really need this, it can be done like below:

```cpp
template <typename U>
Matrix(const int& nrows_, const int& ncols, const U& u,
       typename Rcpp::traits::enable_if<traits::is_arithmetic<U>::value, void>::type* = 0)
```

coatless:
Sadly, when I tried this, it was picked up as ambiguous since the first three parameters are the same. 

However, I think a modification to the `Iterator` based ctor to verify if the object is an iterator should suffix to fix the problem. Browsing quickly through the source, it looks like [`Rcpp::traits`]() will need to be modified to support a SFINAE `is_iterator()`.

 Though, a simpler solution would be to create a matrix ctor with

```cpp
template <typename U>
Matrix( Dimension &dims, const U& u ) : VECTOR( Dimension( dims ), u )
```

Or as @eddelbuettel has proposed in the PR, just simply rely on the `.fill()`.

PR #565, opened 2016-10-23
operator= for Matrix class (close #563)

### LINKED ISSUE #563: Matrix assigning to double changes
Related to #562, this seems like a bug:

``` r
R> cppFunction("NumericMatrix testmat(int n, double fillme) { NumericMatrix M(n, n); M = fillme; return M; }") 
R> testmat(2L, 3)
     [,1] [,2] [,3]
[1,]    0    0    0
[2,]    0    0    0
[3,]    0    0    0
R> 
```

I was expecting a 2x2 matrix will with the value 3.  No mas.

thirdwing:
Just make `Matrix` to use the `operator=` from `Vector` class. This should fix #563 
```cpp
#include<Rcpp.h>

using namespace Rcpp;

//[[Rcpp::export]]
NumericMatrix testmat(int n, double fillme) {
  NumericMatrix M(n);
  M = fillme;
  return M;
}

//[[Rcpp::export]]
NumericVector testvec(int n, double fillme) {
  NumericVector M(n);
  M = fillme;
  return M;
}
/*
> Rcpp::sourceCpp("test.cpp")
> testmat(2, 3)
[1] 3
> testvec(2, 3)
[1] 3
*/
```

Personally I prefer being consistent with `Vector`, since `Matrix` is derived from it.

@kevinushey @eddelbuettel

### DIFF
--- a/inst/include/Rcpp/vector/Matrix.h
+++ b/inst/include/Rcpp/vector/Matrix.h
@@ -85,7 +85,14 @@ public:
         nrows = other.nrows ;
         return *this ;
     }
+
     Matrix& operator=( const SubMatrix<RTYPE>& ) ;
+
+    template <typename T>
+    Matrix& operator=( const T& x) {
+        this->assign_object( x, typename traits::is_sugar_expression<T>::type() ) ;
+        return *this ;
+    }
 
     explicit Matrix( const no_init_matrix& obj) {
         Storage::set__( Rf_allocMatrix( RTYPE, obj.nrow(), obj.ncol() ) );
--- a/inst/include/Rcpp/vector/Vector.h
+++ b/inst/include/Rcpp/vector/Vector.h
@@ -602,6 +602,20 @@ protected:
         internal::r_init_vector<RTYPE>(Storage::get__()) ;
     }
 
+    // sugar
+    template <typename T>
+    inline void assign_object( const T& x, traits::true_type ) {
+        assign_sugar_expression( x.get_ref() ) ;
+    }
+
+    // anything else
+    template <typename T>
+    inline void assign_object( const T& x, traits::false_type ) {
+        Shield<SEXP> wrapped(wrap(x));
+        Shield<SEXP> casted(r_cast<RTYPE>(wrapped));
+        Storage::set__(casted);
+    }
+
 private:
 
     void push_back__impl(const stored_type& object, traits::true_type ) {
@@ -997,20 +1011,6 @@ private:
             Shield<SEXP> casted(r_cast<RTYPE>(wrapped));
             Storage::set__(casted);
         }
-    }
-
-    // sugar
-    template <typename T>
-    inline void assign_object( const T& x, traits::true_type ) {
-        assign_sugar_expression( x.get_ref() ) ;
-    }
-
-    // anything else
-    template <typename T>
-    inline void assign_object( const T& x, traits::false_type ) {
-        Shield<SEXP> wrapped(wrap(x));
-        Shield<SEXP> casted(r_cast<RTYPE>(wrapped));
-        Storage::set__(casted);
     }
 
     // we are importing a real sugar expression, i.e. not a vector


eddelbuettel:
I like the sentiment expressed in

> Personally I prefer being consistent with `Vector`, since `Matrix` is derived from it.

But I don't get your example. Now they both revert to a _scalar_ when we'd expect, respectively, a matrix and a vector?  Colour me confused...

kevinushey:
I mean, the current behavior makes much more sense with literals, e.g.

```cpp
NumericVector x;
x = 10.0;
```

Most users would expect that `x` is now a 'scalar' (vector of length one) with value `10`, I think. I don't think `operator=` should do an element-wise fill.

eddelbuettel:
Right right right. I just looked at #563 and what I wrote. I was surprised with the outcome as I expected an element-wise filling-in.

But you two are quite right.  That is actually not the right thing.  Reassigning _may_ be better.

But changing dimensions _silently_ is also ... a bit weird.  Should we warn or something?

kevinushey:
IMO it's effectively the same thing as doing e.g.

```R
x <- 1:10
x <- 10
```

so the changing of dimension (for the vector case) isn't that strange.

I don't know how I feel about implementing `operator=` for a `NumericMatrix` though. What is expected here -- a true 'scalar'? A numeric matrix with dimensions `[1, 1]`? I think it's confusing enough that leaving `operator=` unimplemented for matrices is fine.

Do we have `fill()` methods for vectors / matrices? If not, we should likely just add those.

eddelbuettel:
But the R example is for R -- a dynamically-typed language.  C++ is trickier. And look eg at [Conrad's advanced constructors for Armadillo](http://arma.sourceforge.net/docs.html#Mat) -- you can even 'freeze' the dimensions (though you can also have them altered):

> when `strict` is set to `true`, the matrix will be bound to the auxiliary memory for its lifetime; the number of elements in the matrix can't be changed

Fill methods, as well as helper functions (per #564) would be good to have.

thirdwing:
Maybe we can just use `std::fill` since we have defined iterator.

```cpp
#include<Rcpp.h>

using namespace Rcpp;

//[[Rcpp::export]]
NumericMatrix test_fill(double fillme) {
  NumericMatrix test(5, 5);
  std::fill(test.begin(), test.end(), fillme);
  return test;
}
/*
> Rcpp::sourceCpp("test.cpp")
> test_fill(3.5)
     [,1] [,2] [,3] [,4] [,5]
[1,]  3.5  3.5  3.5  3.5  3.5
[2,]  3.5  3.5  3.5  3.5  3.5
[3,]  3.5  3.5  3.5  3.5  3.5
[4,]  3.5  3.5  3.5  3.5  3.5
[5,]  3.5  3.5  3.5  3.5  3.5
*/
```

eddelbuettel:
And how about _not_ adding to the ever-expanding constructor soup but rather relying on, say,

```c++
NumericMatrix M(5,5).fill(3.5);
```

with a `fill()` member function similar to the above?

coatless:
@thirdwing 'tis how it is presently done (see option 1 in #562 )

The [`Vector`](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Vector.h#L315) class has a public `fill` method that hooks into [stl or a for](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/Vector.h#L1056-L1069).

thirdwing:
I am not very sure `NumericMatrix M(5,5).fill(3.5);` is valid or not in C++.

We can use `fill` as below without any further changes.

```cpp
#include <Rcpp.h>

using namespace Rcpp;

//[[Rcpp::export]]
NumericMatrix test_fill(double fillme) {
  NumericMatrix test(5, 5);
  test.fill(fillme);
  return test;
}
````

kevinushey:
All in all, my vote is to close this PR -- I'm personally happier with `operator=` undefined for matrices.

I don't think we need to take any action with `operator=` on the Vector class, but could be convinced otherwise.

thirdwing:
@kevinushey I agree with you on the confusion of `operator=` for a Matrix and it is fine to close this.

I sent this PR just because the result in #563 looks quite surprising for me.

eddelbuettel:
All good.

PR #568, opened 2016-10-24
Range sugar uses R_xlen_t as start/end type

thirdwing:

PR #570, opened 2016-10-26
fix range sugar and add unit test (fix error report in #568)

thirdwing:


kevinushey:
LGTM!

eddelbuettel:
Here too.

eddelbuettel:
Just to report back: After this one got '801 passed, eight failed'.  Of the eight one was spurious (a runaway simulation loop, killed after two hours) and one was a new package with what looks like an internal issue.

I ran this set with the old date(time) vectors, that removed two known issues.

PR #573, opened 2016-10-30
updated one unit test for new datetime vectors

eddelbuettel:
treats Inf slightly better
also updated capabilities() so that we can check for it
made one purely cosmetic edit to capabilities and update the one use case

thanks to @dcdillon for the heads-up on the test failing when new datetime vectors were defined in

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -1,9 +1,8 @@
 // -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-
-// jedit: :folding=explicit:
 //
 // api.cpp: Rcpp R/C++ interface class library -- Rcpp api
 //
-// Copyright (C) 2012 - 2015  Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2012 - 2016  Dirk Eddelbuettel and Romain Francois
 //
 // This file is part of Rcpp.
 //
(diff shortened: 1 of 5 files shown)


eddelbuettel:
@kevinushey @thirdwing @jjallaire -- if you get a chance to look at this one which is pretty innocent but extends one test to cope with the 'new' Datetime vectors

src/api.cpp
```diff
@@ -227,7 +233,8 @@ SEXP rcpp_capabilities() {
     SET_STRING_ELT(names, 8, Rf_mkChar("long long"));
     SET_STRING_ELT(names, 9, Rf_mkChar("C++0x unordered maps"));
     SET_STRING_ELT(names, 10, Rf_mkChar("C++0x unordered sets"));
-    SET_STRING_ELT(names, 11, Rf_mkChar("Full C++11 support"));
+    SET_STRING_ELT(names, 11, Rf_mkChar("full C++11 support")); // switched from 'Full' to 'full'
```
eddelbuettel: Purely aesthetics which I usually try not to influence decision making.  All other sentences are lower-case too.

The flag came from the (otherwise completely unused) PR by Christian in 2014 which predates _formal_ C++11 support in R which is now 'a given' so the test was never used anywhere, nor do I expect it to be used.   There is /one/ instance in the package which I changed.

eddelbuettel:
@thirdwing @kevinushey :   thanks for the feedback, now merged and survived Travis

Issue #574, opened 2016-10-31
Signed / unsigned comparisons in IndexHash


nathan-russell:
When compiled with `-Wall` (GCC 4.9.2), the following generates warnings for me: 

```c++
#include <Rcpp.h>

// [[Rcpp::export]]
int usize(Rcpp::IntegerVector x) {
    return Rcpp::unique(x).size();
}
```

> /home/nathan/R/x86_64-pc-linux-gnu-library/3.4/Rcpp/include/Rcpp/hash/IndexHash.h:173:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
>                if (addr == m) {

--------

`IndexHash` has some `int` [members](https://github.com/RcppCore/Rcpp/blob/adae3fe4eb094eb45419b35c49c3f6fd5071267d/inst/include/Rcpp/hash/IndexHash.h#L119) and for the most part seems to stick to signed-signed comparison; [this block](https://github.com/RcppCore/Rcpp/blob/adae3fe4eb094eb45419b35c49c3f6fd5071267d/inst/include/Rcpp/hash/IndexHash.h#L170-L176) being an exception which results in the above warning. There is another such comparison [done here](https://github.com/RcppCore/Rcpp/blob/adae3fe4eb094eb45419b35c49c3f6fd5071267d/inst/include/Rcpp/hash/IndexHash.h#L189-L195) as well. 

Is anyone else able to reproduce this?

eddelbuettel:
Confirmed on Ubuntu 16.04 with g++ 5.4.0:

```
/usr/lib/R/bin/R CMD SHLIB -o 'sourceCpp_7.so'  'nr.cpp'  
ccache g++ -I/usr/share/R/include -DNDEBUG    -I"/usr/local/lib/R/site-library/Rcpp/include" -I"/tmp"    -fpic  -g -O3 -Wall -pipe -Wno-unused -pedantic -c nr.cpp -o nr.o
In file included from /usr/local/lib/R/site-library/Rcpp/include/Rcpp/hash/hash.h:27:0,
                 from /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/sugar.h:28,
                 from /usr/local/lib/R/site-library/Rcpp/include/Rcpp.h:70,
                 from nr.cpp:1:
[...]

PR #575, opened 2016-11-03
Fixed signed comparison in IndexHash; unit tests for unique()

nathan-russell:
This addresses the compiler warning described in #574, which is also present in `SelfHash`. Unit tests for the sugar function `unique` are also added.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2016-11-03  Nathan Russell     <russell.nr2012@gmail.com>
+
+        * inst/include/Rcpp/hash/IndexHash.h: Add casts to eliminate 
+        signed / unsigned comparison warning
+        * inst/include/Rcpp/hash/SelfHash.h: Idem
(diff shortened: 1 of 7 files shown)


eddelbuettel:
Sweet.  Looks good from a quick (and obviously superficial) glance.

nathan-russell:
I actually didn't mean for these changes to get lumped into the existing PR, but hopefully it doesn't turn out to be an issue.

eddelbuettel:
I noticed that too. No big deal --  I think you could have avoided it by starting not only from your fork but with the fork from a distinct branch,

thirdwing:
LGTM. Although this can be in two separate PR.

eddelbuettel:
I don't think it is worth splitting the PRs.  I'll be happy to merge as is.  We don't obsess too much over about squashing all PRs into single commits etc pp.

But of course if someone feels strongly enough about it to make all the changes I won't ignore them...

PR #577, opened 2016-11-05
Added constructor to instantiate newDatetimeVector from VectorBase.

dcdillon:
This fixes the issue where

```c++
#include <Rcpp.h>


// [[Rcpp::export]]
Rcpp::newDatetimeVector test(Rcpp::newDatetimeVector v) {
    Rcpp::newDatetimeVector x = v + 2.;
        return x;
}
```

failed to compile.

It, however, is likely that other issues will crop up at some point due to other missing constructors (since `Vector` has so many and `newDatetimeVector` has so few).  I'd like opinions on whether the rest of the `Vector` constructors should be ported to `newDatetimeVector` at this time, or if we should just wait and see what breaks and port them one at a time.

The same applies to `operator=`, as there don't exist any `operator=` functions that return `newDateTimeVector &`.

### DIFF
--- a/inst/include/Rcpp/date_datetime/newDatetimeVector.h
+++ b/inst/include/Rcpp/date_datetime/newDatetimeVector.h
@@ -28,10 +28,15 @@ namespace Rcpp {
 
     class newDatetimeVector : public NumericVector {
     public:
+        template <int RTYPE, bool NA, typename VEC>
+        newDatetimeVector( const VectorBase<RTYPE,NA,VEC>& other ) :
+            NumericVector(other) {}
+        
         newDatetimeVector(SEXP vec, const char* tz = "") :
             NumericVector(vec) {
             setClass(tz);
         }
+        
         newDatetimeVector(int n, const char* tz = "") :
             NumericVector(n) {
             setClass(tz);


eddelbuettel:
Thank you.  We probably want to 
a) toss `operator=` in while we're at it, and 
b) do the same for newDatevector.
If nobody gets to it I can do it while riding the train downstate next Thursday.

dcdillon:
I'll add that stuff tonight.

eddelbuettel:
Maybe get `newDatetimeVector` right first, and then add things to `newDatevector` ?

dcdillon:
Seems fair.  The one thing that bothers me is that it's likely that some weird issues will bite us later on with missing ctors and assignments, but I also don't want to create a huge PR of largely untested code.

eddelbuettel:
What is your timeline for adding `operator=()`?  Shall I merge now and we revisit extension later?

dcdillon:
Sorry.  I ran into a weird issue on my machine and was getting check fails that I didn't understand and just haven't gotten back to it.  Feel free to merge this and I'll get to the rest when I can

eddelbuettel:
Yes, let's do that.

PR #585, opened 2016-11-12
Feature/no sugar tonight in my coffee

### LINKED ISSUE #584: Rcpp no longer builds with RCPP_NO_SUGAR defined
This is mostly due to the fact that `RNGScope` is defined inside `inst/include/Rcpp/stats/random/random.h` and is used outside the stats directory.  There is also a small issue with the `Matrix op scalar` functions in `inst/include/Rcpp/vector/Matrix.h`.

Moving the definition of `RNGScope` outside of stats (to its own file) and including it unconditionally will fix the former.  Protecting the `Matrix op scalar` functions with `#ifndef RCPP_NO_SUGAR` will fix the latter.

dcdillon:
Fixes #584 by the method detailed in the issue report.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,7 @@
         of RNGScope
         * inst/include/Rcpp/vector/Matrix.h: Protected Matrix op Scalar
         definitions with #ifndef RCPP_NO_SUGAR
+
 2016-11-10  Dirk Eddelbuettel  <edd@debian.org>
 
         * inst/include/Rcpp/date_datetime/newDateVector.h: Added constructor
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2016-11-13  Daniel C. Dillon   <dcdillon@gmail.com>
+        * inst/include/Rcpp/RNGScope.h: Moved definition of RNGScope
+        to its own file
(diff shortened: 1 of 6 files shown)


thirdwing:
I think we can merge and it will be better if @dcdillon can add some change log.

dcdillon:
Modified `inst/NEWS.Rd`.  Hope that's what you meant.

eddelbuettel:
Pretty much every repo has a file [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog). Thought you'd noticed ;-)

dcdillon:
Must be blind.  I looked!  Fixing.

dcdillon:
Would it be reasonable to add a 'no_sugar' plugin so this can be turned on easily on a per-source file basis?

eddelbuettel:
Maybe.  I don't think I'd have use for it.  Why would you turn sugar off, and why can't a `#define` before `#include <Rcpp.h>` do it?

dcdillon:
Just looks cleaner than arbitrarily sticking `#define` in your `.cpp` files.  Not a huge issue, just a thought.

I turn sugar off to get better handling of matrix operations without causing ambiguities in operator calls.

PR #587, opened 2016-11-13
Added assignment operators to newDateVector and newDatetimeVector

dcdillon:


dcdillon:
Might be nice to have a plugin to turn this functionality on as well.

eddelbuettel:
>  this functionality

What are you talking about, exactly?  Is it this:

```bash
edd@max:~/git/rcpp(master)$ git diff
diff --git a/inst/include/Rcpp.h b/inst/include/Rcpp.h
index d365958..fbae1c1 100644
--- a/inst/include/Rcpp.h
+++ b/inst/include/Rcpp.h
@@ -57,6 +57,7 @@
 #include <Rcpp/DataFrame.h>
 
 //  #define RCPP_NEW_DATE_DATETIME_VECTORS 1
+#define RCPP_NEW_DATE_DATETIME_VECTORS 1
 #include <Rcpp/date_datetime/date_datetime.h>
 
 #include <Rcpp/Na_Proxy.h>
edd@max:~/git/rcpp(master)$ 
```

That is a single `-D...` argument to the compiler in `~/.R/Makevars`, as an env.var or, like here, as edit.  Do we really need a plugin for that?

dcdillon:
I just think it makes it cleaner than relying on the developer for it.  Then we can also change how it's implemented behind the scenes (if needed) and change the plugin accordingly to get the right build configuration.

eddelbuettel:
Very nice that you added unit tests too.

eddelbuettel:
Calling for reviews or thumbs up/down ... might be nice to get this in before I run a round of tests tomorrow.

@kevinushey @thirdwing @nathan-russell @jjallaire

kevinushey:
LGTM, although I'm not as familiar with datetime handling (+ the associated traps inherent) with R.

eddelbuettel:
There is not much too it, but it _looks_ super-confusing. `POSIXct` is just out friend `time_t`, generalised by BDR et al to be a 53 digit double (hence with about microsec precision from the remaining bits). It generally stores in UTC as one should and uses a TZ attribute for offset (when formatting for humans, say).

I didn't understand what attributes did eight or nine years ago so the first iteration of these were `std::vector<>` containers which is very _very_ un-`SEXP`.  So now we use `NumericVector` with a class attribute, and the world is a better place.  There aren't too many packages using these types, and some I am involved with so the transition should be easy.   (And `POSXlt`, as you know, is just lists of `struct tm` components _presenting the same content_.)

PR #597, opened 2016-11-22
fixes a small 'thinko'

eddelbuettel:
only whitespace in newDateVector
rolls minor version
updates debian/changelog

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2016-11-22  Dirk Eddelbuettel  <edd@debian.org>
+
+        * DESCRIPTION (Date, Version): roll minor version
+
+        * inst/include/Rcpp/date_datetime/newDatetimeVector.h (Rcpp): Small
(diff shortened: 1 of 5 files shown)


thirdwing:
Let's merge this small PR first?

eddelbuettel:
No worries -- I don't really care about the order.  In a way, @coatless was first, and @jimhester has a more important fix. 

This particular one is my bad, and I noticed when trying to update something in the [RcppExamples](https://github.com/eddelbuettel/rcppexamples) package.  This is not urgent as it depends on the [toggle in Rcpp.h being defined](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp.h#L59) so it is not yet biting anyone unless that is one (or someone turns it one) and they try `operator+` on a new `DatetimeVector`.  

So in a nutshell it doesn't really matter but if you want it merged first go ahead.  Else I'll get to in the new few days.

PR #599, opened 2016-11-27
added format() methods for Date and Datetime

eddelbuettel:
also rolled dev version in config.h

also adds `operator<<()` which doesn't seem to get picked up by compiler, ie we get the underlying date number or seconds since epoch.

### DIFF
--- a/inst/include/Rcpp/date_datetime/Date.h
+++ b/inst/include/Rcpp/date_datetime/Date.h
@@ -106,8 +106,12 @@ namespace Rcpp {
             char txt[32];
             struct tm temp = m_tm;
             temp.tm_year -= baseYear();    // adjust for fact that system has year rel. to 1900
-            ::strftime(txt, 31, fmt, &temp);
-            return std::string(txt);
+            int res = ::strftime(txt, 31, fmt, &temp);
+            if (res == 0) {
+                return std::string("");
+            } else {
+                return std::string(txt);
+            }
         }
 
         friend inline std::ostream &operator<<(std::ostream & os, const Date d);
--- a/inst/include/Rcpp/date_datetime/Date.h
+++ b/inst/include/Rcpp/date_datetime/Date.h
@@ -110,11 +110,7 @@ namespace Rcpp {
             return std::string(txt);
         }
 
-        inline std::ostream &operator<<(std::ostream & s) const {
-            s << this->format() << std::endl;
-            return s;
-        }
-
+        friend inline std::ostream &operator<<(std::ostream & os, const Date d);
 
     private:
         double m_d;                 // (fractional) day number, relative to epoch of Jan 1, 1970
@@ -166,6 +162,11 @@ namespace Rcpp {
     inline bool   operator>=(const Date &d1, const Date& d2) { return d1.m_d >= d2.m_d; }
     inline bool   operator<=(const Date &d1, const Date& d2) { return d1.m_d <= d2.m_d; }
     inline bool   operator!=(const Date &d1, const Date& d2) { return d1.m_d != d2.m_d; }
+
+    inline std::ostream &operator<<(std::ostream & os, const Date d) {
+        os << d.format();
+        return os;
+    }
 
     namespace internal {
 
--- a/inst/include/Rcpp/date_datetime/Date.h
+++ b/inst/include/Rcpp/date_datetime/Date.h
@@ -102,6 +102,20 @@ namespace Rcpp {
             return m_d;
         }
 
+        inline std::string format(const char *fmt = "%Y-%m-%d") const {
+            char txt[32];
+            struct tm temp = m_tm;
+            temp.tm_year -= baseYear();    // adjust for fact that system has year rel. to 1900
+            ::strftime(txt, 31, fmt, &temp);
+            return std::string(txt);
+        }
+
+        inline std::ostream &operator<<(std::ostream & s) const {
+            s << this->format() << std::endl;
+            return s;
+        }
+
+
     private:
         double m_d;                 // (fractional) day number, relative to epoch of Jan 1, 1970
         struct tm m_tm;             // standard time representation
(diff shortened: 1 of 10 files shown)


nathan-russell:
This looks good to me; my only suggestion would be possibly dropping the trailing `<< std::endl;`, as there may be cases where the extra newline and call to `flush` is not desirable.

eddelbuettel:
Thanks for looking at these.  These were so far mostly for seeing what it would -- where I really want them is in the vector case of the new classes.   And yes, the forced newline would be annoying there.  For an individual item it may still make sense.   Not sure yet how to balance it, and when to line break for vectors (every N maybe?).

eddelbuettel:
Also annoying:  

```r
R> library(Rcpp)
R> cppFunction('void foo(Date d) { Rcpp::Rcout << d << std::endl; } ')
R> foo(Sys.Date())
17132
R> cppFunction('void bar(Date d) { Rcpp::Rcout << d.format() << std::endl; } ')
R> bar(Sys.Date())
2016-11-27
R>
```

I'd love for `foo()` to do what `bar()` does.

eddelbuettel:
There is something else going on.  I don't even think the new `operator<<()` gets called there.

inst/include/Rcpp/date_datetime/Datetime.h
```diff
@@ -68,6 +68,20 @@ namespace Rcpp {
...
+        inline std::string format(const char *fmt = "%Y-%m-%d %H:%M:%S") const {
+            char txtiso[64], txtsec[64];
+            time_t t = static_cast<time_t>(std::floor(m_dt));
+            struct tm temp = *localtime(&t); // localtime, not gmtime
+            ::strftime(txtiso, 63, fmt, &temp);
+            ::snprintf(txtsec, 63, "%s.%06d", txtiso, m_us);
+            return std::string(txtsec);
+        }
+
+        inline std::ostream &operator<<(std::ostream & s) const {
```
coatless: Naive but: `Rcpp::Rcout` or `Rcpp::Rstreambuf` instead of `std::ostream`?

eddelbuettel: No, do a `ag operator\<\<` inside the repo directory.  All `operator<<` go to ostream from which they get redirected.

But @dcdillon just reminded me that I can't do this as a member function.  So it was a null-op ;-)

nathan-russell:
I think `foo` is failing because `operator double()` is being called instead of `operator<<`. Changing the definition to this works for me: 

```c++
        // inline std::ostream &operator<<(std::ostream & s) const {
        //     s << this->format() << std::endl;
        //     return s;
        // }
        friend std::ostream& operator<<(std::ostream& os, const Date& d) {
            os << d.format() << std::endl;
            return os;
        }
```

eddelbuettel:
Yes, @dcdillon and I chatted about that.  I now have this working

```
  // end of public block
  friend inline std::ostream &operator<<(std::ostream & s, const Date d);

  // outside class
    inline std::ostream &operator<<(std::ostream & os, const Date d) {
        os << d.format();
        return os;
    }

```

and that worketh even ;-)
```
R> library(Rcpp)
R> cppFunction('void foo(Date d) { Rcpp::Rcout << d << std::endl; } ')
R> foo(Sys.Date())
2016-11-27
R>
```

eddelbuettel:
If anybody else would like to 'bless' (or even "review") it I'd feel better before merging my own PR :)

inst/include/Rcpp/date_datetime/Date.h
```diff
@@ -102,6 +102,16 @@ namespace Rcpp {
             return m_d;
         }
 
+        inline std::string format(const char *fmt = "%Y-%m-%d") const {
+            char txt[32];
+            struct tm temp = m_tm;
+            temp.tm_year -= baseYear();    // adjust for fact that system has year rel. to 1900
+            ::strftime(txt, 31, fmt, &temp);
```
kevinushey: Should we check the return value of `::strftime()` here? What happens to `txt` if the user passes in a `fmt` string not recognized by `::strftime()`?

inst/include/Rcpp/date_datetime/Date.h
```diff
@@ -102,6 +102,16 @@ namespace Rcpp {
...
+            return std::string(txt);
```
kevinushey: Since `txt` isn't zero-initialized (it's a plain C array), it seems like it could be possible that no null byte exists at the end of `txt`. Or does `::strftime()` write a null byte at the end? What if the output of `::strftime()` doesn't fit in `txt`?

inst/include/Rcpp/date_datetime/Datetime.h
```diff
@@ -68,6 +68,17 @@ namespace Rcpp {
 
         operator double()  const { return m_dt; }
 
+        inline std::string format(const char *fmt = "%Y-%m-%d %H:%M:%S") const {
+            char txtiso[64], txtsec[64];
+            time_t t = static_cast<time_t>(std::floor(m_dt));
+            struct tm temp = *localtime(&t); // localtime, not gmtime
+            ::strftime(txtiso, 63, fmt, &temp);
```
kevinushey: Similar questions here re: zero-initialization, checking of return value?

kevinushey:
Some paranoid checks re: state of the `char` array used in conjunction with `::strftime()` but otherwise LGTM.

inst/include/Rcpp/date_datetime/Date.h
eddelbuettel: Then `strftime()` returns zero.  I guess we could check and return an explicit `std::string("")`.  Would be cleaner.

inst/include/Rcpp/date_datetime/Date.h
eddelbuettel: Per previous point and `man strftime`, by catching a return of zero we should catch that.

eddelbuettel:
Thanks, @kevinushey . I extended both `format()` functions to be a wee bit more explicit on the return values.

kevinushey:
Okay, LGTM!

eddelbuettel:
Thank you.  I just added two (small) sets of unit tests which I'll in a minute and then I'll merge.

PR #601, opened 2016-12-03
more tests, and some #nocov tags

eddelbuettel:


eddelbuettel:
Anyone?

thirdwing:
I have to say this is something I know really little, but LGTM.

eddelbuettel:
I am not expert on the [covr](https://cran.r-project.org/package=covr) package either but slowly getting a feel for it via a few other (smaller!!) packages.  It wraps very cleverly around calls to code and 'counts' -- for R by modifying the code, for compiled code via `gcov`.   It helps some -- became aware of a regression on the weekend (in another and smaller package) where a line all of sudden was no longer covered when I changed code, and indeed, an argument was ignored when it should not have been...

eddelbuettel:
Also: if you use Chrome, the plugin is neat as you then see the code at GitHub itself "coloured" green and red depending on whether it has coverage or not.

eddelbuettel:
No further complains to will merge.

Issue #612, opened 2016-12-14
Development version broken for STRICT_R_HEADERS


jeroenooms:
For some packages we need to `#define STRICT_R_HEADERS` because macros like `Calloc`, `Free` and `ERROR` conflict with identically named macros in other libraries. 

This is working fine in Rcpp from CRAN, however in the development version of Rcpp I now get errors on Windows. It looks like a recent change in [sugar/functions/sample.h](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/sugar/functions/sample.h) has started using S-style `Calloc` and `Free` instead of stdlib `calloc` and `free`. 

Is this intentional? It will break many packages that use `<Rcpp.h>` in conjunction with `STRICT_R_HEADERS`.

```r
devtools::install_github("RcppCore/Rcpp")
devtools::install_github("ropensci/tesseract")
```

eddelbuettel:
It is an oversight by our newest team member.  We'll fix.  Thanks for the heads up.

nathan-russell:
@jeroenooms Apologies, I was trying to be consistent with the implementation in R's [`random.c`](https://github.com/wch/r-source/blob/trunk/src/main/random.c#L364-L365) file, but was unaware of this. Wrapping those sections with the appropriate macro guards in `sample.h`, e.g. 

```c++
#ifdef STRICT_R_HEADERS
        HL = static_cast<int*>(R_Calloc(n, int));
        q = static_cast<double*>(R_Calloc(n, double));
#else
        HL = static_cast<int*>(Calloc(n, int));
        q = static_cast<double*>(Calloc(n, double));
#endif
```

seemed to work, as I was able to compile `tesseract` successfully. Will this be sufficient to address the problem?

jeroenooms:
Why can't you use stdlib malloc?

eddelbuettel:
Because it is data that may go back to R.

But do a recursive grep (or my _fav_, `ag` from the silversearcher) for alloc.  We do use `Rf_allocVector` a lot.  Why not use that?   The `#ifdef` above is ... weird.  We're not force to translate the code from R base line by line.

eddelbuettel:
My code reviews skills clearly need honing too.

jeroenooms:
[...]

PR #613, opened 2016-12-14
Use STRICT_R_HEADERS in Rcpp.h

jeroenooms:
This is only a simple suggestion. It defines `STRICT_R_HEADERS` inside `Rcpp.h` to prevent issues like https://github.com/RcppCore/Rcpp/issues/612 in the future.

I would be much in favor of setting `STRICT_R_HEADERS` unconditionally, but that might break an existing package or two (or perhaps not).

### DIFF
--- a/inst/include/Rcpp.h
+++ b/inst/include/Rcpp.h
@@ -23,6 +23,11 @@
 #ifndef Rcpp_hpp
 #define Rcpp_hpp
 
+#ifndef STRICT_R_HEADERS
+#define STRICT_R_HEADERS
+#define RCPP_TEMP_STRICT_R_HEADERS
+#endif
+
 /* it is important that this comes first */
 #include <RcppCommon.h>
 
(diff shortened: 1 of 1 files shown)


eddelbuettel:
*Please* don't do pull requests before discussion.

I though of this as soon as you raised the issue.  But we were just burned somewhere else by another package doing something 'too widely' in its header so I am against _unconditional inclusion_.

jeroenooms:
Well the PR is not unconditionally. If it was not previously defined, it will undefine it at the end. So `STRICT_R_HEADERS` is only defined inside `Rcpp.h`.

eddelbuettel:
Well but as the proponent of the proposal, it is would appear to be your responsibility to assess how widely it affects client packages.  For that you would need to run reverse depends.

If it were unconditional, it wouldn't fail Travis either.  So you have your counter proof [right here](https://travis-ci.org/RcppCore/Rcpp/builds/183965911), no?

Would you be terribly upset if I closed this here and we carry on over in #614?

jeroenooms:
It fails because of https://github.com/RcppCore/Rcpp/issues/612 which is precisely the goal of the PR. Once the bug has been fixed it will no longer fail. But feel free to close I was just trying to discuss this suggestion here. You can move it to another issue.

eddelbuettel:
I do hate red in Travis so submitting this now before we got to the fix needed to pass strikes me as ... suboptimal.

PR #615, opened 2016-12-14
Compile Rcpp with STRICT_R_HEADERS on both Unix and Windows

jeroenooms:
Let's see what Travis thinks of this...


jeroenooms:
OK so this works, it fails with the expected error. Once https://github.com/RcppCore/Rcpp/issues/612 has been fixed, you can consider applying this fix.

eddelbuettel:
Not a chance as it removes a file I want to keep.

PR #621, opened 2016-12-31
Feature/update date code

eddelbuettel:
`src/Date.cpp` basically dates from 2010.  

It contains some stub code we don't use (ie to load a zonefile) but maybe could / should, but that is for another day.

This PR mostly brings in an update to get to what R uses there.  Ultimately I should farm this out but one attempt in repo [rapidatetime](https://github.com/eddelbuettel/rapidatetime) got suck.  Maybe I'll get back to this.

The PR should be harmless.  It tests out fine, and the branch also supports tests of RQuantLib which is one the few packages actually using `Rcpp::Date` a little...

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2016-12-30  Dirk Eddelbuettel  <edd@debian.org>
+
+        * src/Date.cpp: Synchronized internal code with R
+
 2016-12-26  Dirk Eddelbuettel  <edd@debian.org>
 
         * R/Attributes.R: Added #nocov markers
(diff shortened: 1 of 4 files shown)


eddelbuettel:
@kevinushey @thirdwing @jjallaire @nathan-russell -- Happy New Year!

In case any one of you has a minute or two to spare, a quick code review here would be appreciated.  
Should not bite in any way.  New unit test tests in the this branch and in main.

eddelbuettel:
And I just added two more.

kevinushey:
LGTM!

PR #622, opened 2016-12-31
Fix for non-symmetric matrix diagonal fill (Closes #619)

### LINKED ISSUE #619: Matrix Diagonal Fill Bugged for Nonsymmetric
Consider:

```r
Rcpp::cppFunction("NumericMatrix fdiag(NumericMatrix x) {
    x.fill_diag(0.0);
    return x;
}")
```

The issue arises with the [`.fill_diag()`'s `fill_diag__dispatch()` method in `Matrix.h`](https://github.com/RcppCore/Rcpp/blob/e4ca728131fbec1c08e594a84ba5279708297ffa/inst/include/Rcpp/vector/Matrix.h#L176-L198) primarily with a nonsymmetric matrix under either n > p or n < p. 

## Case 1: Nonsymmetric (n < p)

```r
m1 <- matrix(1:10, nrow = 2, ncol = 5, byrow = TRUE)
fdiag(m1)
```

```
     [,1] [,2] [,3] [,4] [,5]
[1,]    0    2    3    0    5
[2,]    6    7    8    9   10
```

Expected:

```
     [,1] [,2] [,3] [,4] [,5]
[1,]    0    2    3    4    5
[2,]    6    0    8    9   10
```

## Case 2: Nonsymmetric (n > p)

```r
m3 <- matrix(1:10, nrow = 5, ncol = 2, byrow = TRUE)
fdiag(m3)
```
```
     [,1] [,2]
[1,]    0    2
[2,]    3    4
[3,]    5    6
[4,]    0    8
[5,]    9   10
```

Expected:

```
     [,1] [,2]
[1,]    0    2
[2,]    3    0
[3,]    5    6
[4,]    7    8
[5,]    9   10
```

## Case 3: n = p

Though, the square matrix iteration works correctly.

```r
m3 <- matrix(1:9, nrow = 3, ncol = 3, byrow = TRUE)
fdiag(m3)
```

```
     [,1] [,2] [,3]
[1,]    0    2    3
[2,]    4    0    6
[3,]    7    8    0
```

coatless:
Fix for the non-symmetric case of the diagonal fill. 

Adds unit tests to the diagonal fill feature as well.

Per discussion in #619, this PR moves the matrix diagonal fill away from using an iterator solution to using the matrix subset operator.

### DIFF
--- a/inst/include/Rcpp/vector/Matrix.h
+++ b/inst/include/Rcpp/vector/Matrix.h
@@ -175,25 +175,25 @@ private:
 
     template <typename U>
     void fill_diag__dispatch( traits::false_type, const U& u) {
-          Shield<SEXP> elem( converter_type::get( u ) ) ;
-        int n = Matrix::ncol() ;
-        int offset = n +1 ;
-        iterator it( VECTOR::begin()) ;
-        for( int i=0; i<n; i++){
-            *it = ::Rf_duplicate( elem );
-            it += offset;
+        Shield<SEXP> elem( converter_type::get( u ) );
+
+        R_xlen_t bounds = ( Matrix::nrow() < Matrix::ncol() ) ?
+                             Matrix::nrow() : Matrix::ncol();
+        
+        for (R_xlen_t i = 0; i < bounds; ++i) {
+            (*this)(i, i) = elem;
         }
     }
 
     template <typename U>
     void fill_diag__dispatch( traits::true_type, const U& u) {
-          stored_type elem = converter_type::get( u ) ;
-        int n = Matrix::ncol() ;
-        int offset = n + 1 ;
-        iterator it( VECTOR::begin()) ;
-        for( int i=0; i<n; i++){
-            *it = elem ;
-            it += offset;
+        stored_type elem = converter_type::get( u );
+		
+        R_xlen_t bounds = ( Matrix::nrow() < Matrix::ncol() ) ?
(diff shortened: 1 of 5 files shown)


inst/include/Rcpp/vector/Matrix.h
```diff
@@ -175,25 +175,25 @@ class Matrix : public Vector<RTYPE, StoragePolicy>, public MatrixBase<RTYPE, tru
...
-        int offset = n +1 ;
-        iterator it( VECTOR::begin()) ;
-        for( int i=0; i<n; i++){
-            *it = ::Rf_duplicate( elem );
-            it += offset;
+        Shield<SEXP> elem( converter_type::get( u ) );
+
+        R_xlen_t bounds = ( Matrix::nrow() < Matrix::ncol() ) ?
+                             Matrix::nrow() : Matrix::ncol();
+
```
eddelbuettel: I was just squinting at this (and mostly wondering if I had written it on one line) when I realized that

```
 R_xlen_t bounds = std::min( Matrix::nrow(), Matrix::ncol() );
```

is probably even easier :)    (Subjec to testing in whether `stdLLmin()` likes the R type etc pp)

eddelbuettel: I guess one could argue that the ternary operator may be lighter-weight that a function call (which, I bet, ends up inlined).  I use that operator a fair amount for alternates too.   It just seemed more compact...

PR #623, opened 2017-01-04
two refinements for Date.cpp required on Windows

eddelbuettel:
We need one change to compile (!!) on Windows, and on test (using six digits precision) fails.

*Edit* Turns out that was a TZ issue instead.


eddelbuettel:
@kevinushey @nathan-russell @thirdwing @jjallaire :  If you have a moment ... I had botched Windows builds, this restores it and is otherwise pretty minimal.

PR #626, opened 2017-01-08
Adds Arg(X) sugar function (closes #625)

### LINKED ISSUE #625: Missing complex number function Arg(X)
[Complex sugar ops](https://github.com/RcppCore/Rcpp/blob/8ac38d7f01494eb44673e452ebf709839864ffdf/inst/include/Rcpp/sugar/functions/complex.h) have support for the following R equivalents:

- `Re`
- `Im`
- `Mod`
- `Conj`

but no support for the `Arg` function that determines the angle between the real and complex plane (counter-clockwise).

 Given the availability of the cpp port of the fortran `atan2` inverse function, the [base R imp of `Arg`](https://github.com/wch/r-source/blob/2a0f317d5842bd2645e770c566375a0dafae2542/src/main/complex.c#L281-L321) is easy to mirror. Two-liner addition to the complex header:

```cpp
inline double complex__Arg( Rcomplex x ){ return ::atan2(x.i, x.r); }
RCPP_SUGAR_COMPLEX( Arg, double )
```

coatless:


inst/include/Rcpp/sugar/functions/complex.h
```diff
@@ -68,6 +68,7 @@ inline double complex__Re( Rcomplex x){ return x.r ; }
 		y.i = -x.i ;
 		return y ;
 	}
+    inline double complex__Arg( Rcomplex x ){ return ::atan2(x.i, x.r); }
```
eddelbuettel: The vertical alignment seems to a little out of whack but we can fix that later.

PR #635, opened 2017-01-17
Protect tm_gmtoff on Solaris (closes #634)

### LINKED ISSUE #634: Date.cpp update breaks compilation on Solaris
Per https://www.r-project.org/nosvn/R.check/r-patched-solaris-sparc/Rcpp-00check.html:

```
/opt/csw/bin/g++ -I/home/ripley/R/gcc/include -DNDEBUG -I../inst/include/ -I/opt/csw/include -I/usr/local/include    -fPIC  -O2 -mcpu=niagara2  -c Date.cpp -o Date.o
Date.cpp: In function ‘std::tm* Rcpp::timesub(const time_t*, int_fast32_t, const Rcpp::state*, std::tm*)’:
Date.cpp:1352:14: error: ‘struct std::tm’ has no member named ‘tm_gmtoff’
         tmp->tm_gmtoff = offset;
              ^
*** Error code 1
make: Fatal error: Command failed for target `Date.o'
```

Ditto for the other Solaris build.

eddelbuettel:
Should work. As usual, cannot test.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2017-01-17  Dirk Eddelbuettel  <edd@debian.org>
+
+        * DESCRIPTION (Version, Date): Roll minor version and Date
+
+        * inst/include/Rcpp/config.h (RCPP_DEV_VERSION): Roll minor
+
+        * src/Date.cpp (Rcpp): Do not access tm_gmtoff on Solaris
+
 2017-01-14  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION: Release 0.12.9
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 0.12.9
-Date: 2017-01-14
+Version: 0.12.9.1
+Date: 2017-01-17
 Author: Dirk Eddelbuettel, Romain Francois, JJ Allaire, Kevin Ushey, Qiang Kou,
  Nathan Russell, Douglas Bates and John Chambers
 Maintainer: Dirk Eddelbuettel <edd@debian.org>
--- a/inst/include/Rcpp/config.h
+++ b/inst/include/Rcpp/config.h
@@ -30,7 +30,7 @@
 #define RCPP_VERSION Rcpp_Version(0,12,9)
 
 // the current source snapshot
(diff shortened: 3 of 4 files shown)


eddelbuettel:
@jjallaire @kevinushey @thirdwing @nathan-russell -- if you have a moment, should be easy to look over

kevinushey:
Looks good to me. Since `tm_gmtoff` is a GNU library extension and not part of ISO C (see: https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html) the conditional might best be expressed as `#ifdef __GNUC__`.

eddelbuettel:
That's a good idea. Lemme try that quickly.

eddelbuettel:
Actually, the exact wording is

> The tm_gmtoff field is derived from BSD and is a GNU library extension; it is not visible in a strict ISO C environment.

implying the BSDs have it too. 

A [Python discussion](https://mail.python.org/pipermail/datetime-sig/2015-September/000955.html) stated

> Most UNIX platforms extend struct tm to include tm_gmtoff and tm_zone
fields that contain the current UTC offset in seconds and the zone
abbreviation.

so maybe it is better as is to cast a wider net?

Ahh, and using `#if defined(__GNUC__)` is also true on MinGW and fails there.  So won't do this--and merge as is.

kevinushey:
Oof, thanks for double-checking!

eddelbuettel:
"Trust and verify" is a beautiful motto :)

coatless:
"Trust, but expect the unexpected" is equally beautiful. I'll try building on a solaris vm later.

kevinushey:
FWIW, Jeroen has a Solaris VM (for VMWare) available here: https://github.com/jeroenooms/solarisvm

eddelbuettel:
It's misleading. Not actually a VM. A "blog post" or README for how to create one.

thirdwing:
There is a download link in the blog. I tried it before.

coatless:
Unfortunately, the build is **broken**. 

Please see screenshot below.

solaris_bugfix

eddelbuettel:
That's because I clearly do not know the difference between OS and compiler.  

Can you make that line 1350 in `src/Date.cpp` end on

```
#if ! (defined(__MINGW32__) || defined(__MINGW64__) || defined(sun) || defined(__sun))
```

instead of checking for the Sun compiler we do not use?

PR #637, opened 2017-01-19
To compile on solaris

thirdwing:
It should compile now. 

solaris11


eddelbuettel:
Thanks for actually checking.

PR #638, opened 2017-01-21
Adds size attribute aliases for DataFrame (closes #630)

### LINKED ISSUE #630: Mirror Matrix Dimensional Member Functions for DataFrame
The proposal here is to improve API consistency between the 2D classes `Matrix` and `DataFrame` by:

1. deprecating the `.nrows()` function in `DataFrame`  in favor of adding `.nrow()` and `.rows()`
2. introduce the `.ncol()` and `.cols()` functions to `DataFrame` that map to `Vector`'s `size()` and `length()` functions.

However, another interesting proposal would be to remove the size attribute access information provided by `.cols()` and `.rows()` for the `Matrix` class and introduce `.Rows(first, last)` and `.Cols(first, last)` that would provide submatrix views. 

APIs
----

For the [`Matrix`](https://github.com/RcppCore/Rcpp/blob/6f81b4684481dbd9bb554dd95e66725fc3b63a8c/inst/include/Rcpp/vector/Matrix.h#L94-L105) class, the following member functions are defined for dimensional information:

|      Member                 |            Description                               |
|:---------------------|:------------------------------------------|
| `.nrow()`,`.rows()`    | number of rows in a `Matrix`              |
| `.ncol()`,`.cols()`       | number of columns in a `Matrix`           |
| `.size()`,`.length()`    | number of items in a `Matrix`, `Vector`   |

Meanwhile, over in [`DataFrame`](https://github.com/RcppCore/Rcpp/blob/6f81b4684481dbd9bb554dd95e66725fc3b63a8c/inst/include/Rcpp/DataFrame.h#L63-L84) land, the member functions are defined a bit differently:

|      Member            |            Description                               |
|:---------------------|:------------------------------------------|
| `.nrows()`               | number of rows in a `DataFrame`              |
| `.size()`,`.length()`   | number of columns in a `DataFrame`   |

The break in consistency regarding the `nrow()` and `ncol()` within `DataFrame` is problematic as it breaks with those accessor functions available in base R. 

`data.frame` and `length()` note
---

Note, there is no need to change the `length()` function call since the R equivalent returns the number of columns instead of the amount of elements, e.g. 

```r
df = data.frame(a = 1:3, b = 3:1)
length(df)
# 2
```

coatless:


inst/include/Rcpp/DataFrame.h
```diff
@@ -83,6 +83,26 @@ namespace Rcpp{
             return LENGTH(rn);
         }
 
+        // Leave in place for a year
```
eddelbuettel: I don't think we have consensus that we what or need that.

I presume I could get my arm twisted to allow `nrow` and `nrows`; I see nothing but needed pain in removing the old one.

coatless: The goal behind the PR was to align the `DataFrame` class to match with:

1. the _R_ functions `ncol()` and `nrow()`;
2. the `Matrix` class member functions `.nrow()`, `.ncol()`, `.rows()`, `.cols()`.

In an ideal world, the old interface _probably_ should be cleaned up as a result. But, it most certainly can be left in place and the `nrows()` aspect can be _omitted_ from documentation.

By my count, the following ~20 packages on CRAN would be affected:

- `BacArena`
- `Biocomb`
-  `Cyclops`
- `deepboost`
- `dplyr`
- `FeatureHashing`
- `FIT`
- `feather`
- `gaston`
-  `medfate`
- `missDeaths`
- `Rfast`
- `Rborist`
- `raptr`
- `readr`
- `readstat13`
- `reshape2`
- `tweenr`
- `valr`
- `wsrf`

eddelbuettel: I have no interest / time / energy to chase *twenty* package maintainers over essentially nothing.

I have tried to get _five_ to change _an error of theirs_ for three weeks now and only got three.  No, we are not going to break that interface.

Additional accessors are perfectly fine.

eddelbuettel:
I took the language about deprecation out and simplified a little.  Now it is a straight-up extension of the API, plus extra tests, and should not create any trouble.

Second set of eyes anyone?  @thirdwing @nathan-russell @kevinushey @jjallaire ?

kevinushey:
LGTM. As an aside, these APIs should probably be returning `R_xlen_t`, but that's probably for a separate PR.

eddelbuettel:
Good catch.  The actual worker function here still uses `LENGTH()`, and that should be `XLENGTH()` with the commensurate change in returning `R_xlen_t`.

coatless:
The following seems to work:

```cpp
        inline R_xlen_t nrow() const {
            SEXP rn = R_NilValue ;
            SEXP att = ATTRIB( Parent::get__() )  ;
            while( att != R_NilValue ){
                if( TAG(att) == R_RowNamesSymbol ) {
                    rn = CAR(att) ;
                    break ;
                }
                att = CDR(att) ;
            }
            if (Rf_isNull(rn))
                return 0;
            if (TYPEOF(rn) == INTSXP && XLENGTH(rn) == 2 && INTEGER(rn)[0] == NA_INTEGER)
                return abs(INTEGER(rn)[1]);
            return XLENGTH(rn);
        }

        // Offer multiple variants to accomodate both old interface here and signatures in other classes
        inline R_xlen_t nrows() const { return DataFrame_Impl::nrow(); }
        inline R_xlen_t rows()  const { return DataFrame_Impl::nrow(); }

        inline R_xlen_t ncol()  const { return DataFrame_Impl::length(); }
        inline R_xlen_t cols()  const { return DataFrame_Impl::length(); }
```

I'll toss the above in a new issue and then submit a PR with the above change? Or just submit the code change within here and add a new note.

kevinushey:
Sorry, looks like I'm not quite right -- while the number of columns (length) can indeed be a 'long' vector, the `row.names` attribute is still expected to be a 'short' vector. From

https://github.com/wch/r-source/blob/45576f65c403f46e03b88e4e0301bde330aea3be/src/main/attrib.c#L181-L203

The code for extracting the true length from a potentially 'compact' representation is:

	int n = (isInteger(s) && LENGTH(s) == 2 && INTEGER(s)[0] == NA_INTEGER)
	    ? INTEGER(s)[1] : (isNull(s) ? 0 : LENGTH(s));
	ans = ScalarInteger((type == 1) ? n : abs(n));

So R indeed is still expecting an integer number of rows for data frames, if I understand correctly.

eddelbuettel:
Thank you both for following.  So I guess I'll merge as is...

coatless:
Based on @kevinushey's digging, the aliases for columns should have their return type updated to match the definition out of `Vector`.

```cpp
        inline R_xlen_t ncol()  const { return DataFrame_Impl::length(); }
        inline R_xlen_t cols()  const { return DataFrame_Impl::length(); }
```

eddelbuettel:
Thanks for making the change.  This should now be good.

kevinushey:
LGTM as well; thanks!

eddelbuettel:
So in it goes -- thanks to @coatless for doing the work and to @kevinushey for being eagle-eyed as usual.

PR #642, opened 2017-01-31
Fixes for upper_tri and lower_tri with unit tests (closes #641)

### LINKED ISSUE #641: `lower_tri` / `upper_tri` non-functional
Both the `lower_tri` and `upper_tri` sugar functions are non-functional for two reasons:

1. Neither function returns a `Matrix` as CTRP's base is `VectorBase` instead of `MatrixBase`.
2. Attempting to store the result in a `LogicalVector` leads the R process to hang.

After glancing at the source, it appears as if both of functions were DOA and should _not_ have been pushed back with Rcpp 0.8.3. Though, this remark is made in light of the fact that there are no unit tests governing these functions.

---

**The following will _hang_ your R session**

```r
Rcpp::cppFunction('
LogicalVector test(NumericMatrix x){
return upper_tri(x);
}')
x <- matrix(1:4, 4, 5)
test(x)
``` 

---

Example of desired output:

``` r
(x <- matrix(1:4, 4, 5))
#>      [,1] [,2] [,3] [,4] [,5]
#> [1,]    1    1    1    1    1
#> [2,]    2    2    2    2    2
#> [3,]    3    3    3    3    3
#> [4,]    4    4    4    4    4
lower.tri(x)
#>       [,1]  [,2]  [,3]  [,4]  [,5]
#> [1,] FALSE FALSE FALSE FALSE FALSE
#> [2,]  TRUE FALSE FALSE FALSE FALSE
#> [3,]  TRUE  TRUE FALSE FALSE FALSE
#> [4,]  TRUE  TRUE  TRUE FALSE FALSE
```

nathan-russell:
With these changes, 

```c++
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
LogicalMatrix UpperTri(NumericMatrix x, bool diag = false) {
    return upper_tri(x, diag);
}

// [[Rcpp::export]]
LogicalMatrix LowerTri(NumericMatrix x, bool diag = false) {
    return lower_tri(x, diag);
}

/*** R

x <- matrix(1:4, 4, 5)

all.equal(UpperTri(x), upper.tri(x))
# [1] TRUE

all.equal(UpperTri(x, TRUE), upper.tri(x, TRUE))
# [1] TRUE

all.equal(LowerTri(x), lower.tri(x))
# [1] TRUE

all.equal(LowerTri(x, TRUE), lower.tri(x, TRUE))
# [1] TRUE

*/
```

### DIFF
--- a/inst/include/Rcpp/sugar/matrix/lower_tri.h
+++ b/inst/include/Rcpp/sugar/matrix/lower_tri.h
@@ -3,7 +3,7 @@
 // lower_tri.h: Rcpp R/C++ interface class library -- lower.tri
 //
 // Copyright (C) 2010 - 2017 Dirk Eddelbuettel and Romain Francois
-// Copyright (C) 2017        Dirk Eddelbuettel, Romain Francois, and Nathan Russell
+// Copyright (C) 2017    Dirk Eddelbuettel, Romain Francois, and Nathan Russell
 //
 // This file is part of Rcpp.
 //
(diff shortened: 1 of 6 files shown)


inst/include/Rcpp/sugar/matrix/lower_tri.h
```diff
@@ -1,8 +1,9 @@
-// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; tab-width: 8 -*-
+// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-
```
eddelbuettel: Thanks for catching that here.  Slowly changing them myself whenever I see one in a file I use.

inst/include/Rcpp/sugar/matrix/lower_tri.h
```diff
@@ -1,8 +1,9 @@
...
 //
 // lower_tri.h: Rcpp R/C++ interface class library -- lower.tri
 //
-// Copyright (C) 2010 - 2011 Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2010 - 2017 Dirk Eddelbuettel and Romain Francois
```
eddelbuettel: I think one usually separates the years, so that would be 2010 - 2016 if the next one 2017 - ...   No biggie.

Issue #640, opened 2017-01-30
Issue with Rcpp and Intel compiler exposed by building haven


wds15:
When installing haven 1.0.0 with the Intel compiler I endup getting this error in Rcpp files. Hadley suggested this is an Rcpp issue (I reported this for haven already):

Installing haven 1.0.0 with the Intel compiler (icpc (ICC) 15.0.3 20150407) I get this error log:

```
icpc -I/CHBS/apps/R/3.2.3/lib64/R/include        -I/usr/local/include -I"/CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include" -I"/CHBS/apps/dev_apps/R/dev_lib/3.2/BH/include"  -Ireadstat -fpic  -march=native -O3  -DEIGEN_USE_MKL_ALL -DEIGEN_NO_DEBUG -DNDEBUG -fp-model precise -fp-model source -mkl=sequential                                                           -c DfReader.cpp -o DfReader.o
In file included from /CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include/Rcpp.h(86),
                 from DfReader.cpp(1):
/CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include/Rcpp/algorithm.h(153): warning #858: type qualifier on return type is meaningless
          static inline RCPP_CONSTEXPR double ZERO() { return 0.0; }
                        ^

In file included from /CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include/Rcpp.h(86),
                 from DfReader.cpp(1):
/CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include/Rcpp/algorithm.h(154): warning #858: type qualifier on return type is meaningless
          static inline RCPP_CONSTEXPR double ONE() { return 1.0; }
                        ^

In file included from /CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include/Rcpp.h(86),
                 from DfReader.cpp(1):
/CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include/Rcpp/algorithm.h(162): warning #858: type qualifier on return type is meaningless
          static inline RCPP_CONSTEXPR int ZERO() { return 0; }
                        ^

In file included from /CHBS/apps/dev_apps/R/dev_lib/3.2/Rcpp/include/Rcpp.h(86),
                 from DfReader.cpp(1):
[...]

PR #643, opened 2017-01-31
Allow intel icc compilation of algorithm.h (closes 640)

eddelbuettel:
Simple PR adding what #640 ended up with in conclusion

Comments welcome, in particular double-checking whether the `else` side of the `#if` should be empty or `const`. Per #640 empty did not work. 

Poking @kevinushey @nathan-russell @dcdillon @jjallaire @thirdwing

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,11 +1,6 @@
-2017-02-01  Dirk Eddelbuettel  <edd@debian.org>
-
-        * inst/include/Rcpp/algorithm.h (RCPP_CONSTEXPR): Further refinement for
-        Intel compiler without C++11
-
 2017-01-31  Dirk Eddelbuettel  <edd@debian.org>
 
-        * DESCRIPTION (Date, Version): Roll minor version
+        * DESCRIPTION (Date, Version): Roll minor version 
 
         * inst/include/Rcpp/algorithm.h: Allow algorithm.h to be compiler under
         Intel's compiler, add copyright header and include guard
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,11 @@
+2017-02-01  Dirk Eddelbuettel  <edd@debian.org>
+
+        * inst/include/Rcpp/algorithm.h (RCPP_CONSTEXPR): Further refinement for
+        Intel compiler without C++11
+
 2017-01-31  Dirk Eddelbuettel  <edd@debian.org>
 
-        * DESCRIPTION (Date, Version): Roll minor version 
+        * DESCRIPTION (Date, Version): Roll minor version
 
         * inst/include/Rcpp/algorithm.h: Allow algorithm.h to be compiler under
         Intel's compiler, add copyright header and include guard
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2017-01-31  Dirk Eddelbuettel  <edd@debian.org>
+
+        * DESCRIPTION (Date, Version): Roll minor version 
+
+        * inst/include/Rcpp/algorithm.h: Allow algorithm.h to be compiler under
+        Intel's compiler, add copyright header and include guard
+
 2017-01-31  Nathan Russell  <russell.nr2012@gmail.com>
 
         * inst/include/Rcpp/sugar/matrix/upper_tri.h: Inherit from MatrixBase
--- a/ChangeLog
+++ b/ChangeLog
@@ -548,7 +548,7 @@
 
 2016-07-10  Nathan Russell <russell.nr2012@gmail.com>
 
-        * inst/include/Rcpp/algorithm.h: Accomdate clang compiler
+        * inst/include/Rcpp/algorithm.h: Accomodate clang compiler
 
 2016-07-03  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 0.12.9.1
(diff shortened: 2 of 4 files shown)


kevinushey:
Looks good to me, given the discussion in https://github.com/RcppCore/Rcpp/issues/640.

coatless:
Still providing similar warnings. Just ran this branch on ICC with Intel Compiler 15.0.3/16.0/17.0 on R 3.2.2. Output across versions was:

```bash
> devtools::install_github("rcppcore/rcpp", ref="bugfix/issue640", force=TRUE)
Downloading GitHub repo rcppcore/rcpp@bugfix/issue640
from URL https://api.github.com/repos/rcppcore/rcpp/zipball/bugfix/issue640
Installing Rcpp
'/usr/local/R/R-3.2.2/lib64/R/bin/R' --no-site-file --no-environ --no-save  \
  --no-restore CMD INSTALL  \
  '/tmp/RtmpWjmcXR/devtools9e5b4c9d99a5/RcppCore-Rcpp-98ecceb'  \
  --library='/home/balamut2/Rlibs' --install-tests

* installing *source* package 'Rcpp' ...
** libs
icc -I/usr/local/R/R-3.2.2/lib64/R/include -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -g -O2  -c Date.cpp -o Date.o
In file included from ../inst/include/Rcpp.h(86),
                 from Date.cpp(31):
../inst/include/Rcpp/algorithm.h(177): warning #858: type qualifier on return type is meaningless
          static inline RCPP_CONSTEXPR double ZERO() { return 0.0; }
                        ^
```

problematic_output

wds15:
Could you run the Intel tests again, but this time enable C++11 with the argument -std=C++11 ? I think you do not have 11 enabled in your tests.

coatless:
Aye, when I went to build against the Rcpp version I opted for the default settings (e.g. no `CXX_STD=CXX11` / `PKG_CXXFLAGS=-std=c++11`). 

Switching over to compiling with C++11 did lead to just one warning that was present before:

```
Date.cpp(642): warning #437: reference to local variable of enclosing function is not allowed
                      2 * sizeof *sp + 4 * TZ_MAX_TIMES];
```

round_two_output

wds15:
Looks much better now. Thanks for rerunning.

eddelbuettel:
That is sadly not good enough.  Rcpp itself is *not* built using `-std=c++11`.  

While we allow it it client packages (and eg haven may be using), we do not use it ourselves to only require a more basic compiler. g++-4.6.* and g++-4.4.* are still used on ancient RH systems.   Now we did not get the warning there yet.

Do we need a third branch in that `#define` to shut up icc when C++11 is not used?

wds15:
If you want to shut up Intel in non c++11 mode, then I am afraid that you need another fix for the else branch.

Optimally one can figure out a way which makes Intel and gcc happy; otherwise we need as you say another Intel quirk in there which I can try to find out what ifdef is needed to detect if Intel is used.

eddelbuettel:
Or we just merge as is, and the pleasure of using `icc` will include a bunch of warnings.  

Not the end of the world.

wds15:
Fine with me. Maybe I can look at it some time, now that I have a direct way to trigger the warning. After all its a warning, no error; so feel free to proceed.

eddelbuettel:
I made the change adding a simple test for the intel compiler.  Will commit, and a quick review should let this in so that we close this chapter.

eddelbuettel:
Re-upping poke to  @kevinushey @nathan-russell @dcdillon @jjallaire @thirdwing @coatless: All good?
I added two lines to define `RCPP_CONSTEXPR` as empty in an `#eflif` when do not have C++11.  The overall fallback of `constexpr` is behind it.

coatless:
So, forcing it to be blank under the Intel compiler actually triggers an error and kills the compilation.

Seems like to quiet this the value should not be defined in the declaration but in the definition per [C++ a member with an in-class initializer must be const](http://stackoverflow.com/questions/13274876/c-a-member-with-an-in-class-initializer-must-be-const).

```
> devtools::install_github("rcppcore/rcpp", ref="bugfix/issue640", force=TRUE)
Downloading GitHub repo rcppcore/rcpp@bugfix/issue640
from URL https://api.github.com/repos/rcppcore/rcpp/zipball/bugfix/issue640
Installing Rcpp
'/usr/local/R/R-3.2.2/lib64/R/bin/R' --no-site-file --no-environ --no-save  \
  --no-restore CMD INSTALL  \
  '/tmp/RtmpHtWzLF/devtools1c585c634b8c/RcppCore-Rcpp-f17ee6a'  \
  --library='/home/balamut2/Rlibs' --install-tests

* installing *source* package 'Rcpp' ...
** libs
icc -I/usr/local/R/R-3.2.2/lib64/R/include -DNDEBUG -I../inst/include/ -I/usr/local/include    -fpic  -g -O2  -c Date.cpp -o Date.o
In file included from ../inst/include/Rcpp.h(86),
                 from Date.cpp(31):
../inst/include/Rcpp/algorithm.h(177): error: a member with an in-class initializer must be const
          static RCPP_CONSTEXPR int RTYPE = REALSXP;
```

problematic_output

eddelbuettel:
Thanks so much for testing, @coatless.   The current state is clearly worse.  

I think your SO makes it clear that @kevinushey was (as usual) right: we need to differentiate between use of `const` and `constexpr` in returns and other cases.   @dcdillon, any chance you would like to tackle this?   

I think _for now_ I am going to revert my last commit, and the merge as it was.  It will build everywhere, but be noisy under icc when C++11 is not on (which folks could always turn locally I presume).

PR #648, opened 2017-02-13
new c++17 plugins

eddelbuettel:
The PR is short, it just adds plugins for c++17 and c++1z given that g++-6 is becoming widely available.

There is a bunch on white-space / line-ending change in here too.


eddelbuettel:
With KK's thumbs up I'll merge this now.

PR #652, opened 2017-03-01
minimal change for dynamic registration

eddelbuettel:


kevinushey:
LGTM!

Also, I think I finally figured out the semantics behind the use of native symbol registration and how `.Call()` works. Suppose you have an entry in your `R_CallMethods` struct like:

    {"C_routine", (DL_FUNC) &routine, 0}

This routine can be called by the registered name (as string), with e.g.

    .Call("C_routine", ..., PACAKGE = ...)

Or, with the generated 'native routine info' object, as:

    .Call(C_routine, ...)

However,

    .Call("routine")

will no longer work. In Rcpp's case, since the registration scheme here just uses the same name for the registered symbol as the C routine, no real change is needed in e.g. any of the existing `.Call()` invocations in Rcpp itself.

eddelbuettel:
Nice. Now, did we ever do `.Call("routine")` with the `PACKAGE=...` form?

[ I had meant to look at all this over the winter because "they" (BDR?) kept making these small changes and usec measurement statements but I never really dug in.  Tried briefly once or twice and got distracted each time... ]

eddelbuettel:
Coincidentally, I and a million other people just got a note from Kurt saying that we used

```
.Call("somefunc", ..., package="foo")
```

when it needs `PACKAGE="foo"`.  Oops.

PR #656, opened 2017-03-15
Add argument to XPtr template to run finalizer on exit (fixes #655)

### LINKED ISSUE #655: Running XPtr finalizer on exit
[I tried posting on the mailing list but having issues with my UCLA email...]

I am experiencing odd behavior in packages that use XPtr (V8, tesseract, xml2) which is caused by the fact that Rcpp does not run the finalizer when R quits. This can lead to corrupted connections, memory leaks, etc.

The problem is that Rcpp hardcodes the call to `R_RegisterCFinalizerEx` with [onexit = FALSE](https://github.com/RcppCore/Rcpp/blob/6f81b4684481dbd9bb554dd95e66725fc3b63a8c/inst/include/Rcpp/XPtr.h#L151) which makes that R skips over the finalizer when exiting.

Currently there is no way to fix this on the package side. Would there be a way to expose `onexit` as an option in Rcpp? Or better: make the default `TRUE` (imo you should always cleanup after yourself). But I can live with opt-in if there is a good reason for not finalizing.

jeroen:
See https://github.com/RcppCore/Rcpp/issues/655. This adds an additional parameter to `XPtr` which specifies if the finalizer should run on exit in the same spirit as the underlying `R_RegisterCFinalizerEx`.

I confirmed that this works. Default is still `FALSE` so the current behavior will be unchanged.

### DIFF
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -11,6 +11,8 @@
       DataFrame (James Balamuta in \ghpr{638} addressing \ghit{630}).
       \item Fixed single-character handling in \code{Rstreambuf} (Iñaki Ucar in
       \ghpr{649} addressing \ghit{647}).
+      \item XPtr gains a parameter \code{finalizeOnExit} to enable running the
+       finalizer when R quits (Jeroen Ooms in \ghpr{656} addressing \ghit{655}).
     }
     \item Changes in Rcpp Sugar:
     \itemize{
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2017-03-15  Jeroen Ooms  <jeroenooms@gmail.com>
+
+        * inst/include/Rcpp/XPtr.h: added finalizeOnExit parameter
+
 2017-02-28  Dirk Eddelbuettel  <edd@debian.org>
 
         * src/Rcpp_init.cpp (R_init_Rcpp): Call R_useDynamicSymbols()
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 0.12.9.4
-Date: 2017-02-28
+Version: 0.12.9.5
+Date: 2017-03-15
 Author: Dirk Eddelbuettel, Romain Francois, JJ Allaire, Kevin Ushey, Qiang Kou,
  Nathan Russell, Douglas Bates and John Chambers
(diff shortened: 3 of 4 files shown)


jeroen:
OK perfect, I'll add a NEWS item and bump the version.

eddelbuettel:
Great, and a changelog entry, please.

Are the data structures in your client packages C or C++?  I am just wondering if you could also get this for free for C++ destructor semantics on your side .  Just checking ...

inst/NEWS.Rd
```diff
@@ -7,6 +7,8 @@
   \itemize{
     \item Changes in Rcpp API:
     \itemize{
+      \item XPtr gains a parameter \code{finalizeoOnExit} to enable running the
```
jjallaire: Spelling of parameter here (finalizeoOnExit should be finalizeOnExit)

eddelbuettel: May also move the entry to the bottom.  They are mostly chronological.  In any even I can clean up both when finalizing the release "real soon now".

jjallaire:
@eddelbuettel I think the default finalizer will delete the C++ object thereby calling the destructor (so without this change there is actually no way to guarantee that the destructor is called prior to the process exiting)

@jeroen FWIW in some cases I've found it better to not let C++ destructors get called at process exit because the runtime context of an exiting process can be so stripped down (or non-deterministic across multiple C++ objects getting destroyed) that the destructor either crashes or hangs. i.e. sometimes it's better to just let the OS clean up than to try to explicitly do so within the process. Obviously this doesn't work in all cases, but I know that many things like mutexes, sockets, file handles, memory, etc. are freed automatically by the OS at process exit.

eddelbuettel:
Leaving the default at `false` is definitely the way to go.

jeroen:
OK fixed the typo and moved the `\item{}` to the bottom.

eddelbuettel:
I launched a full reverse-depends check just in case, but we won't know til tomorrow how it does.

eddelbuettel:
Rev deps were all fine, results in https://github.com/RcppCore/rcpp-logs/commit/8103eff9e9068b6b90ccc16cba2b485e3b4c4494

jeroen:
>Rev deps were all fine, results in RcppCore/rcpp-logs@8103eff

Let's send it to CRAN then :D

eddelbuettel:
Yes, which is why I ran the two previous reverse-depends.  Probably by Friday.

Given the long tail of these reverse dependencies, Rcpp is a littler slower to make it from Incoming but you should be able to rely on it come next week.

PR #663, opened 2017-04-03
Added support for throwing exceptions without call stacks.

jimhester:
There is currently no way to generate an exception from Rcpp which does _not_ have an attached R (and cpp) call stacks. This extra information can be confusing for users, so it is useful to be able to disable it if desired (like `stop(call. = FALSE)` does in R code.

This could be implemented by providing overloaded `forward_exception_to_r()`and `exception_to_r_condition()` functions for `Rcpp::exceptions`, but I felt like it made the intention more clear to define new functions instead.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2017-04-03  Jim Hester  <james.f.hester@gmail.com>
+
+        * inst/include/Rcpp/exceptions.h: Added support for throwing
+        exceptions without call stacks.
+        * inst/include/Rcpp/macros/macros.h: Idem
+        * inst/unitTests/cpp/exceptions.cpp: Idem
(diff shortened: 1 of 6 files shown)


eddelbuettel:
Did you see [Contributing.md](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md)  ?

jimhester:
Yes, but not until this was already implemented, since it was already done might as well look at the code

eddelbuettel:
Meant to reopen this earlier after we all seemed to have reached consensus on #664, but got sidetracked.

eddelbuettel:
This looks fine to me.  Any other takers?

kevinushey:
+1

PR #667, opened 2017-04-10
[WIP] Feature improve exception message details (closes #184)

### LINKED ISSUE #184: better error messages
There are a number of places where Rcpp can fail because `TYPEOF(x)` does not equal `RTYPE`.  Since both of these are known at the point that the error is given it would help if these were displayed, e.g.  rather than just `"some message"` use:  `std:string("some message TYPEOF(x): ") + 
                CHAR(Rf_type2str(TYPEOF(x))) + 
                " RTYPE: " + 
                CHAR(Rf_type2str(RTYPE))`
At least that will give some bit of clue as to what the problem might be.

Any other info that might be added to the error message to help locate the problem could be helpful too.

This applies to RcppEigen too.

coatless:
### Overview

This is a work in progress (WIP) PR to upgrade the exception messages found in Rcpp. 

For details on the contents of the PR, please see #184. (The issue ticket contains GitHub search URL for each Rcpp exception message.)

### Changes:

- Added `RCPP_ADVANCED_EXCEPTION_CLASS` macro that mimics the structure of the `Rcpp::stop()` / `Rcpp::warning()` implementations.
- Removed `no_such_field` 
- Changed throw type from `not_compatible` to `not_a_matrix` in `inst\include\Rcpp\vector\Matrix.h` (FYI, a `throw` was missing on one of the statements).

Presently, I've held off on removing  `reference_creation_error` as it is not used but it is referenced within the documentation.

### Misc

This branch will likely need to be rebased after @jimhester's #663 PR.

### DIFF
--- a/inst/include/Rcpp/api/meat/DottedPairImpl.h
+++ b/inst/include/Rcpp/api/meat/DottedPairImpl.h
@@ -53,7 +53,7 @@ namespace Rcpp{
 		} else {
 			if( ref.isNULL( ) ) throw index_out_of_bounds() ;
 
-                        if( static_cast<R_xlen_t>(index) > ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds() ;
+			if( static_cast<R_xlen_t>(index) > ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds("index is out of bounds. Requested %i with %i available", static_cast<R_xlen_t>(index), ::Rf_xlength(ref.get__()) ) ;
 
 			size_t i=1;
 			SEXP x = ref.get__() ;
@@ -70,7 +70,7 @@ namespace Rcpp{
     template <typename T>
 	void DottedPairImpl<CLASS>::replace( const int& index, const T& object ) {
 	    CLASS& ref = static_cast<CLASS&>(*this) ;
-        if( static_cast<R_xlen_t>(index) >= ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds() ;
+        if( static_cast<R_xlen_t>(index) >= ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds("index is out of bounds. Requested %i with %i available", static_cast<R_xlen_t>(index), ::Rf_xlength(ref.get__()) ) ;
 
         Shield<SEXP> x( pairlist( object ) );
         SEXP y = ref.get__() ;
@@ -84,7 +84,7 @@ namespace Rcpp{
 	template <typename CLASS>
     void DottedPairImpl<CLASS>::remove( const size_t& index ) {
         CLASS& ref = static_cast<CLASS&>(*this) ;
-        if( static_cast<R_xlen_t>(index) >= Rf_xlength(ref.get__()) ) throw index_out_of_bounds() ;
+        if( static_cast<R_xlen_t>(index) >= Rf_xlength(ref.get__()) ) throw index_out_of_bounds("index is out of bounds. Requested %i with %i available", static_cast<R_xlen_t>(index), ::Rf_xlength(ref.get__()) ) ;
         if( index == 0 ){
             ref.set__( CDR( ref.get__() ) ) ;
         } else{
--- a/inst/include/Rcpp/api/meat/DottedPairImpl.h
+++ b/inst/include/Rcpp/api/meat/DottedPairImpl.h
@@ -53,7 +53,12 @@ namespace Rcpp{
 		} else {
 			if( ref.isNULL( ) ) throw index_out_of_bounds() ;
 
-			if( static_cast<R_xlen_t>(index) > ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds("index is out of bounds. Requested %i with %i available", static_cast<R_xlen_t>(index), ::Rf_xlength(ref.get__()) ) ;
+			if( static_cast<R_xlen_t>(index) > ::Rf_xlength(ref.get__()) ) {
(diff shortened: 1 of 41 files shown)


inst/include/Rcpp/api/meat/DottedPairImpl.h
```diff
@@ -53,7 +53,7 @@ namespace Rcpp{
 		} else {
 			if( ref.isNULL( ) ) throw index_out_of_bounds() ;
 
-                        if( static_cast<R_xlen_t>(index) > ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds() ;
+			if( static_cast<R_xlen_t>(index) > ::Rf_xlength(ref.get__()) ) throw index_out_of_bounds("index is out of bounds. Requested %i with %i available", static_cast<R_xlen_t>(index), ::Rf_xlength(ref.get__()) ) ;
```
kevinushey: Minor, but to avoid overly long lines like this I often use an idiom like (pseudo-ish code):

```
if (condition) {
    const char* fmt =
        "Index out of bounds: [index=%s; extent=%s]"
    throw index_out_of_bounds(fmt, index, length);
}
```

eddelbuettel: And once can wrap the `"...."` over two lines as `"abc" \n "def"`.

Luckily we have no strict 80 column policy here :)

inst/include/Rcpp/exceptions.h
```diff
@@ -89,6 +89,48 @@ namespace Rcpp {
             file_io_error("file already exists", file) {}	// #nocov end
     };
 
+    #define RCPP_ADVANCED_EXCEPTION_CLASS(__CLASS__, __WHAT__)                                                                                                                                                     \
```
kevinushey: Minor suggestion, but it'd be nice if we had a C++11 version (using variadic templates) and a C++98 version (using your current code-bloat version). We could do something like:

```
#if __cplusplus >= 201103L
# include <Rcpp/exceptions/cpp11/exceptions.h>
#else
# include <Rcpp/exceptions/cpp98/exceptions.h>
#endif
```

Let me know if you think it's worth the effort.

coatless: It would be very nice to have Vardiac template support... Mmm. 

Would something along the lines of:

```
#define RCPP_CPP11_EXCEPTION_CLASS(__CLASS__, __WHAT__)     
     class __CLASS__ : public std::exception{                                       \
     public:                                                                        \
         __CLASS__( const std::string& message ) throw() : message( __WHAT__ ){} ;  \
        template <typename... Args>                                                 \
         __CLASS__( const char* fmt, Args&&... args ) throw() :                     \
         message(  tfm::format(fmt, std::forward<Args>(args)...) ){} ;              \
         virtual ~__CLASS__() throw(){} ;                                           \
         virtual const char* what() const throw() { return message.c_str() ; }      \
     private:                                                                       \
         std::string message ;                                                      \
     } ;
```

Be acceptable? 

**Note:** The defines in [`/inst/include/Rcpp/utils/tinyformat.h`](https://github.com/RcppCore/Rcpp/blob/6f81b4684481dbd9bb554dd95e66725fc3b63a8c/inst/include/Rcpp/utils/tinyformat.h#L131-L136) will also need to receive the `if __cplusplus >= 201103L` check.

inst/include/Rcpp/api/meat/DottedPairImpl.h
coatless: Aye, that makes it a lot more manageable. I'll switch everything over.

inst/include/Rcpp/exceptions.h
kevinushey: Yes, something exactly like that is what I was thinking of :)

Indeed, I think we somewhat bluntly said "make tinyformat work in C++98 mode" a while back without consideration for C++11 mode, so it would make sense to guard the define there according to whether we're using C++98 or C++11.

eddelbuettel: We have some years-old tests for macros for C++11 which can probably do with brush-up.  Now that we will have a decent compiler "most of the time" we may as well use it.

But we can of course not loose sight of the poor souls clinging to their old RHEL/CentOS/... systems from the last ice age.  It's is probably worth our while to ensure those build too.

coatless: Enabling C++11 mode for `tinyformat` triggered a lot of error spam that is stuttering shell. So, I've left out this change until I can investigate the file a bit more after I grab some sleep.

inst/include/Rcpp/api/meat/DottedPairImpl.h
coatless: Pushed through a commit that addressed this feedback.

inst/include/Rcpp/exceptions.h
eddelbuettel: I think we may have tried that too in the past and retreated.  Foggy memory on this though...

coatless: I'm going to push an updated version of the `tinyformat.h`. It last received love in Aug 29, 2015. I need a fix released in [March 2016](https://github.com/c42f/tinyformat/commit/5abc03b35457f1af1b0c7257129017ff5a82e504) to quiet a `[-Wuninitialized]` complaint. After that, everything else should work.

eddelbuettel: Sure.

But let's open an issue ticket, preferably with a link to the tinyformat repo and changelog, and take this out of this somewhat growing discussion.

PR #668, opened 2017-04-11
Feature/update modules example

eddelbuettel:
This PR contains updates to the two 'sub-packages' used for testing and demo:
- `inst/unit/testRcppModule/`
- `inst/unit/testRcppClass/`

The first (for Modules) is in good shape and updates the Modules example to current R 3.4.0 standards.

The second (for John's Rcpp Classes) is in less ideal shape, and still emits warning.  John suspects these may come from R itself so this can still be committed.

[ I tried to squash-commit Jim's PR in so that this would be mergeable, seems like that failed.  My master now has these two commit's cherried in, but I won't push that til I see some nodding or thumbs-up here. ]

[ Edit: Was able to sort it out here.  It only affected the ChangeLog. ]

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,32 @@
+2017-04-09  Dirk Eddelbuettel  <edd@debian.org>
+
+        * inst/unitTests/testRcppModule/src/init.c (R_init_testRcppModule): Call
+        R_registerRoutines() and R_useDynamicSymbols(); also ensure
+        _rcpp_module_boot_* is registered for each module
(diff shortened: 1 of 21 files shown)


eddelbuettel:
@jjallaire @kevinushey @thirdwing @nathan-russell :  I take lack of screaming as silent consent and will fold this in tomorrow or the next, unless someone beats me to it.  It only affects demo / test packages illustrating Modules (and Rcpp Classes) so it is unlikely to do any damage.

kevinushey:
LGTM!

PR #670, opened 2017-04-13
Explicitly call parent constructor

krlmlr:
for copy constructor, to avoid warning in `-Wextra`.


kevinushey:
LGTM

eddelbuettel:
I turned a rev.dep on for a combination of #670 and #671 but it will run into the night.

eddelbuettel:
And given the very recent change to #671 I just updated to that and restarted.

eddelbuettel:
Reverse-depends check [worked just fine](https://github.com/RcppCore/rcpp-logs/commit/ea021dda73bd61681ef1b8ea485d1b767d13619f), so merging this.

PR #671, opened 2017-04-13
BEGIN_RCPP uses the variables it defines

krlmlr:
Will be used instead of `BEGIN_RCPP` for generated wrappers for C++ interfaces (`[[Rcpp::interfaces(cpp)]]`).

### DIFF
--- a/inst/include/Rcpp/macros/macros.h
+++ b/inst/include/Rcpp/macros/macros.h
@@ -65,6 +65,11 @@
 #define END_RCPP VOID_END_RCPP return R_NilValue;
 #endif
 
+#ifndef BEGIN_RCPP_RETURN_ERROR
+#define BEGIN_RCPP_RETURN_ERROR                                                \
+    try {
+#endif
+
 #ifndef END_RCPP_RETURN_ERROR
 #define END_RCPP_RETURN_ERROR                                                  \
   }                                                                            \
--- a/inst/include/Rcpp/macros/macros.h
+++ b/inst/include/Rcpp/macros/macros.h
@@ -29,7 +29,9 @@
 #ifndef BEGIN_RCPP
 #define BEGIN_RCPP                                                                               \
     int rcpp_output_type = 0 ;                                                                   \
+    (void)rcpp_output_type;                                                                      \
     SEXP rcpp_output_condition = R_NilValue ;                                                    \
+    (void)rcpp_output_condition;                                                                 \
     try {
 #endif
 
@@ -63,11 +65,6 @@
 
 #ifndef END_RCPP
 #define END_RCPP VOID_END_RCPP return R_NilValue;
-#endif
-
-#ifndef BEGIN_RCPP_RETURN_ERROR
-#define BEGIN_RCPP_RETURN_ERROR                                                \
-    try {
 #endif
 
 #ifndef END_RCPP_RETURN_ERROR
--- a/src/attributes.cpp
+++ b/src/attributes.cpp
@@ -2581,7 +2581,8 @@ namespace attributes {
             }
             std::string args = ostrArgs.str();
             ostr << args << ") {" << std::endl;
-            ostr << "BEGIN_RCPP" << std::endl;
+            ostr << (cppInterface ? "BEGIN_RCPP_RETURN_ERROR" : "BEGIN_RCPP")
+                 << std::endl;
             if (!function.type().isVoid())
                 ostr << "    Rcpp::RObject rcpp_result_gen;" << std::endl;
             if (!cppInterface && attribute.rng())
--- a/src/attributes.cpp
+++ b/src/attributes.cpp
@@ -2581,8 +2581,7 @@ namespace attributes {
             }
             std::string args = ostrArgs.str();
             ostr << args << ") {" << std::endl;
-            ostr << (cppInterface ? "BEGIN_RCPP_RETURN_ERROR" : "BEGIN_RCPP")
-                 << std::endl;
+            ostr << "BEGIN_RCPP" << std::endl;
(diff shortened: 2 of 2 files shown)


eddelbuettel:
This one looks, quite frankly, suspicious.  Or maybe I need my 2nd cup of coffee for the morning.

Under a certain condition you add an opening brace as well as `try`.  I see no matching closing brace.  Hm.

krlmlr:
This is just a stripped-down version of `BEGIN_RCPP`: https://github.com/RcppCore/Rcpp/pull/671/files#diff-9815ed15f5a34bd50e396336af995d42R30

eddelbuettel:
Ack.

The name sucks.  Maybe `BEGIN_RCPP_NO_VARS`  ?

And what did your mother tell you about _opening issues tickets before PRs_?   I know you don;t believe me yet that  "Suggests ! Depend" (but you will ;-) ) but also PR != Issue.  ;-)

eddelbuettel:
And of course, this one is basically untestable by reverse-depends `R CMD check` as that tickles compilations and checks almost certainly almost none of which will involve a call to `compileAttributes()`. 

But it seems fine on the merits.

krlmlr:
This is a companion to `END_RCPP_RETURN_ERROR`, both are relatively close to each other in the generated code. But I can change it to use any name you prefer.

Testing it involves using it in a package; maybe a bunch of acceptance/integration tests in Rcpp that just compile an Rcpp package will be helpful? Gábor has created a few utility packages that will make this a pleasant exercise.

A patch says more than a thousand words (and is faster to write, too).

kevinushey:
Looks good to me!

jjallaire:
LGTM as well.

eddelbuettel:
Yes, this one is simple enough. I clearly misread when I just got up.  

I think I'd still like to rename it.

jjallaire:
I think it's most clear with the current name as it will provide a
symmetrical macro name, so we'd have

BEGIN_RCPP

END_CPP

or

BEGIN_RCPP_NO_RETURN

END_RCPP_NO_RETURN

To me this seems pretty clear.



On Thu, Apr 13, 2017 at 3:35 PM, Dirk Eddelbuettel <notifications@github.com

jjallaire:
Sorry,

BEGIN_RCPP_RETURN_ERROR

END_RCPP_RETURN_ERROR

On Thu, Apr 13, 2017 at 3:35 PM, Dirk Eddelbuettel <notifications@github.com

eddelbuettel:
But it does nothing w.r.t. to error handlers or triggers.  Which is why I find putting 'ERROR' in there a bit weird.

Otherwise of course on board re the symmetry.

jjallaire:
The second macro does though (it returns an error rather than throwing one to workaround the win32 can't throw exceptions across modules when static linking to gcc runtime issue). So "RETURN_ERROR" really just means "this code will return an error rather than throwing one".

I know you are already aware of all that, just trying to elaborate on why I think the macro name is acceptable as an enclosing "concept" rather than literally what the individual macro does.

krlmlr:
One thing I haven't considered: packages that are using the new macro won't work with an older Rcpp version. What's the best way to ensure this? Should we patch `DESCRIPTION`?

eddelbuettel:
No, we will not overwrite every user's DESCRIPTION file.  We're not roxygen2 or devtools or other rude packages.

The package is one of a large number of entirely _optional_ features and when maintainers Alice or Bob use, they are expected to edit their DESCRIPTION files.

jjallaire:
Creating a situation where a new version of Rcpp is required just so we can
avoid an unused variable warning seems like a bad trade to me.

Could we just add a pragma to get around the warning or just live with it?

krlmlr:
The last commit might work better (doesn't require changes in generated code), but I need to verify that it's actually helpful.

krlmlr:
Confirmed that the proposed solution actually clears the "unused variable" warnings.

eddelbuettel:
Reverse-depends check [worked just fine](https://github.com/RcppCore/rcpp-logs/commit/ea021dda73bd61681ef1b8ea485d1b767d13619f), so merging this.

Proposed change is also nice and shorter. Good.

krlmlr:
Awesome, thanks! Are you planning to release mid-May?

eddelbuettel:
Yes, I see no reason to depart from the bi-monthly schedule (apart from having a very busy May myself, but we'll cope).  Any plans/concerns regarding the timeline?

I did of course skip one your packages in testing, but we've been down that road before ;-)

Also just added ChangeLog and NEWS.  Alter as needed in next PR and keep in mind that [Contributing.md](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) suggests having them in the PR in the first place.

eddelbuettel:
Doh. I guess it would help for dplyr 0.6.0?

krlmlr:
Don't worry, locally I have `-Wextra`, and I can use Rcpp from GitHub too. I was asking to assess if it's worthwhile to patch .travis.yml to enable `-Wextra` on Travis, but we can wait one month here.

krlmlr:
Thanks for editing the ChangeLog, I wasn't sure to which PR to add the change, and I sure didn't want to add three changes in three PRs... Is there a nice established way to generate change logs from GitHub commit messages?

eddelbuettel:
The single best support is in Emacs -- hitting C-x 4 a in any source file adds a stanza for that file.  

All git2changelog converters that I have seen are deficient in the sense that you generally want to condense the entry a little (and then once more for NEWS).  So I think of this more as a human-edited file.  The machine readable log is indeed in git.

I just keep complaining and adding them by hand where missing.  Worked for 10+ years. ;-)

PR #672, opened 2017-04-13
Remove name for unused arguments

krlmlr:
~~with `(void)arg;` in the function's body.~~

### DIFF
--- a/inst/include/Rcpp/sugar/functions/rep.h
+++ b/inst/include/Rcpp/sugar/functions/rep.h
@@ -54,6 +54,7 @@ public:
         Rep_Single( const T& x_, R_xlen_t n_) : x(x_), n(n_){}
 
         inline T operator[]( R_xlen_t i ) const {
+		(void)i;
 		return x;
 	}
         inline R_xlen_t size() const { return n ; }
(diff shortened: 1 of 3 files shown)


kevinushey:
LGTM, although I wonder about a couple of the non-used arguments (it seems like they should likely be used?)

kevinushey:
Oh, it looks like a bunch of classes within `attributes.cpp` have the same interface but some don't need to handle the accepted arguments.

jjallaire:
Those can just be changed to have only the type in the signature rather
than the type and variable name then they should pass.

eddelbuettel:
Woot. And just installed with `ccache g++ -I/usr/share/R/include -DNDEBUG -I../inst/include/ -fpic  -g -O3 -Wall -pipe -pedantic -Wextra -c`.  That's first. 

Thanks for that.  I am sure I will have to remove `-Wextra` from `Makevars` when compiling other packages but it is rather nice to have Rcpp squeaky clean.

krlmlr:
You're welcome. I tend to use a package-local `Makevars` (with a corresponding .Rprofile): https://github.com/krlmlr/scriptlets/commit/846793438bd842057e46831a89a948e314457f04.

eddelbuettel:
Not familiar with those hooks. Does it require devtools, and would standard R tools adhere to them too?

krlmlr:
Works with base R, also in R CMD INSTALL. `R_MAKEVARS_USER` is described in [R-admin](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Customizing-package-compilation).

eddelbuettel:
Nice! I'll follow up on that.  I tend to be aware of what is in R-exts, but they trick me now by hiding more stuff in R-inst and R-admin.

PR #680, opened 2017-04-22
Sugar function 'trimws' with unit tests (closes #679)

### LINKED ISSUE #679: Sugar function trimws
I put together a sugar function `trimws` to mirror `base::trimws` (added in R 3.2.0). For those not familiar with this, it removes either leading (`which = "left"`), trailing (`which = "right"`), or leading and trailing (`which = "both"`) whitespace (defined as any of `' '`, `'\t'`, `'\r'`, or `'\n'`) from strings, returning `NA` on `NA` input. 

In my fork, usage looks like this: 

```c++
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
CharacterVector trim(CharacterVector x, const char* which = "both")
{
    return trimws(x, which);
}

/*** R

trim("  x y z \t \n \r   ")
# [1] "x y z"

trim("  x y z \t \n \r   ", "left")
# [1] "x y z \t \n \r   "

trim("  x y z \t \n \r   ", "right")
# [1] "  x y z"

*/
```

#### Other features: 
* Default `which` argument is `"both"`, as in `base::trimws` 
* Only the first letter of `which` is checked (`'b'`, `'l'`, and `'r'` are valid first letters) to mimic the use of `match.arg` in R 
* It is also overloaded for the `CharacterMatrix` and `String` classes
* Much faster than the `base` version, which uses regex to accomplish this

-------

If this seems like a worthwhile addition, I can submit a PR shortly.

nathan-russell:
As described in #679, this adds a sugar version of the base R function `trimws`.

### DIFF
--- a/inst/include/Rcpp/sugar/functions/strings/trimws.h
+++ b/inst/include/Rcpp/sugar/functions/strings/trimws.h
@@ -95,58 +95,72 @@ inline const char* trim_both(const char* str) {
 
 
 inline Vector<STRSXP> trimws(const Vector<STRSXP>& x, const char* which = "both") {
-    typedef const char* (*trim_function)(const char*);
-    trim_function trim = NULL;
+    R_xlen_t i = 0, sz = x.size();
+    Vector<STRSXP> res = no_init(sz);
 
     if (*which == 'b') {
-        trim = sugar::detail::trim_both;
+        for (; i < sz; i++) {
+            if (traits::is_na<STRSXP>(x[i])) {
+                res[i] = x[i];
+            } else {
+                res[i] = sugar::detail::trim_both(x[i]);
+            }
+        }
     } else if (*which == 'l') {
-        trim = sugar::detail::trim_left;
+        for (; i < sz; i++) {
+            if (traits::is_na<STRSXP>(x[i])) {
+                res[i] = x[i];
+            } else {
+                res[i] = sugar::detail::trim_left(x[i]);
+            }
+        }
     } else if (*which == 'r') {
-        trim = sugar::detail::trim_right;
+        for (; i < sz; i++) {
+            if (traits::is_na<STRSXP>(x[i])) {
+                res[i] = x[i];
+            } else {
+                res[i] = sugar::detail::trim_right(x[i]);
+            }
+        }
     } else {
         stop("Invalid `which` argument '%s'!", which);
         return Vector<STRSXP>::create("Unreachable");
-    }
-
-    R_xlen_t i = 0, sz = x.size();
-    Vector<STRSXP> res(sz);
-
-    for (; i < sz; i++) {
-        if (traits::is_na<STRSXP>(x[i])) {
-            res[i] = x[i];
-        } else {
-            res[i] = (*trim)(x[i]);
-        }
     }
 
     return res;
 }
 
 inline Matrix<STRSXP> trimws(const Matrix<STRSXP>& x, const char* which = "both") {
-    typedef const char* (*trim_function)(const char*);
-    trim_function trim = NULL;
+    R_xlen_t i = 0, nr = x.nrow(), nc = x.ncol(), sz = x.size();
+    Matrix<STRSXP> res = no_init(nr, nc);
 
     if (*which == 'b') {
-        trim = sugar::detail::trim_both;
+        for (; i < sz; i++) {
+            if (traits::is_na<STRSXP>(x[i])) {
+                res[i] = x[i];
+            } else {
+                res[i] = sugar::detail::trim_both(x[i]);
+            }
+        }
     } else if (*which == 'l') {
-        trim = sugar::detail::trim_left;
+        for (; i < sz; i++) {
+            if (traits::is_na<STRSXP>(x[i])) {
+                res[i] = x[i];
+            } else {
+                res[i] = sugar::detail::trim_left(x[i]);
+            }
+        }
     } else if (*which == 'r') {
-        trim = sugar::detail::trim_right;
+        for (; i < sz; i++) {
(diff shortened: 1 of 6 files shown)


inst/include/Rcpp/sugar/functions/strings/trimws.h
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    while (isws(*str)) {
+        ++str;
+    }
+
+    return str;
+}
+
+inline const char* trim_right(const char* str) {
```
kevinushey: It seems like these functions should also accept the string length as well -- this would help avoid a call to `::strlen()` when the length is already known, and it should already be known for R's `CHARSXP`s.

inst/include/Rcpp/sugar/functions/strings/trimws.h
```diff
@@ -0,0 +1,184 @@
...
+    static std::string buff;
```
kevinushey: IIUC, the use of a static buffer here implies that we would effectively hold on to a (potentially large) string in memory each time this function is called (and that memory is never released). Is there any chance we could avoid this?

In theory, it implies someone calling this with a very large string would end up 'leaking' memory (or, I guess more aptly, letting a large bit of memory remain allocated until R exits)

inst/include/Rcpp/sugar/functions/strings/trimws.h
```diff
@@ -0,0 +1,184 @@
...
+        trim = sugar::detail::trim_left;
+    } else if (*which == 'r') {
+        trim = sugar::detail::trim_right;
+    } else {
+        stop("Invalid `which` argument '%s'!", which);
+        return Vector<STRSXP>::create("Unreachable");
+    }
+
+    R_xlen_t i = 0, sz = x.size();
+    Vector<STRSXP> res(sz);
```
kevinushey: You might consider using `Vector<STRSXP> res = no_init(sz);` to avoid an unneeded initial allocation.

inst/include/Rcpp/sugar/functions/strings/trimws.h
```diff
@@ -0,0 +1,184 @@
...
+
+    for (; i < sz; i++) {
+        if (traits::is_na<STRSXP>(x[i])) {
+            res[i] = x[i];
+        } else {
+            res[i] = (*trim)(x[i]);
```
kevinushey: It'd be nice if we could avoid calling through a function pointer here (just to avoid the extra indirection), although not strictly necessary.

kevinushey:
A couple mostly minor comments but LGTM!

inst/include/Rcpp/sugar/functions/strings/trimws.h
nathan-russell: I wasn't thrilled about using `strlen`, but I assumed this was unavoidable because that is how the length is determined [in the `string_proxy` class](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/vector/string_proxy.h#L160). I'm not terribly familiar with R's internal string cache, but if the lengths of `CHARSXP`s are known as you suggest, I'd be more than happy to find a way to incorporate that in place of `strlen`. Can you get back to me on this?

inst/include/Rcpp/sugar/functions/strings/trimws.h
nathan-russell: You make a good point here. I was primarily looking to avoid repeatedly allocating and deallocating memory for a new `std::string` on each function call, which I assumed would happen if had used a buffer with automatic storage duration. However, I think I can just create a (non-`static`) variable in the enclosing function (`trimws`) and pass it by reference to `trim_right` and `trim_both` so that the memory gets released when `trimws` returns.

inst/include/Rcpp/sugar/functions/strings/trimws.h
nathan-russell: Good catch -- I'll fix this too.

inst/include/Rcpp/sugar/functions/strings/trimws.h
nathan-russell: That's fair. I had done this just to keep the body more concise, but in hindsight I think a few extra lines of code are preferable to the extra cost of dereferencing (assuming these aren't inlined).

inst/include/Rcpp/sugar/functions/strings/trimws.h
kevinushey: I believe you can simply call `Rf_length()` on a `CHARSXP` to retrieve its length, as the length will be stored in the `SEXP` header.

inst/include/Rcpp/sugar/functions/strings/trimws.h
kevinushey: One possibility -- this function could instead return a pair of pointers (the new 'start' and 'end' of the string), and the calling function can decide how to use those pointers (e.g. duplicate into some new memory, or just use as-is if were appropriate).

inst/include/Rcpp/sugar/functions/strings/trimws.h
kevinushey: Ah, it looks like you've effectively done that now (with the call to `LENGTH()`). Great!

inst/include/Rcpp/sugar/functions/strings/trimws.h
kevinushey: The use of a `std::string*` passed as a pointer in the latest commits looks fine to me as well.

kevinushey:
LGTM as well!

PR #686, opened 2017-04-26
suggested fix for maacheck (closes #685)

### LINKED ISSUE #685: fix potential protect errors
From https://github.com/kalibera/rprotect:

> maacheck
> 
> This tool is quite simple and only reports instances of multiple allocating arguments pattern, such as (from Rcpp code):
> 
> Rf_lang2(::Rf_install("simpleError"), Rf_mkString(str.c_str()))
> 
> This is a real error: Rf_mkString returns an unprotected newly allocated object. Rf_install may allocate and may be called after Rf_mkString, and hence may trigger garbage collection that will corrupt the object allocated by Rf_mkString.

The associated fix needs to occur here: https://github.com/RcppCore/Rcpp/blob/e81493bd988a5ee93e5a35458662dd2582de1c5d/inst/include/Rcpp/exceptions.h#L294

eddelbuettel:


eddelbuettel:
@kevinushey @nathan-russell @thirdwing @jjallaire -- any comments? 

It's the last `Rf_mkString` and `Rf_install` interaction remaining.

eddelbuettel:
Re-pinging @kevinushey @nathan-russell @thirdwing @jjallaire just in case but will commit unless anybody screams.  This should be harmless.

eddelbuettel:
Thank you both!

PR #687, opened 2017-05-05
Reduce -Wconversion warnings for date-time code

krlmlr:
As discussed on [Rcpp-devel](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2017-April/009574.html).

### DIFF
--- a/inst/include/Rcpp/date_datetime/Date.h
+++ b/inst/include/Rcpp/date_datetime/Date.h
@@ -127,7 +127,7 @@ namespace Rcpp {
         // update m_tm based on m_d
         void update_tm() {
             if (R_FINITE(m_d)) {
-                time_t t = (time_t)(24*60*60 * m_d);      // (fractional) days since epoch to seconds since epoch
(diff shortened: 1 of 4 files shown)


inst/include/Rcpp/date_datetime/Date.h
```diff
@@ -127,7 +127,7 @@ namespace Rcpp {
         // update m_tm based on m_d
         void update_tm() {
             if (R_FINITE(m_d)) {
-                time_t t = 24*60*60 * m_d;      // (fractional) days since epoch to seconds since epoch
+                time_t t = (time_t)(24*60*60 * m_d);      // (fractional) days since epoch to seconds since epoch
```
coatless: I think you may need to use a `static_cast<time_t>()` in place of the c-style / regular cast.

eddelbuettel:
This looks good to me. It would help if we had more tests for some of the rounding etc issues, but blame for that is mostly on me.

PR #688, opened 2017-05-05
Reduce -Wconversion warnings for sugar code

krlmlr:
2/n.

### DIFF
--- a/inst/include/Rcpp/sugar/functions/strings/trimws.h
+++ b/inst/include/Rcpp/sugar/functions/strings/trimws.h
@@ -133,7 +133,8 @@ inline Vector<STRSXP> trimws(const Vector<STRSXP>& x, const char* which = "both"
 }
 
 inline Matrix<STRSXP> trimws(const Matrix<STRSXP>& x, const char* which = "both") {
-    R_xlen_t i = 0, nr = x.nrow(), nc = x.ncol(), sz = x.size();
+    R_xlen_t i = 0, sz = x.size();
+    int nr = x.nrow(), nc = x.ncol();
     Matrix<STRSXP> res = no_init(nr, nc);
     std::string buffer;
 
--- a/inst/include/Rcpp/sugar/functions/rowSums.h
+++ b/inst/include/Rcpp/sugar/functions/rowSums.h
@@ -63,12 +63,12 @@ inline void incr(Rcomplex* lhs, const Rcomplex& rhs) {
 
 
 inline void div(double* lhs, R_xlen_t rhs) {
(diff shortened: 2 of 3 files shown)


inst/include/Rcpp/sugar/functions/strings/trimws.h
```diff
@@ -133,7 +133,8 @@ inline Vector<STRSXP> trimws(const Vector<STRSXP>& x, const char* which = "both"
 }
 
 inline Matrix<STRSXP> trimws(const Matrix<STRSXP>& x, const char* which = "both") {
-    R_xlen_t i = 0, nr = x.nrow(), nc = x.ncol(), sz = x.size();
+    R_xlen_t i = 0, sz = x.size();
+    int nr = x.nrow(), nc = x.ncol();
```
eddelbuettel: Just double-checking here:  Even with `R_xlen_t`, rows and cols are still `int` in R itself, but number of matrix elements can exceed `int`.  Do I have that correct?

krlmlr: I think that's the way it's implemented currently, see [R-ints](https://cran.r-project.org/doc/manuals/r-devel/R-ints.html#Large-matrices).

coatless: Per https://stat.ethz.ch/R-manual/R-devel/library/base/html/LongVectors.html 

> Arrays (including matrices) can be based on long vectors provided each of their dimensions is at most 2^31 - 1: 

Thus, each value is bigger than `UINT_MAX` 65,535 // 2^16 − 1. The `ULONG_MAX` 4,294,967,295 // 2^32 − 1
 has enough.

eddelbuettel: Me too, was just trying to confirm.  Because otherwise nrow, ncol should not be int here and in Matrix.  

Have a test running with the now-merged PR.  Things take some time of course.

eddelbuettel: Not uint though.  R also used signed ints.

eddelbuettel:
```r
R> .Machine$integer.max
[1] 2147483647
R> 
```

coatless: Quoted the 2byte bounds instead of 4byte blah.

eddelbuettel:
Tests are good so far (just committed for #687) so in it goes with a fresh test.

PR #691, opened 2017-05-08
Reduce -Wconversion warnings

krlmlr:
by using `size_t` instead of `int` for local variables.

3/n, I'm afraid the PRs after this will be more challenging to review...

### DIFF
--- a/inst/include/Rcpp/api/meat/module/Module.h
+++ b/inst/include/Rcpp/api/meat/module/Module.h
@@ -23,12 +23,12 @@
 namespace Rcpp {
 
     inline List Module::classes_info(){
-	    int n = classes.size() ;
+	    size_t n = classes.size() ;
 	    CharacterVector names(n) ;
 	    List info(n);
 	    CLASS_MAP::iterator it = classes.begin() ;
 	    std::string buffer ;
-	    for( int i=0; i<n; i++, ++it){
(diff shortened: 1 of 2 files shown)


eddelbuettel:
> I'm afraid the PRs after this will be more challenging to review...

Maybe let's discuss them more here on issues or on the list?   Or we keep'em in branches and don't immediately PR?  Other ideas?

krlmlr:
I could do a "for-discussion" PR in my own repo to reduce clutter here.

kevinushey:
LGTM

eddelbuettel:
@kevinushey @krlmlr when you have a moment, can you check package hBayesDM?  Something have changed there in its dependencies, but for Rcpp 0.12.10 (as on CRAN) and our master, my builds (simple R CMD INSTALL) goes crazy and completely hogs my (big-ish) machine at work.  I killed the tests by the time the load gets to 30 ...

I don't think it is us. I may just skip the package going forward.

kevinushey:
This is presumedly with the CRAN version of `hBayesDM`? I'll see if I can reproduce...

kevinushey:
I see the same thing on macOS -- `clang` is just sitting there chomping up CPU and gobbling up more and more memory (at 2.5GB now).

I suspect that whatever the issue is, it's independent of Rcpp.

PR #692, opened 2017-05-09
create src/init.c for skeleton

eddelbuettel:
This lets `Rcpp.package.skeleton()` create `src/init.c` (by making the required call).

Should `compileAttributes()` do that too?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2017-05-09  Dirk Eddelbuettel  <edd@debian.org>
+
+        * R/Rcpp.package.skeleton.R (Rcpp.package.skeleton): Under R 3.4.0, run
+        tools::package_native_routine_registration_skeleton to create src/init.c
+
 2017-05-07  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll minor version again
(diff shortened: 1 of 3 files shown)


jjallaire:
Yes I think `compileAttributes` should do this as well, however I think it should also insert a preamble (similar to it's other auto-generated files) so that if the user creates their own src/init.c then we don't overwrite it.

eddelbuettel:
I think I agree.  And if/when it does, we need to remove it from here as well.

Also need to add `.registration=TRUE` to `NAMESPACE` as this PR does.

eddelbuettel:
@jjallaire I only glanced at `src/attributes.cpp` and `R/Attributes.R` and I am not sure how we can reuse.
Or should we just duplicate 'looking for token and if not found, abort; else write token and append output from `tools::package_native_....()` ?

jjallaire:
I can take a look sometime later this week or next. The biggest issue I see is that if users have a file other than init.c which includes a package initialization function we'll be adding another one (which will break their compilation). Note this will happen on their local system rather than on CRAN so it's recoverable, but still ugly. Another possibility would be to scan for package init functions as part of our normal scanning and then only write init.c if there is no package init function.

eddelbuettel:
Point well taken on files not-named `init.c`.  We may well have to scan then.  OTOH the good news is that I have been relying on the `tools::package_native_routine_registration_skeleton()` function for several months now with many uploads, and it just works.  So this doesn't seem unsurmountable.   Wish I had thought of this fairly clean way of integrating things earlier.

eddelbuettel:
I'll fold this to tick off another micro-issue.

PR #694, opened 2017-05-16
Register native routines automatically within compileAttributes 

jjallaire:
There are two main components here:

1. Use the metadata that we already have within `compileAttributes` to automatically emit a package init function that registers all exported C++ functions with R via `R_registerRoutines`.

    Note that users have been doing this manually for several months and some packages may already have an init function for other reasons. Consequently, we only do automatic registration if no existing package init function is found.

2. Registering native functions and calling `useDynLib(foo, .registration = TRUE)` is actually not sufficient to have R wrappers *bind to* the registered routines. To do that we actually need to generate this code:

    ```r
    .Call(reticulate_py_is_none, x)
    ```

    Rather than this code:

     ```r
     .Call('reticulate_py_is_none', PACKAGE = 'reticulate', x)
     ```

     Note that we only want to generate the former code when the user has included `.registration = TRUE` (otherwise the routines won't be present in the package namespace). As a result, we scan the `NAMESPACE` file for `useDynLib(pkgname, .registration = TRUE)` and only generate R wrappers of that form if it's found.

Relevant documentation from *Writing R Extensions* is here: https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Registering-native-routines

Of particular interest is the bit about `R_useDynamicSymbols(info, FALSE)`. My reading of this is that by calling this you are saying that it still okay to to use this syntax:

```r
.Call('reticulate_py_is_none', PACKAGE = 'reticulate', x)
```

However, *only if* the symbol has been registered. This is important for us because we are now going to always pass `FALSE` to `R_useDynamicSymbols`, and we don't want that to break packages that in turn neglect to pass `.registration = TRUE` to `useDynLib`. Since we register all the Rcpp exported functions they will work using either the character based or symbol based `.Call` lookup.

This branch of the reticulate package demonstrates the change in code generation that occurs:

https://github.com/rstudio/reticulate/compare/feature/register-native-routines

Note that we've changed our `@useDynLib` roxygen to:

```
#' @useDynLib reticulate, .registration=TRUE
```

And that we've removed entirely the `init.cpp` file previously provided as a stub to pass CRAN tests.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2017-05-16  JJ Allaire  <jj@rstudio.com>
+
+        * R/Attributes.R: Automatically generate native routine registrations.
+        * src/attributes.cpp: Automatically generate native routine registrations.
+        * R/Rcpp.package.skeleton.R: Don't generate native routines when creating
+        a package using attributes.
+
 2017-05-09  Dirk Eddelbuettel  <edd@debian.org>
 
         * R/Rcpp.package.skeleton.R (Rcpp.package.skeleton): Under R 3.4.0, run
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 0.12.10.4
+Version: 0.12.10.5
 Date: 2017-05-07
 Author: Dirk Eddelbuettel, Romain Francois, JJ Allaire, Kevin Ushey, Qiang Kou,
  Nathan Russell, Douglas Bates and John Chambers
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -404,6 +404,13 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
     depends <- unique(.splitDepends(depends))
     depends <- depends[depends != "R"]
 
+    # check the NAMESPACE file to see if dynamic registration is enabled
+    namespaceFile <- file.path(pkgdir, "NAMESPACE")
+    if (!file.exists(namespaceFile))
+        stop("pkgdir must refer to the directory containing an R package")
+    pkgNamespace <- readLines(namespaceFile, warn = FALSE)
+    registration <- any(grepl("^\\s*useDynLib.*\\.registration\\s*=\\s*TRUE.*$", pkgNamespace))
+
     # determine source directory
     srcDir <- file.path(pkgdir, "src")
     if (!file.exists(srcDir))
@@ -449,7 +456,7 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
 
     # generate exports
     invisible(.Call("compileAttributes", PACKAGE="Rcpp",
-                    pkgdir, pkgname, depends, cppFiles, cppFileBasenames,
+                    pkgdir, pkgname, depends, registration, cppFiles, cppFileBasenames,
                     includes, verbose, .Platform))
 }
 
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -422,7 +422,7 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
         dir.create(rDir)
 
     # get a list of all source files
-    cppFiles <- list.files(srcDir, pattern = "\\.((c(c|pp))|(h(pp)?))$")
+    cppFiles <- list.files(srcDir, pattern = "\\.((c(c|pp)?)|(h(pp)?))$")
 
     # derive base names (will be used for modules)
     cppFileBasenames <- tools::file_path_sans_ext(cppFiles)
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -1154,3 +1154,43 @@ sourceCppFunction <- function(func, isVoid, dll, symbol) {
     as.character(token)
 }
 
+.extraRoutineRegistrations <- function(routines) {
+
+    declarations = character()
+    call_entries = character()
+
+    # if we are running R 3.4 or higher we can use an internal utility function
+    # to automatically discover additional native routines that require registration
+    if (getRversion() >= "3.4") {
+
+        # get the generated code from R
+        con <- textConnection(object = NULL, open = "w")
+        on.exit(close(con), add = TRUE)
+        tools::package_native_routine_registration_skeleton(
+            dir = ".",
+            con = con,
+            character_only = FALSE
+        )
+        code <- textConnectionValue(con)
+
+        # look for lines containing call entries
+        matches <- regexec('^\\s+\\{"([^"]+)",.*$', code)
+        matches <- regmatches(code, matches)
+        matches <- Filter(x = matches, function(x) {
+            length(x) > 0
+        })
+        for (match in matches) {
+            routine <- match[[2]]
+            if (!routine %in% routines) {
+                declaration <- grep(sprintf("^extern .* %s\\(.*$", routine), code,
+                                    value = TRUE)
+                declarations <- c(declarations, sub("^extern", "RcppExport", declaration))
+                call_entries <- c(call_entries, match[[1]])
+            }
+        }
+    }
+
+    # return extra declaratiosn and call entries
+    list(declarations = declarations,
+         call_entries = call_entries)
+}
--- a/R/Rcpp.package.skeleton.R
+++ b/R/Rcpp.package.skeleton.R
@@ -185,7 +185,9 @@ Rcpp.package.skeleton <- function(name = "anRpackage", list = character(),
         message(" >> copied the example module file ")
     }
 
-    if (getRversion() >= "3.4.0") {
+    # generate native routines if we aren't using attributes (which already generate
+    # them automatically) and we have at least R 3.4
+    if (!attributes && getRversion() >= "3.4.0") {
         con <- file(file.path(src, "init.c"), "wt")
         tools::package_native_routine_registration_skeleton(root, con=con)
         close(con)
--- a/R/Rcpp.package.skeleton.R
+++ b/R/Rcpp.package.skeleton.R
@@ -187,13 +187,15 @@ Rcpp.package.skeleton <- function(name = "anRpackage", list = character(),
 
     # generate native routines if we aren't using attributes (which already generate
     # them automatically) and we have at least R 3.4
-    if (!attributes && getRversion() >= "3.4.0") {
-        con <- file(file.path(src, "init.c"), "wt")
-        tools::package_native_routine_registration_skeleton(root, con=con)
-        close(con)
-        message(" >> created init.c for package registration")
-    } else {
-        message(" >> R version older than 3.4.0 detected, so NO file init.c created.")
+    if (!attributes) {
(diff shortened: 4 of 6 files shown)


kevinushey:
LGTM!

If I understand correctly, this means users can implicitly opt-out of Rcpp native routine registration just by making sure they provide their own call to `R_init_<pkg>` somewhere; is that right?

jjallaire:
Yes, that's right. If they have their own init function then we don't
generate any of the registration code.

However, even in that case if they do `useDynLib(pkgname, .registration =
TRUE)` then we will generate the R wrappers using symbol names rather than
character vectors. This should mean that users who prefer or who have
become accustomed to using the R 3.4 helper function will now get generated
R wrapper functions that correctly bind to the registered routines (i.e.
users that had previously been diligently registering all their routines to
pass CRAN checks were not actually *using* those routines because our
generated code in RcppExports.R was still referring to functions via
character vector rather than symbol.

kevinushey:
That sounds like the right behavior to me as well. FWIW from what I understand, the `R_useDynamicSymbols(dllInfo, FALSE)` bit implies the following:

```
.Call("routine", PACKAGE = "package") # okay
.Call(routine)                        # okay given R native routine object called 'routine' 
.Call("routine")                      # fail: won't discover routine
```

I think in theory the 'old' way we constructed Rcpp exports would still work fine -- it's only when users fail to provide the `PACKAGE =` argument will lookup fail. (Of course it's better for us to use the routine wrapper objects directly)

jjallaire:
Yes, that is my understanding as well.

eddelbuettel:
I am not entirely sure I agree with all of this.   For months now, I have been doing two things:
- add native registration by using R-devel or R 3.4.0
- add `.registration=TRUE` in `NAMESPACE

but nothing else. In particular, I did not bother to alter all the "quoted" .Call() uses.   I also never overwrote the default I on the boolean in  `R_useDynamicSymbols()`.   Grepping for this now it seems like all my C++ packages use `FALSE` and the few plain C packages (digest, RApiDatetime, RApiSerialize) use `TRUE`.

I have the suspicion I am still missing something here.

jjallaire:
Yes, that's all correct. You were setting the table properly to use native symbols however compileAttributes wasn't generating code that actually used the symbols. e.g. in my reticulate example as a result of registering the symbols the following was sitting in the reticulate namespace:

```r
reticulate:::reticulate_py_is_none
```

However, when calling "py_is_none" the string based version of lookup rather than the symbol based lookup was being used.

Net: all of that registration was really just meeting the requirements of CRAN checks but not actually put into use by the package.

The `FALSE` setting is recommended in Writing R Extensions as a good sanitary practice. There is a further call `R_forceSymbols(dll, TRUE)` you can make (also documented there) which would actually require the use of the symbols and prevent any character based lookup. We don't want to generate that code though because it would break people who haven't bother to do `useDynLib(pkgname, .registration = TRUE)`

eddelbuettel:
> Net: all of that registration was really just meeting the requirements of CRAN checks but not actually put into use by the package.

I think it is not quite as dire.  On one package (anytime, use Rcpp, has FALSE) I hit a "unpleasant many-legged creature" when altering a function signature, running (older) `compileAtttributes()` and still couldn't -- as `init.c` needed an update for the change number of arguments.  So _some_ checking was done by R.

But I think I agree: nicer calls, with the _symbol_ rather than string, are the real goal.  Getting there.

This is a huge help.  Running an RcppArmadillo test now; wonder how to best test this.  Probably need to rebuild a few packages with / without the existing `init.c`.

jjallaire:
Yes, R was definitely checking but when it came time to call the function
it was doing string based lookup.

I believe Doug Bates actually suggested this feature like a week after we
came out with Rcpp attributes and we said "good idea, we'll do that later"
(as it turned out much, much later!). His argument was actually about the
performance benefit of just having the pointer in hand vs. having to look
it up with each call, which I think for a frequently called function would
indeed add up.



On Tue, May 16, 2017 at 1:27 PM, Dirk Eddelbuettel <notifications@github.com

eddelbuettel:
The performance aspect is nice but the delta seems a little marginal from my reading of WRE.

Now, I will admit that I got totally confused a few weeks ago over one aspect I couldn't quite remember the details:  we added `Rcpp::interface()` a while back, but did we also add the 'can it be called' testing layer? I had forgotten about that.

And should we not now push harder for `Rcpp::interface()` ?

eddelbuettel:
And/or do you know a package using `Rcpp::interface()` to provide C(++) code for another package?

jjallaire:
I don't know of anyone using Rcpp::interfaces and couldn't find any on
GitHub!

I think the idea was if someone was e.g. creating a bunch of lower level
functions that would be useful for other C/C++ codebases they could do
that. In practice though it doesn't seem like people are doing this.

On Tue, May 16, 2017 at 1:51 PM, Dirk Eddelbuettel <notifications@github.com

eddelbuettel:
Yeah.  I have it in RQuantLib, and "more code got generated" (that I sort-of had forgotten about) but I also do not know of a user package ...   (And we now have header-only subset of QuantLib in Quantuccia so it is even more moot.)

eddelbuettel:
First real test:  "boom".  In `anytime`, I have `src/init.c` but Rcpp created registration in `src/RcppExports.cpp` too:

```sh
g++ -Wl,-S -shared -L/usr/lib/R/lib -Wl,-Bsymbolic-functions -Wl,-z,relro -o anytime.so RcppExports.o anytime.o init.o -L/usr/lib/R/lib -lR
init.o: In function `R_init_anytime':
/home/edd/git/anytime/src/init.c:36: multiple definition of `R_init_anytime'
RcppExports.o:/home/edd/git/anytime/src/RcppExports.cpp:136: first defined here
collect2: error: ld returned 1 exit status
/usr/share/R/share/make/shlib.mk:6: recipe for target 'anytime.so' failed
make: *** [anytime.so] Error 1
ERROR: compilation failed for package ‘anytime’
* removing ‘/usr/local/lib/R/site-library/anytime’
* restoring previous ‘/usr/local/lib/R/site-library/anytime’
edd@max:~/git/anytime(feature/numeric_input_change)$ 
```

Quick guess:  Maybe because patch currently does not scan C files?

jjallaire:
Good one and yes that was the problem. Just pushed a fix.

On Wed, May 17, 2017 at 6:46 AM, Dirk Eddelbuettel <notifications@github.com

eddelbuettel:
@jjallaire Actual bug.  Using my small-ish / recent-ish RcppQuantuccia package, I removed `src/init.c` and called `compileAttribute()`.  The `RcppExports` where updated, but:

```sh
~/git/rcppquantuccia(master)$ R CMD INSTALL .
* installing *source* package ‘RcppQuantuccia’ ...
** libs
g++ -std=gnu++11 -I/usr/share/R/include -DNDEBUG -I../inst/include/ -I. -I"/usr/local/lib/R/site-library/Rcpp/include" -I"/usr/local/lib/R/site-library/BH/include"    -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g -c RcppExports.cpp -o RcppExports.o
g++ -std=gnu++11 -I/usr/share/R/include -DNDEBUG -I../inst/include/ -I. -I"/usr/local/lib/R/site-library/Rcpp/include" -I"/usr/local/lib/R/site-library/BH/include"    -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g -c dates.cpp -o dates.o
g++ -std=gnu++11 -I/usr/share/R/include -DNDEBUG -I../inst/include/ -I. -I"/usr/local/lib/R/site-library/Rcpp/include" -I"/usr/local/lib/R/site-library/BH/include"    -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g -c utils.cpp -o utils.o
g++ -std=gnu++11 -shared -L/usr/lib/R/lib -Wl,-Bsymbolic-functions -Wl,-z,relro -o RcppQuantuccia.so RcppExports.o dates.o utils.o -L/usr/lib/R/lib -lR
installing to /usr/local/lib/R/site-library/RcppQuantuccia/libs
** R
** inst
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded
Error: package or namespace load failed for ‘RcppQuantuccia’ in .doLoadActions(where, attach):
 error in load action .__A__.1 for package RcppQuantuccia: .Call("RcppQuantuccia_RcppExport_registerCCallable", PACKAGE = "RcppQuantuccia"): "RcppQuantuccia_RcppExport_registerCCallable" not available for .Call() for package "RcppQuantuccia"
Error: loading failed
Execution halted
ERROR: loading failed
* removing ‘/usr/local/lib/R/site-library/RcppQuantuccia’
* restoring previous ‘/usr/local/lib/R/site-library/RcppQuantuccia’
~/git/rcppquantuccia(master)$ 
```

jjallaire:
Indeed! I think that this mechanism has broken the aforementioned
Rcpp::interfaces(cpp) which you must be using in this package. I'll fix
this tomorrow AM.

On Wed, May 17, 2017 at 4:21 PM, Dirk Eddelbuettel <notifications@github.com

jjallaire:
Just pushed a fix (and tested it with RcppQuantuccia).

There's one more lingering issue: if users of attributes are combining that with exporting C/C++ functions without attributes then our current code generation won't pick that up. I don't know if that scenario exists as a practical matter but it seems like somewhere in the wild it must. We have a couple different options here:

1. Under R 3.4 actually call the `tools` exported function to do the code generation (but still include it within RcppExports.cpp). That would pickup all of the exported functions not just the ones exported using attributes.

2. Say that this is a special case and if users have standalone old-style exports they should write their own package init function (however we'd probably only get the chance to communicate this *after* they regenerate and notice their package no longer loads!)

3. As an extension to #2 to prevent any build errors from ever occurring, actually detect additional old-style exports and don't write our own registration in that case (similar to how we don't write registrations when we see a package init function).

I think option (1) vs. option (2+3) are about the same amount of work. I probably favor (1).

eddelbuettel:
Thanks for the fix, will test.

I like 1) and was surprised you didn't go calling it directly.  Shall we enable that more generally, or 'just' in the case of exported functions?

jjallaire:
I implemented (1) so this case should now be covered. The reason I didn't have R do all of the code generation is that it generates different code (assumes it's in a .c rather than .cpp file) and we may need to eventually inject additional code into the init function (e.g. imagine a package author wants our generated registration but *also* wants to run some code on init, we could scan for the definition of Rcpp_init_pkgname and then call it from R_init_pkgname if it exists). Calling R from this spot within code generation also won't find the definitions in RcppExports.R since they haven't yet been written (i.e. we'd need to always inject them into the generated R code, so easier to just generate our own code in the first place).

eddelbuettel:
Ahh.  I see now.   

This should be good to merge then.

krlmlr:
Re Rcpp::interfaces: I'm actually using it for plogr (logging) and bindrcpp (creating/populating environments with bindings that call a C++ functiion).

PR #697, opened 2017-05-23
use pkg dir rather that getwd() for native routine registration

jjallaire:
Execute tools::package_native_routine_registration_skeleton within package rather than current working directory.

Typically `compileAttributes` is called with the package directory as the current working directory. However, it can be invoked via e.g. `compileAttributes(pkgdir = "foo")`. In that case we were calling `tools::package_native_routine_registration_skeleton` in the current working directory rather than in the specified `pkgdir`. This would result in missing registrations (or over-registrations if the current directory happened to be another package).

The bug should not affect most normal uses of `compileAttributes` so there isn't a need to do a patch release for this fix.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2017-05-23  JJ Allaire  <jj@rstudio.com>
+
+        * R/Attributes.R: Execute tools::package_native_routine_registration_skeleton
+          within package rather than current working directory.
+        * src/attributes.cpp: Execute tools::package_native_routine_registration_skeleton
+          within package rather than current working directory.
+
 2017-05-22  Dirk Eddelbuettel  <edd@debian.org>
 
         * R/inline.R (Rcpp.plugin.maker): Whitespace changes
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 0.12.11
+Version: 0.12.11.1
 Date: 2017-05-20
 Author: Dirk Eddelbuettel, Romain Francois, JJ Allaire, Kevin Ushey, Qiang Kou,
  Nathan Russell, Douglas Bates and John Chambers
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -1154,7 +1154,7 @@ sourceCppFunction <- function(func, isVoid, dll, symbol) {
     as.character(token)
 }
 
-.extraRoutineRegistrations <- function(routines) {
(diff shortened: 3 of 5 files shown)


jjallaire:
As it turns out this does fix the devtools problem after all.

To further clarify, there will be a potential problem if you invoke compileAttributes on a package within the working directory of *another* package that includes native routines.

So with the devtools test, it was probably invoking something like this with devtools as the working directory:

```r
devtools::install("/tmp/test-pkg-dir")
```

Wherein `compileAttributes` was called under the hood by devtools.

So as a practical matter, this will affect out-of package source directory use of devtools package building functions. Note that this is all at development time and any error will be immediately manifest and correctable by installing the development version of Rcpp (just reinforcing that I think no patch release to CRAN is required).

jimhester:
CRAN wants us to fix the failing devtools test caused by the change in behavior. 

So either there needs to be a devtools release that works around this or an Rcpp release with the fix from this PR.

We would prefer the latter, `devtools::install('xyz')` is pretty common devtools usage.

eddelbuettel:
Feel free to point them to the pull request and tell them to help themselves.

jjallaire:
Rcpp releases are huge affairs (16 hours of revdep checking, CRAN rebuilding a huge number of binaries, etc.). The devtools workaround would just be to setwd to the pkg dir before calling compileAttributes, something like this:

```r
withr_with_dir(pkg$path, {
   Rcpp::compileAttributes()
})
```

hadley:
Is there a reason this R CMD check failure wasn't reported to us during the initial round of release checks?

jjallaire:
Because devtools lists Rcpp as only a Suggests dependency rather than
Imports (and we only check Depends, Imports, LinkingTo)

eddelbuettel:
Is there a reason the devtools team didn't test the release candidate?

hadley:
Because we were never notified of it?

But this thread clearly isn't going anywhere, so I don't see any point in continuing the conversation. We'll rush a devtools fix for CRAN.

eddelbuettel:
If you depend on tool, maybe you should follow it?  

This is all spillover from the R 3.4.0 release and their somewhat poorly communicated decision and their impact on us.   We made the package registration code change late, and it was risqu'e to release.  So you got burned.  Such is life.  

I don't use devtools.  I believe a few folks with Rcpp Core do.  Still nobody noticed.  Not ideal.  

And again:  "Suggests != Depends".  My life is too short too, so I cannot test Suggests: or recursive Depends:.  One when the R Consortium builds a more powerful test battery we may get to issues like this.  Right now we don't.

PR #698, opened 2017-05-24
Do not use dir.exists to retain backwards-compatibility

epipping:
The file R/Attributes.R uses the function `dir.exists()` which
was only added in R 3.2.0. The relevant line was added to
R/Attributes.R between 0.12.5 and 0.12.6.

### DIFF
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -1129,7 +1129,7 @@ sourceCppFunction <- function(func, isVoid, dll, symbol) {
                            R.version$platform,
                            utils::packageVersion("Rcpp"),
                            sep = "-"))
-    if (!dir.exists(dir))
+    if (!utils::file_test("-d", dir))
         dir.create(dir, recursive = TRUE)
 
     dir


epipping:
If compatibility with versions older than 3.2 is important (debian 8.8 still only has 3.1) one might want to turn the `dir.exists` back into `file.exists` instead.

eddelbuettel:
Debian 8.8 is irrelevant as its sources for Rcpp match its sources for base R.   Plus, we have the actively supported backport for R available via CRAN.

We could use the [backports](https://github.com/mllg/backports) package, or at least Suggests: it if we cared.

Team:  Anybody see a problem depending on the now-two-year old R 3.2.0 ?

kevinushey:
I'm guessing there are a lot of users in enterprise environments who are stuck with R 3.1.x or even R 3.0.x who would appreciate if we could remain compatible with R 3.0.0 and above, so I would be marginally in favor of just using `file.exists()` or `utils::file_test("-d", ...)` to be backwards compatible -- especially since the change is so small in this case.

eddelbuettel:
Yes, that is a good policy.  I can look into that while traveling the next two days.  Of course I would be equally happy to receive a short PR :)

epipping:
`utils::file_test("-d", ...)` is the perfect solution for this problem. Since I see no disadvantages in following that route, for simplicity, I've now hijacked my own pull request and turned it into this different solution for the same problem by force-pushing.

jjallaire:
Yes, we have to remember that we bump our required version we implicitly bump the required version of 92% of CRAN (via recursive dependencies).

PR #703, opened 2017-06-01
Fix native registration for exports with name attribute (fixes #702)

### LINKED ISSUE #702: Name attribute in RcppExports no longer works
I updated Rcpp to version 0.12.11 from 0.12.10 and suddenly exported functions with name="" attributes no longer work.

This appears to be due to the CallEntries in RcppExport.cpp using the incorrect name to derive a function pointer.

Example exported code:
```
// [[Rcpp::export(name=".stick_breaking_mixture")]]
Rcpp::List stickBreakingMixture(
    unsigned int nLoop,
    unsigned int nWarmUp,
    //...
```

RcppExport wrapper:
```
// stickBreakingMixture
Rcpp::List stickBreakingMixture(unsigned int nLoop, unsigned int nWarmUp, ...);
RcppExport SEXP BayesSpec_stickBreakingMixture(SEXP nLoopSEXP, SEXP nWarmUpSEXP, ...) {
```

Corresponding CallEntries entry:
```
    {"BayesSpec_.stick_breaking_mixture", (DL_FUNC) &BayesSpec_.stick_breaking_mixture, 9},
```
I think this function pointer should be `&BayesSpec_stickBreakingMixture`

Note that I have checked and this is not due simply to the use of '.' in the exported name.

jjallaire:


eddelbuettel:
Thanks for the quick PR.  Will merge, and the possibly rearrange ChangeLog and NEWS slightly to make it clearer that is work on its way to 0.12.12 ...

PR #705, opened 2017-06-02
Automatically register init functions for RcppModules (closes #704)

### LINKED ISSUE #704: Rcpp Modules not working via skeleton generator due to dynamic registration
https://stackoverflow.com/questions/44331117/rcpp-modules-not-working-in-rcpp-0-12-11

@jjallaire we got some more testing / tuning to do :-/

jjallaire:
Addresses https://github.com/RcppCore/Rcpp/issues/704

RcppModules include a native routine "boot" function of the form `_rcpp_module_boot_*`. These functions are not discovered by `tools::package_native_routine_registration_skeleton` so generally need to be hand-added to `init.c`.

This also implies that our current auto-generation of native routine registrations for `RcppExports.cpp` won't register these functions and will therefore fail to create a loadable package. 2 alternatives here:

1) When we see an Rcpp module simply don't attempt to generate native routine registrations (sort of like we do now when we see `init.c`). 

2) Generate native routine registrations for the Rcpp modules.

Happily, we already parse for Rcpp module declarations (since `sourceCpp` supports exporting modules) so it's a very straightforward change to generate the `_rcpp_module_boot_*` functions, which is what we do in this PR.

### DIFF
--- a/R/Rcpp.package.skeleton.R
+++ b/R/Rcpp.package.skeleton.R
@@ -138,7 +138,6 @@ Rcpp.package.skeleton <- function(name = "anRpackage", list = character(),
             file.copy(file, src)
             message(" >> copied ", file, " to src directory" )
         }
-        compileAttributes(root)
     }
 
     if (example_code) {
@@ -146,10 +145,6 @@ Rcpp.package.skeleton <- function(name = "anRpackage", list = character(),
             file.copy(file.path( skeleton, "rcpp_hello_world_attributes.cpp"),
                       file.path( src, "rcpp_hello_world.cpp"))
             message(" >> added example src file using Rcpp attributes")
-            compileAttributes(root)
-            message(" >> compiled Rcpp attributes")
-            message(" >> do NOT modify by hand either RcppExports.cpp or ",
-                    "RcppExports.R")
         } else {
             header <- readLines(file.path(skeleton, "rcpp_hello_world.h"))
             header <- gsub("@PKG@", name, header, fixed = TRUE)
@@ -216,6 +211,11 @@ Rcpp.package.skeleton <- function(name = "anRpackage", list = character(),
 
     if (isTRUE(remove_hello_world)) {
         rm("rcpp_hello_world", envir = env)
+    }
+
+    if (attributes) {
+        compileAttributes(root)
+        message(" >> compiled Rcpp attributes ")
     }
 
     invisible(NULL)
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 2017-06-02  JJ Allaire  <jj@rstudio.com>
 
         * src/attributes.cpp: Automatically register init functions for RcppModules
+        * R/Rcpp.package.skeleton.R: compileAttributes only after all code is generated
 
 2017-06-01  JJ Allaire  <jj@rstudio.com>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2017-06-02  JJ Allaire  <jj@rstudio.com>
+
+        * src/attributes.cpp: Automatically register init functions for RcppModules
+
 2017-06-01  JJ Allaire  <jj@rstudio.com>
 
         * src/attributes.cpp: Fix registration for exports with name attribute.
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -17,6 +17,7 @@
       or newer (Elias Pipping in \ghpr{698}).
       \item Fix native registration for exports with name attribute (\ghpr{703}
       addressing \ghit{702}).
+      \item Automatically register init functions for RcppModules.
     }
   }
 }
--- a/src/attributes.cpp
+++ b/src/attributes.cpp
@@ -675,8 +675,11 @@ namespace attributes {
         // for generating C++ interfaces
         std::vector<Attribute> cppExports_;
 
-        // for generating native routine registration
+        // for generating Rcpp::export native routine registration
         std::vector<Attribute> nativeRoutines_;
+
+        // for generating module native routine registration
+        std::vector<std::string> modules_;
     };
 
     // Class which manages generating PackageName_RcppExports.h header file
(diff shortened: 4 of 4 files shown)


eddelbuettel:
Nicely done.

eddelbuettel:
Shame on me for not escalating this in March when I adjusted the example we unit test.

eddelbuettel:
Not _quite_ there yet. 

After fresh install from this branch:

```sh
/tmp$ Rscript -e "Rcpp::Rcpp.package.skeleton('foo', module = TRUE)"  Creating directories ...
Creating DESCRIPTION ...
Creating NAMESPACE ...
Creating Read-and-delete-me ...
Saving functions and data ...
Making help files ...
Done.
Further steps are described in './foo/Read-and-delete-me'.

Adding Rcpp settings
 >> added Imports: Rcpp
 >> added LinkingTo: Rcpp
 >> added useDynLib directive to NAMESPACE
 >> added import(methods, Rcpp) directive to NAMESPACE
 >> added example src file using Rcpp attributes
 >> compiled Rcpp attributes
 >> do NOT modify by hand either RcppExports.cpp or RcppExports.R
 >> added Rd file for rcpp_hello_world
 >> copied the example module file 
/tmp$ R CMD INSTALL foo
* installing *source* package ‘foo’ ...
** libs
ccache g++ -I/usr/share/R/include -DNDEBUG  -I"/usr/local/lib/R/site-library/Rcpp/include"    -fpic  -g -O3 -Wall -pipe -pedantic -Wextra -Wno-empty-body -Wno-unused -c Num.cpp -o Num.o
ccache g++ -I/usr/share/R/include -DNDEBUG  -I"/usr/local/lib/R/site-library/Rcpp/include"    -fpic  -g -O3 -Wall -pipe -pedantic -Wextra -Wno-empty-body -Wno-unused -c RcppExports.cpp -o RcppExports.o
ccache g++ -I/usr/share/R/include -DNDEBUG  -I"/usr/local/lib/R/site-library/Rcpp/include"    -fpic  -g -O3 -Wall -pipe -pedantic -Wextra -Wno-empty-body -Wno-unused -c rcpp_hello_world.cpp -o rcpp_hello_world.o
ccache g++ -I/usr/share/R/include -DNDEBUG  -I"/usr/local/lib/R/site-library/Rcpp/include"    -fpic  -g -O3 -Wall -pipe -pedantic -Wextra -Wno-empty-body -Wno-unused -c rcpp_module.cpp -o rcpp_module.o
ccache g++ -I/usr/share/R/include -DNDEBUG  -I"/usr/local/lib/R/site-library/Rcpp/include"    -fpic  -g -O3 -Wall -pipe -pedantic -Wextra -Wno-empty-body -Wno-unused -c stdVector.cpp -o stdVector.o
g++ -shared -L/usr/lib/R/lib -Wl,-Bsymbolic-functions -Wl,-z,relro -o foo.so Num.o RcppExports.o rcpp_hello_world.o rcpp_module.o stdVector.o -L/usr/lib/R/lib -lR
installing to /usr/local/lib/R/site-library/foo/libs
** R
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded
Error: package or namespace load failed for ‘foo’ in .doLoadActions(where, attach):
 error in load action .__A__.1 for package foo: loadModule(module = "NumEx", what = TRUE, env = ns, loadNow = TRUE): Unable to load module "NumEx": Failed to initialize module pointer: Error in FUN(X[[i]], ...): no such symbol _rcpp_module_boot_NumEx in package foo

Error: loading failed
Execution halted
ERROR: loading failed
* removing ‘/usr/local/lib/R/site-library/foo’
/tmp$ 
```

jjallaire:
I think the problem is that compileAttributes is called within
Rcpp::package.skeleton *before* the module code is added. I'll change the
order and it should resolve things.

On Fri, Jun 2, 2017 at 12:17 PM, Dirk Eddelbuettel <notifications@github.com

jjallaire:
Okay, just pushed a fix

eddelbuettel:
Works like a charm now.

R/Rcpp.package.skeleton.R
```diff
@@ -218,6 +213,11 @@ Rcpp.package.skeleton <- function(name = "anRpackage", list = character(),
         rm("rcpp_hello_world", envir = env)
     }
 
+    if (attributes) {
+        compileAttributes(root)
+        message(" >> compiled Rcpp attributes ")
+    }
+
```
eddelbuettel: Maybe I am short of coffee bu `if (attributes) {` relies on `attributes==TRUE`.  We could have a case of `modules=TRUE` (as in #704) _and also_ `attributes=FALSE`, no?   Needless complication?

jjallaire:
I think it's a little more involved than first appears, if we do this:

```bash
Rcpp::Rcpp.package.skeleton('foo', module = TRUE, attributes = FALSE)
```

Then the `attributes = FALSE` causes us to generate init.c (which will in turn cause compileAttributes to do no registration). So calling compileAttributes in that case won't actually change the output, but rather will generate an RcppExports.cpp that has nothing in it.

I'd say if the user asks for modules and no attributes then they mean no auto-generation behavior and they'll futz with init.c manually themselves.

eddelbuettel:
Agreed.  Let's go with this now and clean up spillage should it occur.  `Rcpp.package.skeleton()` is already used less than it used to given the 'make me a package' feature in a certain IDE.  And use of modules is even less common.  And modules with attributes=FALSE ... may just never happen.

BenoitLondon:
Thanks for the fix! Do you know when you will release it? I wanted to know if I should use master or wait for the release.

eddelbuettel:
@BenoitLondon See my [recent post to the rcpp-devel list](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2017-June/009602.html) with pointers to the drat repo.  You can install Rcpp 0.12.11.2 from there.  (Master works too but the git repo could be unstable.)

BenoitLondon:
Thanks

PR #706, opened 2017-06-04
Protect function argument to rcpp_set_stack_trace()

krlmlr:
which *may* be gc-ed otherwise.

I was unable to create a reproducible example for this problem, but I have [two stack traces](https://gist.github.com/krlmlr/501043e57573abe9d47d13be69fb88ff) from test runs for dplyr with valgrind and gctorture enabled that indicate that this may be the source of a problem. I'm not seeing these errors anymore after applying this fix.

### DIFF
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -8,6 +8,7 @@
     \item Changes in Rcpp API:
     \itemize{
       \item The \code{tinyformat.h} header now ends in a newline (\ghit{701}).
+      \item Fixed rare protection error that occurred when fetching stack traces during the construction of an Rcpp exception (Kirill Müller in \ghit{706}).
     }
     \item Changes in Rcpp Attributes:
     \itemize{
--- a/src/barrier.cpp
+++ b/src/barrier.cpp
@@ -121,7 +121,6 @@ SEXP rcpp_get_stack_trace() {
 
 // [[Rcpp::register]]
 SEXP rcpp_set_stack_trace(SEXP e) {
-    Rcpp::Shield<SEXP> ep(e);
     SET_VECTOR_ELT(get_rcpp_cache(), 3, e);
     return R_NilValue;
 }
--- a/src/barrier.cpp
+++ b/src/barrier.cpp
@@ -121,6 +121,7 @@ SEXP rcpp_get_stack_trace() {
 
 // [[Rcpp::register]]
 SEXP rcpp_set_stack_trace(SEXP e) {
+    Rcpp::Shield<SEXP> ep(e);
     SET_VECTOR_ELT(get_rcpp_cache(), 3, e);
     return R_NilValue;
 }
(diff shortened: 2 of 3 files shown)


inst/NEWS.Rd
```diff
@@ -20,6 +20,10 @@
       \item Automatically register init functions for Rcpp Modules (\ghpr{705}
       addressing \ghit{704}).
     }
+    \item Bug fixes:
```
eddelbuettel: There has never been an entry Bug fixes.  Please place under Rcpp API.

src/barrier.cpp
```diff
@@ -121,7 +121,9 @@ SEXP rcpp_get_stack_trace() {
 
 // [[Rcpp::register]]
 SEXP rcpp_set_stack_trace(SEXP e) {
+    PROTECT(e);
```
eddelbuettel: Should this be `Rcpp::Shield<>` instead?

eddelbuettel:
Also, remember that we (generally) prefer Issue tickets prior to PRs...

krlmlr:
Next time I have a reprex I file an issue first. Promise ;-)

eddelbuettel:
Thanks for quickly making the changes.  Does it still pass _your_ local test?

kevinushey:
Thanks for catching this! I don't think this is quite the right change, though.

Normally, we should be able to operate under the assumption that `SEXP` arguments received by functions are already protected by someone in a parent scope; e.g. this is what is done for methods called from R through the `.Call()` interface and I think we should preserve these semantics in other C++-specific code as well.

It looks to me like the real issue here is that the `SEXP` returned by calls to `stack_trace()` is not protected, e.g. here:

https://github.com/RcppCore/Rcpp/blob/7687508db0a15f252e9a5c902d36d3addd87da24/inst/include/Rcpp/exceptions.h#L34

https://github.com/RcppCore/Rcpp/blob/7687508db0a15f252e9a5c902d36d3addd87da24/inst/include/Rcpp/exceptions.h#L39

And so the returned `SEXP` from `stack_trace()` should be protected before being sent to those function calls.

eddelbuettel:
> Next time I have a reprex I file an issue first. Promise ;-)

Well whether you want it or not we're having the review now :)

krlmlr:
@kevinushey: Good point, thanks for chiming in.

@eddelbuettel: I'm fine with that.

krlmlr:
Concerning the local test: I can no longer replicate the original heisenbug. I'll run a test on dplyr using this branch of Rcpp with valgrind and gctorture.

krlmlr:
FWIW, `Rf_ScalarString()` does protect its argument.

eddelbuettel:
> FWIW, Rf_ScalarString() does protect its argument.

Meaning we'd no longer need this?  Colour me confused.    

In any event, this is unlikely to do harm, and as @kevinushey pointed out both a good catch anyway and now in a better spot so I'll merge.

krlmlr:
I was pointing out that some functions seem to be caller-protected, others callee-protected. But then `Rf_ScalarString()` may just be the big exception of the rule, I don't know.

lionel-:
The rule in the R source code is that the caller protects arguments and the callee can expect its arguments to be protected. There are documented exceptions, e.g. `setAttrib()`.

PR #708, opened 2017-06-05
make Rcpp compile and install on haiku-os

armgong:
make Rcpp compile and install on haiku-os , but don't know how to write unit test

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -33,7 +33,7 @@ using namespace Rcpp;
 #endif
 
 #if defined(__GNUC__) || defined(__clang__)
-    #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX) || defined(__MUSL__)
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Looks good.  And "we have your word" that this helps on [Haiku-OS](https://en.wikipedia.org/wiki/Haiku_(operating_system)) as we cannot test?

armgong:
run unit test on haiku-os and R 3.3.1 gcc 5.4.0 , one unit test file crash R (runit.wrap.R will try find what cause R crash ) , all others success.

[Rcpp-unitTests.txt](https://github.com/RcppCore/Rcpp/files/1052235/Rcpp-unitTests.txt)

eddelbuettel:
I don't follow. I see

> Number of test functions: 558 
> Number of errors: 0 
> Number of failures: 0

armgong:
this is one crash (in runit.wrap.R) and I have to remove this  file from tests
Executing test function test.wrap.map.string.double  ... 
 *** caught segfault ***
 address 0x12574d2dfb0, cause 'memory not mapped'
 An irrecoverable exception occurred. R is aborting now ...
 Segmentation violation

eddelbuettel:
Oh I see.  The string library implementation may be different.  

But yes, by all means please investigate.  And no hurry.

armgong:
I will investigate it in weekend ,thanks a lot

PR #713, opened 2017-06-14
Add Shield around parameters in Rcpp::interfaces (fixes #712)

### LINKED ISSUE #712: Protection problems with C++ interfaces
My package "bindrcpp" provides a C++ interface via `// [[Rcpp::interfaces(cpp)]]`. I was able to track down spurious segmentation faults (in dplyr, https://github.com/tidyverse/dplyr/issues/2811) to (what I think is) a protection error in the auto-generated C++ wrappers for these interfaces.

The commit https://github.com/krlmlr/bindrcpp/commit/6d1e2bde24fd8a26a3911923cd2e6cd0a1a29180 shows both the problem and the solution: In general, `wrap()` will perform allocation of R objects, and without extra shielding these can be garbage collected even before the wrapped function is called, because the latter has all `SEXP` in its signature.

After applying this particular commit, I'm not seeing test failures anymore, which before occurred after a few iterations. Happy to submit a PR.

CC @lionel-.

jjallaire:


eddelbuettel:
Hah. Saves me some testing and means I can go straight to rev.dep testing this :)

src/attributes.cpp
```diff
@@ -2172,7 +2172,7 @@ namespace attributes {
 
                 const std::vector<Argument>& args = function.arguments();
                 for (std::size_t i = 0; i<args.size(); i++) {
-                    ostr() << "Rcpp::wrap(" << args[i].name() << ")";
+                    ostr() << "Shield<SEXP>(Rcpp::wrap(" << args[i].name() << "))";
```
eddelbuettel: We could make this Rcpp::Shield for consistency with Rcpp::wrap but I see other identifiers without the namespace qualifier.

jjallaire:
Yeah I originally had it with the qualifier then noticed the same thing you
did (already have unqualified references elsewhere so why press the issue
here).

On Wed, Jun 14, 2017 at 3:14 PM, Dirk Eddelbuettel <notifications@github.com

eddelbuettel:
Looks good, and have a rev.dep run chugging along.  May as well merge now.

coatless:
Would it be worth it to prefix with `Rcpp::` where applicable?

kevinushey:
The fact that we're injecting a `using namespace Rcpp;` statement into another package's namespace within a header file is a little bit wonky, so it might be worth doing.

jjallaire:
I don't recall why that using namespace Rcpp is there but I don't think I
would have done it capriciously so I wouldn't just rip it out and hope for
the best. It doesn't pollute the namespace of users of the package, just
the implementation namespace of the package (which is by definition using
the Rcpp namespace).

eddelbuettel:
Also, aren't we talking `RcppExports.cpp` here -- a single compilation unit and not included into other files.

kevinushey:
Good point :) Okay, let's not touch anything there.

krlmlr:
Thanks for the very quick fix!

eddelbuettel:
Thanks to you for the heads-up. I also ran a full rev.dep check yesterday, and we're good.  I did sixteen out of 1050 or so packages failing, and about eight looked like possible dplyr interactions I had not seen (and do no worry about for Rcpp as they also fail at CRAN for the releases version -- hence unlikely to be us).
See https://github.com/RcppCore/rcpp-logs/commit/1f85c64724edfbb834d25a0ba50f72f3fd0ba2d2

krlmlr:
Thanks. Most of the eight new failures look familiar, I hope the upcoming dplyr point release will fix them.

PR #715, opened 2017-06-17
Explicit cast from size_t to int

krlmlr:
to support `-Wconversion`.


krlmlr:
Thanks. The other option is to store the data as `R_xlen_t`, which seems slightly more future-proof but does change the API of the class.

eddelbuettel:
I _think_ `dim` is always an integer vector.

PR #717, opened 2017-06-18
additional #nocov tags, reformating wrap.h

eddelbuettel:


eddelbuettel:
I'll quietly merge this, there wasn't much here.  `wrap.h` got reindented, else it's just `#nocov` tags.

PR #720, opened 2017-06-27
explicitly make exported functions visible

jeroen:
Marks exported Rcpp functions as visible, on supported platforms. The `attribute_visible ` macro is defined in `<R_ext/Visibility.h>`. If the compiler supports visibility it is defined as 

```c
 __attribute__ ((visibility ("default")))
```
And otherwise it is blank. The idea is that we can compile packages with `-fvisibility=hidden` as defined in `make` variable `$(C_VISIBILITY)` on supported platforms. So pkg authors can do:

```make
PKG_CXXFLAGS=$(C_VISIBILITY)
```

Limiting visibility to exported functions reduces the chances of symbol conflicts like we saw in https://github.com/ropensci/hunspell/issues/21. Here both rstudio and the hunspell package include a copy of libhunspell but different versions. Apparently their C++ classes start conflicting under certain circumstances. Hiding the `libhunspell` symbols from the dll solved the problem.

### DIFF
--- a/inst/include/RcppCommon.h
+++ b/inst/include/RcppCommon.h
@@ -126,7 +126,7 @@ namespace Rcpp {
 #include <Rcpp/complex.h>
 #include <Rcpp/barrier.h>
 
-#define RcppExport extern "C"
+#define RcppExport extern "C" attribute_visible
 
 #include <Rcpp/exceptions.h>
 


eddelbuettel:
Interesting.  I was so far only aware of a user-space clash between Boost and RStudio which I had hit with anytime and (the new, non CRAN) RcppMLPACK.  This got fixed in the RStudio dailies a little while ago thanks to @kevinushey 

I am always a little wary of making a global change but this one may be sensible. Maybe to wait till I get home though.

Alternatively could we change this to defining and inly if another variable is defined?  I seem to have

```
C_VISIBILITY = -fvisibility=hidden
F77_VISIBILITY = -fvisibility=hiddenb
```

so should we condition on that?

jeroen:
I think the conditioning is already done by R itself. Header `R_ext/Visibility.h` has:

```c
#ifdef HAVE_VISIBILITY_ATTRIBUTE
# define attribute_visible __attribute__ ((visibility ("default")))
# define attribute_hidden __attribute__ ((visibility ("hidden")))
#else
# define attribute_visible
# define attribute_hidden
#endif
```

Where `HAVE_VISIBILITY_ATTRIBUTE` is defined in `Rconfig.h` if the compiler supports it. When `HAVE_VISIBILITY_ATTRIBUTE` is not defined, both `C_VISIBILITY` and `attribute_visible` are blank.

eddelbuettel:
Thanks for that, very helpful. I very vaguely recalled it but did not have time to hunt.

jjallaire:
Seems like a good change to me, I'll merge it. I can't forsee any problems with revdeps but we'll see on the next pass!

Issue #723, opened 2017-06-30
Registration, autogenerated symbols and 'no documentation' false positive


eddelbuettel:
What can we do when `.registration=TRUE` and we now generate globally-visible identifiers that have to be R and C++ compatible?   

Under the default exporting regexp of `exportPattern("^[[:alpha:]]+")` these are visible, so R nags.

We cannot prefix with dot or underscore because (I think) in either case one language whines.  But if we do nothing, R whines.  Shall we (auot-)create an Rd (or roxygen) stub?

jjallaire:
I don't fully get the scenario so hard to say. What's a concrete example of a function that triggers this error?

I personally don't love the `exportPattern("^[[:alpha:]]+")` as there are typically more non-exported than exported functions in a package and this forces users into naming functions with dot prefixes. But I realize it's common convention so something we need to deal with.

We could always generate a single catch-all Rd file sort of like roxygen does for re-exports. e.g. https://github.com/rstudio/keras/blob/master/man/reexports.Rd

eddelbuettel:
> I don't fully get the scenario so hard to say. What's a concrete example of a function that triggers this error?

It hit me during the class. If we use the package generators (either via `Rcpp*.package.skeleton()` or the equivalent generators in RStudio) then we get a wild-carded `exportPattern(...)`.  _Personally_ I am fine editing it but for newbies it sucks.

Agreed on the re-exports file. I know how to write those in Rd, I failed to get it done with `roxygenize()` when I tried earlier today on the train.   I guess the .Rd route is the way to go then.

jjallaire:
It seems like `.registration = TRUE` and `exportPattern("^[[:alpha:]]+")` together is an anti-pattern, as it will result in spamming the search path with a bunch of intermediate generated functions with awkward names (i.e. prefaced with `pkgname_`).

I wonder if we should move away from generating packages with `exportPattern("^[[:alpha:]]+")` entirely. I agree that it's trickier for beginners to get that they need to add exports to `NAMESPACE`, but the alternative that they never learn about `NAMESPACE` and have to use dot prefacing for all private functions seems just as bad.

eddelbuettel:
I hear you (and I wholeheartedly agree about the anti-pattern).  

But the pattern is what R does by default.  I am a little hesitant to move away from it.

I think ultimately the best may be to have two regexps: all matching 'x' yet excluding to cover our per-package pattern.   Defensible idea?

[...]

PR #725, opened 2017-07-02
Generate C++ native routines with underscore prefix

jjallaire:
Avoid export from package when standard [[ alpha ]] exportPattern is used in NAMESPACE

This addresses https://github.com/RcppCore/Rcpp/issues/723.

Note that this change does introduce the use of reserved (but not invalid) C/C++ symbols (the underscore prefix). We made some effort last year I recall to remove such symbols from generated attributes code, presumably to get a clean bill of health from linting tools?

I've tried this with Rcpp::interfaces(r, cpp) as well as with the tic.drat package (package with dot in the name). I've also tried it with packages that include their own non-generated native functions. Also of course have run the Rcpp tests. Even still, this will change the code in every existing RcppExports.* file so is worthy of some extra review/caution/testing.


jjallaire:
Note that if R core decides to stop auto-exporting the registered native symbols then we could revert the use of the underscore prefix. That's something that couldn't happen until the release of R 3.5 though.

PR #730, opened 2017-07-08
Inteface change: accept R_xlen_t instead of int in no_init()

krlmlr:
This is one of the two changes necessary to compile RSQLite with `-Wconversion` without warnings.


eddelbuettel:
Passed a full reverse-depends check with flying colour, merging.

PR #731, opened 2017-07-08
Interface change: use R_xlen_t for loop unrolling

krlmlr:
To compile RSQLite with `-Wconversion -Werror`.

### DIFF
--- a/inst/include/Rcpp/internal/wrap.h
+++ b/inst/include/Rcpp/internal/wrap.h
@@ -116,9 +116,9 @@ namespace Rcpp {
             Shield<SEXP> x(Rf_allocVector(RTYPE, size));
 
             typedef typename ::Rcpp::traits::storage_type<RTYPE>::type STORAGE;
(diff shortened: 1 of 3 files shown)


krlmlr:
Thanks for reviewing. Is it important to update the copyright year in the files? I have missed a few instances (also in the other PR).

eddelbuettel:
I do it when I remember it (which I usuallly do) but I am also not that religious about.  You can add a commit, or we clean up another time...

kevinushey:
LGTM, although I would (very marginally) prefer to see us selecting `R_xlen_t` vs. `int` depending on whether the input `SEXPTYPE` supports long vectors or not. Perhaps something for a separate PR some day?

eddelbuettel:
Fair point. I guess a reverse depends check will give us some clues.

eddelbuettel:
Passed a full reverse-depends check with flying colour, merging.

PR #737, opened 2017-08-04
Update Module.h

assassin5615:
Fixed compilation error in LOAD_RCPP_MODULE

### DIFF
--- a/inst/include/Rcpp/Module.h
+++ b/inst/include/Rcpp/Module.h
@@ -453,8 +453,8 @@ static VARIABLE_IS_NOT_USED SEXP moduleSym = NULL;
 
 // this macro is called by code wanting to load a module -- see RInside's rinside_module_sample0.cpp
 #define LOAD_RCPP_MODULE(NAME) \
-    Shield<SEXP> __load_module_call__( Rf_lang2( GET_MODULE_SYM, _rcpp_module_boot_##NAME() ); \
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Nice.  Can we possibly complement it with a test case?  Do you have a usage example?

assassin5615:
actually, I am also wondering how this macro is intended to be used. currently, I am only using this macro as a reference for how to load a Module.

I am working on something that runs R inside c++, to use this Macro directly, it needs to be changed to something like LOAD_RCPP_MODULE(NAME, SEXP) \..... SEXP=Rf_eval(...), then I can write following code
`RInside R(argc, argv);
LOAD_RCPP_MODULE(Test, R['Test'])`

anyone knows why this macro was added in the first place?

eddelbuettel:
How Modules get loaded changed once (basically from the R side). This macro may or may not predate that change, I am not sure at the moment.  I always trigger the load from the R side.

eddelbuettel:
What name+email should I use for you in the ChangeLog?

assassin5615:
you can use Lei Yu <leiyu_1981@qq.com>.  

I am working on something that needs to load module from the c++ side.  I have to assign the value returned by Rf_eval in the macro to something like R['my_own_module'] to use it.  if it can be used directly(no need to assign the return value to something), the macro is useful when it comes to loading module in c++.

PR #763, opened 2017-10-02
Avoid Coverity warning with Rostream::Rostream()

### LINKED ISSUE #760: Resource leak, and other warnings from Coverity
When analyzing RMariaDB, a warning most likely related to Rcpp came up.

Source code location: https://github.com/RcppCore/Rcpp/blob/f9b700abcf97c408786150de0fd65adf6823699e/inst/include/Rcpp/iostream/Rstreambuf.h#L49

Problem: The `new Buffer` seems to be never freed. This might be a false positive, but maybe not.

Screenshot (because Coverity reports are private):

screenshot from 2017-10-02 09-20-11

The other warning is about a constructor that doesn't initialize class members in tinyformat.h:

https://github.com/RcppCore/Rcpp/blob/f9b700abcf97c408786150de0fd65adf6823699e/inst/include/Rcpp/utils/tinyformat.h#L497

Screenshot:

screenshot from 2017-10-02 09-25-09

I've ignored the exception warnings so far, but we may take a look at them later.

krlmlr:
Fixes #760. Coverity report follows.

### DIFF
--- a/inst/include/Rcpp/iostream/Rstreambuf.h
+++ b/inst/include/Rcpp/iostream/Rstreambuf.h
@@ -43,18 +43,11 @@ namespace Rcpp {
     template <bool OUTPUT>
     class Rostream : public std::ostream {
         typedef Rstreambuf<OUTPUT> Buffer ;
-        Buffer* buf ;
+        Buffer buf ;
     public:
         Rostream() :
-            std::ostream( new Buffer ),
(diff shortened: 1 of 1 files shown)


krlmlr:
Coverity accepts this code without warnings.

eddelbuettel:
Will merge.

For next time please remember that ChangeLog (trivial via Emacs, still easy enough by hand) and NEWS.Rd entries are always appreciated too.

PR #767, opened 2017-10-30
Feature/deprecate old date datetime

eddelbuettel:


eddelbuettel:
The third commit is a needed for a change in behaviour / regression in pinp discussed [here](https://github.com/eddelbuettel/pinp/issues/44).  It's generally better without an "always forced" `\pnasbreak` but on occassion we still seem to need one -- as in these two vignettes (but not the others).

eddelbuettel:
Had some issue with my rev.dep setup which delayed this.  All looking good.

`pinp` had another release as there were minor side effects. This may require a follow-up commit to clean up https://github.com/RcppCore/Rcpp/pull/767/commits/b3fa43ca3202b8d6f2594ecb89d2bc3e7a3df868 here.

eddelbuettel:
All good, _lots_ of reverse-depends checks committed [here](8104fa79c09f59dfaa5d0cd4214569c0e64b7f8c).  Will need to update RQuantLib and RcppQuantuccia to the new date vectors (and that code is either ready or easy to add).

PR #768, opened 2017-11-05
Added `inject()` and `extract()` to `Vector`

dcdillon:
This will allow a user to bitwise copy data from C++ structures into elements in an R vector of arithmetic types.  For instance, a `int64_t` could be stored in an `Rcpp::NumericVector` not by value but as a bitwise copy.  This data could then be operated on accordingly in R.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2017-11-05  Daniel C. Dillon <dcdillon@gmail.com>
+
+        * inst/include/Rcpp/vector/Vector.h: Added inject() and extract()
+        functions to allow insertion and retrieval of raw C++ data into
+        and from Rcpp::Vector of arithmetic types.
+        * inst/unitTests/cpp/Vector.cpp: Unit tests
+        * inst/unitTests/runit.Vector.R: Unit tests
+
 2017-11-04  Dirk Eddelbuettel  <edd@debian.org>
 
         * vignettes/Rcpp-FAQ.Rmd: Add 'skip_final_break: true' toggle
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -13,7 +13,9 @@
       allocating one (Kirill Müller in \ghpr{763}).
       \item New \code{DateVector} and \code{DatetimeVector} classes are now the
       default fully deprecating the old classes as announced one year ago.
-      
+      \item New \code{Vector::inject()} and \code{Vector::extract()} functions
+      for copying raw byte data from C++ into an element or elements in
+      \code{Vector}.
     }
     \item Changes in Rcpp Package:
     \itemize{
--- a/inst/include/Rcpp/vector/Vector.h
+++ b/inst/include/Rcpp/vector/Vector.h
@@ -378,28 +378,24 @@ public:
     inline typename traits::enable_if<
         traits::is_arithmetic< stored_type >::value
         && sizeof(T) <= sizeof(stored_type), void >::type
-    inject(R_xlen_t i, const T &val)
-    {
-        memcpy(&cache.ref(offset(i)), &val, sizeof(T));
+    inject(R_xlen_t i, const T &val) {
+	    std::memcpy(&cache.ref(offset(i)), &val, sizeof(T));
     }
     
     template< typename T >
     inline typename traits::enable_if<
         traits::is_arithmetic< stored_type >::value
         && sizeof(T) == sizeof(stored_type), void >::type
-    inject(R_xlen_t start, const T *val, std::size_t len)
-    {
-        memcpy(&cache.ref(offset(start)), val, sizeof(T) * len);
+    inject(R_xlen_t start, const T *val, std::size_t len) {
+	    std::memcpy(&cache.ref(offset(start)), val, sizeof(T) * len);
     }
     
     template< typename T >
     inline typename traits::enable_if<
         traits::is_arithmetic< stored_type >::value
         && sizeof(T) < sizeof(stored_type), void >::type
-    inject(R_xlen_t start, const T *val, std::size_t len)
-    {
-        for (std::size_t i = 0; i < len; ++i)
-        {
+    inject(R_xlen_t start, const T *val, std::size_t len) {
+        for (std::size_t i = 0; i < len; ++i) {
             inject(start + i, *(val + i));
         }
     }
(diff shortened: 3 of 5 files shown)


eddelbuettel:
(Context here is my need to get `int64_t` representing nanosecond resolution time into `NumericVector` so that the S4 class `nanotime` can be instantiated.  Had an earlier solution, @dcdillon helper with a better one.  Naming is kinda up in the air: we found `raw*` to be natural for getter/setter but then there is possibly confusion with R's `raw` type.)

dcdillon:
@eddelbuettel, indeed.  If anyone has a better suggestion, I'm all ears.  Changing the names is the simplest thing in the world to do.

dcdillon:
I should explain how it works:

For inserting a single value at a single spot in the array we have:

```
template< typename T >
Rcpp::Vector::inject(R_xlen_t i, const T &value)
```

This function will only compile if the `stored_type` of the `Rcpp::Vector` is arithmetic (in practice if it is `int32_t` or `double` which are 4 and 8 bytes long respectively).  Additionally, the `sizeof(T)` must be less than or equal to `sizeof(stored_type)`.

If this is all true, then the code will compile as a bitwise copy of `value` into the first `sizeof(T)` bytes of the appropriate `stored_type`.

We have also, an array variant of `inject()` as follows:

```
template< typename T >
Rcpp::Vector::inject(R_xlen_t start, const T *value, std::size_t len)
```

There are two cases of this function.
* If `sizeof(T) == sizeof(stored_type)` this will simply be equivalent to `memcpy()` the array at `value` into the array starting at the index `start` of the R storage.
* If `sizeof(T) < sizeof(stored_type)` this will `inject()` each individual element in the `value` array into the corresponding element in the R storage starting at the index `start`.

If `sizeof(T) > sizeof(stored_type)` it will fail to compile.

The `extract()` functions work in exactly the same way.

This is done, so that we can copy bitwise data into R, but it preserves the notion of the R representation being a vector of elements.

kevinushey:
IMHO operations that circumvent the type system (putting a `U` in a `T`) shouldn't be part of the interface of the class. I would marginally prefer not changing Rcpp at all, but just document how existing tools in Rcpp + STL can be used to accomplish this.

FWIW here's a small example showing how we might accomplish this just using a plain old `memcpy()`, just using `bit64` to show that we indeed did accomplish what we set out to do:

````
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
NumericVector int64() {
  int64_t item = 123;
  NumericVector v(1);
  std::memcpy(v.begin(), &item, sizeof item);
  return v;
}

/*** R
library(bit64)
x <- int64()
class(x) <- "integer64"
x
*/
````

Giving, as expected:

```
> x
integer64
[1] 123
```

**Comment by kevinushey** [comment]
I am a fan of the extra compile-time safety provided here so perhaps that gives it enough reason to be included? I still somewhat feel that these shouldn't be part of the `Vector<T>` class but instead should be free functions in the Rcpp namespace (or similar).

**Comment by dcdillon** [comment]
I just write the code ;).  Whatever you guys agree on is fine by me.

**Comment by eddelbuettel** [comment]
Good example, @kevinushey and _almost_ what I ended up doing.  You can set the S3 attribute at the C++ side:  

```c++
// [[Rcpp::export]]
NumericVector int64() {
  int64_t item = 123;
  NumericVector v(1);
  std::memcpy(v.begin(), &item, sizeof item);
  v.attr("class") = "integer64";         // added
  return v;
}
```

The only thing biting me now is that ... we need to call the R function `nanotime()` on `v` as I haven't managed to set the right S4 attributes on `v` (and part of is may be how `nanotime` is a S4 wrapper around an S3 class, namely `integer64`.

/cc @lsilvest

kevinushey:
What about having free functions that construct Rcpp vectors from arbitrary types? Rather than attempting to inject arbitrary memory into an existing memory, we could allow the user to write code like:

```
int64_t data = 123;
NumericVector x = NumericVector::embed(data);
```

It seems like this sort of action is more common than the more general 'inject arbitrary memory at some location in an Rcpp type'? I'm curious what @eddelbuettel thinks.

eddelbuettel:
Either is fine.  Not sure which I find more "idiomatic".

dcdillon:
Either way it's set up, if we don't give elementwise access and modification to it, someone will want it.  So both the full `embed()` of an array and a resetting of particular elements once it's been `embed()'ed` will likely be necessary.

dcdillon:
Just to present another use case, this could open the door for an implementation of a `ranged_double` or somesuch in R.  We could define a C++ structure as follows:

```
struct ranged_double
{
   int32_t offset:24;
   int32_t base_index:8;
};
```
This structure takes 32 bytes and would fit in a `integer` vector.  We could then add attributes to the R class called `base` and `increment` which are vectors of `numeric`.

The actual value of the element in the vector is `base[base_index] + increment[base_index] * value`.  This would essentially halve the memory footprint of a vector of `numeric` at the expense of the calculation shown above.

I'm just making the argument that there is value in having an R compatible basic type (`double`, `int32_t`) and then reinterpreting it on the other side via some mechanism (like `bit64` does).

This has no bearing on how the implementation falls out, just goes to the potential usefulness.

eddelbuettel:
Could have legs as a longer-term project as we "may have to" redo `bit64` and its `(long long)` casts.  `int64_t` and C++ cleanlyness would be better.

Then again, it works and is a big piece of work.  If it ain't broke...

dcdillon:
...then rewrite it in c++17

dcdillon:
I don't care when we make a decision, but if we don't come to a conclusion today, this will wait until next weekend at the earliest.

eddelbuettel:
Hm.  Tried (and failed) to push a clean-up commit to the @dcdillon branch.  Must have missed a step.

So we now have a new branch called pr768 that is a commit ahead of master.  @dcdillon can you cherryr-pick that one commit in?   Or is there a better way?   I may be absent-minded...

eddelbuettel:
Did you pick my commit?  Or replicated what was in it?

dcdillon:
Just replicated.  I was running out the door

eddelbuettel:
Having simple similar code in semi-regular use now as free functions, I am no longer sure we need `inject` and `extract` all that much as member of `vector`.   Other thoughts?

dcdillon:
Any thoughts on this?  It's the weekend again and I can do or not do code I expect.  I suppose we could even make it an add-on package `RcppDoBadThingsWithMemory` or perhaps `RcppBits` if we think providing a safe way of doing this is worthwhile.  Otherwise we can drop it and put it in the "If you want to do scary things, don't look at us" bucket.

I come from an industry where we do many bad things with bits in the name of performance, so I'm biased to providing some sort of interface, but, by the same token, I know how to do bad things with bits on my own.  So I don't know what the right thing to do is.

eddelbuettel:
On the margin I think I am with @kevinushey that this may well be a free function, or even remain out of Rcpp.

I have something now I am happy with, and I may just write up what I do for a piece in the Rcpp Gallery, and thank @dcdillon and @lsilvest profusely for helping me get there.  But keep changes out of Rcpp.

eddelbuettel:
So general consensus after more discussion is to not merge this PR.  Free functions work fine, Dan may wrap some up in a helper package, and I just [wrote something up for the Rcpp Gallery](http://gallery.rcpp.org/articles/creating-integer64-and-nanotime-vectors/).

PR #777, opened 2017-11-25
correct calls to Rf_warning (closes #776)

### LINKED ISSUE #776: Rcpp::warning does not follow the correct interface for Rf_warning
[`Rf_warning`](https://github.com/wch/r-source/blob/eb468006b82d96917db88e2310286b54a27b47b7/src/main/errors.c#L296-L308) is defined with a function prototype like that of printf

> void warning(const char *format, ...)

However `Rcpp::warning()` calls it instead with only a null terminated string (I linked here to the C11 code, but the C89 code has the same problem).

https://github.com/RcppCore/Rcpp/blob/fc3c0ed269458b69fbd60ee42005b83990cdca3f/inst/include/Rcpp/exceptions/cpp11/exceptions.h#L45-L48

This means that any values passed to `Rcpp::warning()` which happen to have format specifiers in them will access random memory, outputting garbage or potentially crashing R.

For a simple example illustrating the issue.

``` r
Rcpp::cppFunction('void test() { Rcpp::warning("%s", "a str with %o in it"); }')
test()
#> Warning in test(): a str with 1543134670 in it
```

eddelbuettel:
Quick code reviews would be welcome and appreciated   
@coatless @jimhester @kevinushey @nathan-russell

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,15 @@
+2017-11-25  Dirk Eddelbuettel  <edd@debian.org>
+
+        * DESCRIPTION (Version, Date): Roll minor version
+
+        * inst/include/Rcpp/exceptions/cpp11/exceptions.h: Ensure Rf_format gets
+        a format string
+        * inst/include/Rcpp/exceptions/cpp98/exceptions.h (Rcpp): Ditto
+
 2017-11-18  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll minor version
-        
+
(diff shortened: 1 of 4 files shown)


jimhester:
LGTM, thanks Dirk!

eddelbuettel:
Well thanks to you for opening the ticket. It's a good thing to be fixes, so good you stumbled on it in xml2 and sorry it neeeded a change at your end.

eddelbuettel:
Passed rev.dep checks, so in it goes.

PR #779, opened 2017-12-01
Fixed missing Rcpp namespace in export interface generation

coatless:
This PR adds the `Rcpp::` namespace prefix to an `as<>()` call in the Rcpp Attributes generation procedure for a C++ interface. e.g.,

```
// [[Rcpp::interfaces(r, cpp)]]
```

In particular, appending `Rcpp::`, fixes the following errors:

```
error: expected '(' for function-style cast or type construction
            throw Rcpp::exception(as<std::string>(rcpp_result_gen).c_str());
                                     ~~~~~~~~~~~^
error: no member named 'c_str' in 'Rcpp::RObject_Impl<PreserveStorage>'
            throw Rcpp::exception(as<std::string>(rcpp_result_gen).c_str());
                                                 ~~~~~~~~~~~~~~~~~ ^
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2017-11-30  James J Balamuta  <balamut2@illinois.edu>
+
+        * src/attributes.cpp: Fixed missing Rcpp namespace in export interface
+        generation
+
 2017-11-25  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll minor version
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -9,6 +9,12 @@
     \itemize{
       \item Calls from exception handling to `Rf_warning()` now correctly set an
       initial format string (Dirk in \ghpr{777} fixing \ghit{776}).
+    }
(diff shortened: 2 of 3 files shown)


eddelbuettel:
Can you remind me please in which issue ticket we discussed this?

eddelbuettel:
OTOH seems harmless and minimal enough to not even warrant a ChangeLog or NEWS entry. I mean it add what, six characters in one place?

coatless:
We didn't. I avoided filing one as it is a one-line fix isolated to cross-package exports with:

```
// [[Rcpp::interfaces(r, cpp)]
```

eddelbuettel:
Out of curiousity, how did yout tickle this?  That line has probably been three a while.  I don't often set a `cpp` interface but as I recall I do in at least one package...   Will just merge.

coatless:
I was using `cpp` interface to share code between multiple packages. The generation under the first package went okay, but while trying to generate an interface with the second package an error was raised.

eddelbuettel:
Nice.  "Gallery post" ?  I don't think we have usage examples of that.

Issue #780, opened 2017-12-01
Strange behavior when assigning values in Environment, due to GC


wch:
I've been running into a very hard-to-debug problem which I believe is due to a GC during assignment to an Environment. In some cases, the result of assignment into an Environment has the wrong type, and in other cases, the Environment somehow gets lost.

Here is an example to run in a clean R session:

```R
library(Rcpp)
cppFunction(
  code = '
  Environment f() {
    Environment env = Function("new.env")();
    env["A"] = "a";
    env["B"] = "b";
    env["C"] = "c";
    env["FOO"] = "foo";
    return env;
  }
  '
)

gctorture(TRUE)
e <- f()
gctorture(FALSE)
e$A
e$B
e$C
e$FOO
```

When I run this on my Mac, here's what it prints out for the last commands:

```
> gctorture(TRUE)
> e <- f()
> gctorture(FALSE)
> e$A
[1] "a"
> e$B
<CHARSXP: "B">
> e$C
[1] "c"
> e$FOO
<CHARSXP: "FOO">
```

Inspecting the objects gives this:

```
> .Internal(inspect(e$A))
@7fd168ce9c68 16 STRSXP g1c1 [MARK,NAM(2)] (len=1, tl=0)
  @7fd168868598 09 CHARSXP g1c1 [MARK,gp=0x61] [ASCII] [cached] "a"
> .Internal(inspect(e$B))
@7fd168ce9c98 09 CHARSXP g1c1 [MARK,NAM(2),gp=0x61] [ASCII] [cached] "B"
```

If I run the code a second time, then the results turn out normal. This is the output from a second run:

```
> gctorture(TRUE)
> e <- f()
> gctorture(FALSE)
> e$A
[1] "a"
> e$B
[1] "b"
> e$C
[1] "c"
> e$FOO
[1] "foo"
```

I get the exact same result when I run the same code in the r-base Docker image. To reproduce:

```
docker run --rm -ti r-base
install.packages('Rcpp')
# Now run the code from above
```

When I run the same code in a custom build of R-devel with [PROTECTCHECK](https://github.com/wch/r-source/blob/5b2b46fc/src/main/memory.c#L121-L149) defined, I get something else weird:

```
....
> gctorture(TRUE)
> e <- f()
> gctorture(FALSE)
> e$A
Error: object 'e' not found
> e$B
Error: object 'e' not found
> e$C
Error: object 'e' not found
> e$FOO
Error: object 'e' not found
```

[...]

PR #784, opened 2017-12-04
protect the temporary SEXPs produced by wrap

kevinushey:
Should resolve https://github.com/RcppCore/Rcpp/issues/780.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2017-12-04  Kevin Ushey  <kevinushey@gmail.com>
+
+        * inst/include/Rcpp/api/meat/proxy.h: Protect temporary SEXP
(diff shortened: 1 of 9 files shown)


eddelbuettel:
Thanks Kevin -- looks very good, will probably merge tomorrow!

kevinushey:
I went through and cleaned up a number of other usages of `wrap()` -- some of these changes err on the side of caution, but I think it's worth it since the stack-based protection mechanism is quite cheap.

eddelbuettel:
@kevinushey I presume you saw but something in the last one upset the unit tests.

eddelbuettel:
FYI, full rev dep running, 100 packages in and picture perfect.  Will probably merge by tomorrow.

Issue #771, opened 2017-11-14
R_NewHashedEnv should not be in Rcpp?


mattfidler:
I am using the following code in my package:

```c++
...
Environment e = new_env();
...
```
I get the following:

```
* checking compiled code ... NOTE
File 'RxODE/libs/x64/RxODE.dll':
  Found non-API call to R: 'R_NewHashedEnv'

Compiled code should not call non-API entry points in R.
```

Apparantly, this isn't allowed for CRAN packages...?

This is called in the following file

https://github.com/RcppCore/Rcpp/blob/8f2a101789b29e572bf3abe5115d9ddb48c1b0c7/inst/include/Rcpp/Environment.h#L405

Hopefully this is reproducable enough for a bug report.

mattfidler:
Here's my sessionInfo()
```
R version 3.4.2 (2017-09-28)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 14393)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] RxODE_0.6-3     devtools_1.13.3

loaded via a namespace (and not attached):
 [1] rex_1.1.2                 Rcpp_0.12.13             
 [3] xml2_1.1.1                magrittr_1.5             
 [5] roxygen2_6.0.1            lattice_0.20-35          
 [7] R6_2.2.2                  brew_1.0-6               
 [9] RcppArmadillo_0.8.100.1.0 dparser_0.1.8            
[11] stringr_1.2.0             tools_3.4.2              
[13] grid_3.4.2                R.oo_1.21.0              
[15] withr_2.1.0               commonmark_1.4           
[17] rprojroot_1.2             assertthat_0.2.0         
[19] lazyeval_0.2.1            digest_0.6.12            
[21] crayon_1.3.4              Matrix_1.2-11            
[23] R.utils_2.6.0             testthat_1.0.2           
[25] mvnfast_0.2.2             memoise_1.1.0            
[27] inline_0.3.14             stringi_1.1.5            
[29] compiler_3.4.2            backports_1.1.1          
[31] desc_1.1.1                PreciseSums_0.1          
[...]

PR #785, opened 2017-12-05
use public R APIs for new_env

kevinushey:
Closes https://github.com/RcppCore/Rcpp/issues/771. We use `Rcpp::Function` + `new.env()` instead of an (unfortunately) unexported R API for constructing environments.

### DIFF
--- a/inst/include/Rcpp/api/meat/Environment.h
+++ b/inst/include/Rcpp/api/meat/Environment.h
@@ -44,6 +44,15 @@ Environment_Impl<StoragePolicy>::Environment_Impl( int pos ){
    Storage::set__(env) ;
 }
 
+inline Environment new_env(int size = 29) {
+    Function newEnv("new.env");
+    return newEnv(_["size"] = size, _["parent"] = R_EmptyEnv);
+}
+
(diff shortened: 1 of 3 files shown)


inst/include/Rcpp/api/meat/Environment.h
```diff
@@ -44,6 +44,15 @@ Environment_Impl<StoragePolicy>::Environment_Impl( int pos ){
    Storage::set__(env) ;
 }
 
+inline Environment new_env(int size = 29) {
+    Function newEnv("new.env");
```
eddelbuettel: Should we be fancy and memoise / keep a once init'ed `Function` ?  Probably not worth it.

kevinushey: We could do that (just make it a static object so it's only initialized once on first call). Certainly doesn't hurt but agree we don't gain much.

eddelbuettel:
I think I'll just squash merge this later and then run a new rev dep.  Unless you want to rebase...

kevinushey:
Squash merge sounds good to me.

PR #789, opened 2017-12-14
Use protect-unwind API and add Rcpp_fast_eval()

lionel-:
* Use protected evaluation when compiling Rcpp against R 3.5 (see https://github.com/wch/r-source/commit/c8cc6084e40d737b2dc39b1857f7b3ae1850b451). This prevents leaks when R code jumps (caught condition, restart invokation, interrupt, return, error).

* Add `Rcpp_fast_eval()`. Unlike `Rcpp_eval()`, it doesn't wrap the code in `tryEval()` to avoid the catching overhead. It falls back to `Rcpp_eval()` when compiled against old versions of R so packages can use it without worrying about backward compatibility.

  Here are the differences between the new function and `Rcpp_eval()`:
    - Better performance.
    - R errors are not rethrown on the C++ stack as `Rcpp::condition`. Instead they are rethrown as `Rcpp::internal::LongjumpException` as with all other longjump causes.
    - It has the control flow semantics of the C function `Rf_eval()` rather than the R function `base::eval()`. E.g. you can return from a frame environment on the R stack.

These changes should be fully backward compatible. I have tested this branch on R oldrel, release and devel here: https://travis-ci.org/lionel-/Rcpp/builds/316466207 (without vignettes and with the binary package test disabled). I'm not sure how to adapt the current Travis config to use R devel.

I have checked this branch with dplyr (which calls into R a lot) on both R release and R devel. I have not run the reverse dependencies checks. Would it be possible to run those on your server Dirk?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,44 @@
+
+2017-12-13  Lionel Henry  <lionel@rstudio.com>
+
+        * inst/include/Rcpp/api/meat/Rcpp_eval.h: Add Rcpp_fast_eval() for safe
+        and fast evaluation of R code using the new protect-unwind API in R 3.5.
+        Unlike Rcpp_eval(), this does not evaluate R code within tryCatch() in
+        order to avoid the catching overhead. R longjumps are now correctly
+        intercepted and rethrown. Following this change the C++ stack is now
+        safely unwound when a longjump is detected while calling into R code.
+        This includes the following cases: caught condition of any class, long
+        return, restart jump, debugger exit.
+
+        Rcpp_eval() also uses the protect-unwind API in order to gain safety.
+        To maintain compatibility it still catches errors and interrupts in
+        order to rethrow them as typed C++ exceptions. If you don't need to
+        catch those, consider using Rcpp_fast_eval() instead to avoid the
+        overhead.
+
+        These improvements are only available for R 3.5.0 and greater. When
+        compiled with old versions of R, Rcpp_fast_eval() falls back to
+        Rcpp_eval(). This is in contrast to internal::Rcpp_eval_impl() which
+        falls back to Rf_eval() and which is used in performance-sensititive
+        places.
+
+        Note that with this change, Rcpp_eval() now behaves like the C function
+        Rf_eval() whereas it used to behave like the R function base::eval().
+        This has subtle implications for control flow. For instance evaluating a
+        return() expression within a frame environment now returns from that
+        frame rather than from the Rcpp_eval() call. The old semantics were a
+        consequence of using evalq() internally and were not documented.
+
+        * inst/include/Rcpp/exceptions.h: Add LongjumpException and
+        resumeJump() to support Rcpp_fast_eval().
+
+        * inst/include/Rcpp/macros/macros.h: Catch LongjumpException and call
+        resumeJump(). If resumeJump() doesn't jump (on old R versions), throw an
+        R error (this normally should not happen).
+
+        * inst/include/RcppCommon.h: Add Rcpp_fast_eval() to the public API and
+        internal::Rcpp_eval_impl() to the private API.
+
 2017-12-05  Kevin Ushey  <kevinushey@gmail.com>
 
         * inst/include/Rcpp/Environment.h: Use public R APIs
--- a/ChangeLog
+++ b/ChangeLog
@@ -39,6 +39,9 @@
         * inst/include/RcppCommon.h: Add Rcpp_fast_eval() to the public API and
         internal::Rcpp_eval_impl() to the private API.
 
+        * inst/include/Rcpp/Environment.h: Use safe evaluation
+        * inst/include/Rcpp/Language.h: idem
+
 2017-12-05  Kevin Ushey  <kevinushey@gmail.com>
 
         * inst/include/Rcpp/Environment.h: Use public R APIs
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,9 +16,10 @@
         catch those, consider using Rcpp_fast_eval() instead to avoid the
         overhead.
 
-        These improvements are only available for R 3.5.0 and greater. When
-        compiled with old versions of R, Rcpp_fast_eval() falls back to
-        Rcpp_eval(). This is in contrast to internal::Rcpp_eval_impl() which
+        These improvements are only available for R 3.5.0 and greater. You also
+        need to explicitly define `RCPP_PROTECTED_EVAL` before including Rcpp.h.
+        When compiled with old versions of R, Rcpp_fast_eval() always falls back
+        to Rcpp_eval(). This is in contrast to internal::Rcpp_eval_impl() which
         falls back to Rf_eval() and which is used in performance-sensititive
         places.
 
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -11,6 +11,20 @@
       set an initial format string (Dirk in \ghpr{777} fixing \ghit{776}).
       \item The 'new' Date and Datetime vectors now have \code{is_na} methods
       too. (Dirk in \ghpr{783} fixing \ghit{781}).
+
+      \item Evaluation of R code is now safer when compiled against R
+      3.5. Longjumps of all kinds (condition catching, returns,
+      restarts, debugger exit) are appropriately detected and handled,
+      e.g. the C++ stack unwinds correctly. The new function
+      \code{Rcpp_fast_eval()} can be used for performance-sensitive
+      evaluation of R code. Unlike \code{Rcpp_eval()}, it does not try
+      to catch errors with \code{tryEval} in order to avoid the catching
+      overhead. While this is safe thanks to the stack unwinding
+      protection, this also means that R errors are not transformed to
+      an \code{Rcpp::exception}. If you are relying on error rethrowing,
+      you have to use the slower \code{Rcpp_eval()}. On old R versions
+      \code{Rcpp_fast_eval()} falls back to \code{Rcpp_eval()} so it is
+      safe to use against any versions of R.
     }
     \item Changes in Rcpp Attributes:
     \itemize{
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -11,20 +11,21 @@
       set an initial format string (Dirk in \ghpr{777} fixing \ghit{776}).
       \item The 'new' Date and Datetime vectors now have \code{is_na} methods
       too. (Dirk in \ghpr{783} fixing \ghit{781}).
-
       \item Evaluation of R code is now safer when compiled against R
-      3.5. Longjumps of all kinds (condition catching, returns,
-      restarts, debugger exit) are appropriately detected and handled,
-      e.g. the C++ stack unwinds correctly. The new function
-      \code{Rcpp_fast_eval()} can be used for performance-sensitive
-      evaluation of R code. Unlike \code{Rcpp_eval()}, it does not try
-      to catch errors with \code{tryEval} in order to avoid the catching
-      overhead. While this is safe thanks to the stack unwinding
-      protection, this also means that R errors are not transformed to
-      an \code{Rcpp::exception}. If you are relying on error rethrowing,
-      you have to use the slower \code{Rcpp_eval()}. On old R versions
-      \code{Rcpp_fast_eval()} falls back to \code{Rcpp_eval()} so it is
-      safe to use against any versions of R.
+      3.5 (you also need to explicitly define \code{RCPP_PROTECTED_EVAL}
+      before including \code{Rcpp.h}). Longjumps of all kinds (condition
+      catching, returns, restarts, debugger exit) are appropriately
+      detected and handled, e.g. the C++ stack unwinds correctly.
+      \item The new function \code{Rcpp_fast_eval()} can be used for
+      performance-sensitive evaluation of R code. Unlike
+      \code{Rcpp_eval()}, it does not try to catch errors with
+      \code{tryEval} in order to avoid the catching overhead. While this
+      is safe thanks to the stack unwinding protection, this also means
+      that R errors are not transformed to an \code{Rcpp::exception}. If
+      you are relying on error rethrowing, you have to use the slower
+      \code{Rcpp_eval()}. On old R versions \code{Rcpp_fast_eval()}
+      falls back to \code{Rcpp_eval()} so it is safe to use against any
+      versions of R.
     }
     \item Changes in Rcpp Attributes:
     \itemize{
--- a/inst/include/Rcpp/Environment.h
+++ b/inst/include/Rcpp/Environment.h
@@ -109,7 +109,7 @@ namespace Rcpp{
 
             /* We need to evaluate if it is a promise */
             if( TYPEOF(res) == PROMSXP){
-                res = Rf_eval( res, env ) ;
+                res = internal::Rcpp_eval_impl( res, env ) ;
             }
             return res ;
         }
@@ -129,7 +129,7 @@ namespace Rcpp{
 
             /* We need to evaluate if it is a promise */
             if( TYPEOF(res) == PROMSXP){
-                res = Rf_eval( res, env ) ;
+                res = internal::Rcpp_eval_impl( res, env ) ;
             }
             return res ;
         }
@@ -151,7 +151,7 @@ namespace Rcpp{
 
             /* We need to evaluate if it is a promise */
             if( TYPEOF(res) == PROMSXP){
-                res = Rf_eval( res, env ) ;
+                res = internal::Rcpp_eval_impl( res, env ) ;
             }
             return res ;
         }
@@ -174,7 +174,7 @@ namespace Rcpp{
 
             /* We need to evaluate if it is a promise */
             if( TYPEOF(res) == PROMSXP){
-                res = Rf_eval( res, env ) ;
+                res = internal::Rcpp_eval_impl( res, env ) ;
             }
             return res ;
         }
--- a/inst/include/Rcpp/Language.h
+++ b/inst/include/Rcpp/Language.h
@@ -145,10 +145,10 @@ namespace Rcpp{
         }
 
         SEXP fast_eval() const {
-            return Rf_eval( Storage::get__(), R_GlobalEnv) ;
+            return internal::Rcpp_eval_impl( Storage::get__(), R_GlobalEnv) ;
         }
         SEXP fast_eval(SEXP env ) const {
-            return Rf_eval( Storage::get__(), env) ;
+            return internal::Rcpp_eval_impl( Storage::get__(), env) ;
         }
 
         void update( SEXP x){
--- a/inst/include/Rcpp/api/meat/Rcpp_eval.h
+++ b/inst/include/Rcpp/api/meat/Rcpp_eval.h
@@ -39,7 +39,15 @@ namespace internal {
 
     inline void Rcpp_maybe_throw(void* data, Rboolean jump) {
         if (jump) {
-            throw LongjumpException(static_cast<SEXP>(data));
+            SEXP token = static_cast<SEXP>(data);
+
+            // Keep the token protected while unwinding because R code might run
+            // in C++ destructors. Can't use PROTECT() for this because
+            // UNPROTECT() might be called in a destructor, for instance if a
+            // Shield<SEXP> is on the stack.
+            ::R_PreserveObject(token);
+
+            throw LongjumpException(token);
         }
     }
 
--- a/inst/include/Rcpp/api/meat/Rcpp_eval.h
+++ b/inst/include/Rcpp/api/meat/Rcpp_eval.h
@@ -21,15 +21,15 @@
 #include <Rcpp/Interrupt.h>
 #include <Rversion.h>
 
-#if (defined(R_VERSION) && R_VERSION >= R_Version(3, 5, 0))
-#define R_HAS_UNWIND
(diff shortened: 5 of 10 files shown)


eddelbuettel:
>  Would it be possible to run those on your server Dirk?

Sure!  But that is only using r-release.  So we'd know of regressions.  No more, no less.

lionel-:
At first it may be best to condition the use of the protect-unwind API with an explicit `#define`.

eddelbuettel:
Which is not in the PR, correct?  So why rev.dep check runs with it on unconditionally?

lionel-:
Your revdeps should opt out automatically of the new API because they are compiled against R release. I am wondering whether it's best to also opt out of it by default on R 3.5 until `R_UnwindProtect()` gets more widely tested. Maybe that's not necessary if the next release of Rcpp comes after 3.5 is out and the revdep checks with protected evaluation enabled turn out fine.

eddelbuettel:
We tend(ed) to release Rcpp every two months.  By that timeline it will be January and March before R 3.5.

lionel-:
Ok, so it seems best to condition so it is not automatically enabled when R 3.5 comes out. Does `#define RCPP_PROTECTED_UNWIND` sound good? Cf last commit.

eddelbuettel:
Reverse depends concluded, I just [committed the summary](https://github.com/RcppCore/rcpp-logs/commit/dcd6930ee4eccd7ca2fcefed7151c326d1899954)

No compilation errors.  A few new errors in testing from `testthat` which may be due to changes in that package. I will try to find some time test those against stock Rcpp later.

eddelbuettel:
We're good. Those ten new ones were all spurious (two missing depends, one phantom gone under re-run, seven apparently due to testthat as they also fail under previous Rcpp tests).

lionel-:
Great. Do you think we could get this merged for the next Rcpp release? Having it on CRAN will make it easier to start testing `R_ProtectUnwind()` in our packages by enabling `RCPP_PROTECT_UNWIND` on the R devel Travis builds.

eddelbuettel:
I think I'll just merge it as it, and we build from here.  I trust the subsequent commits you made; I made let it run here once more with it.  

Looks like a job well done (though we'll only know for sure with R 3.5.*).  And you of course get extra brownie points for embracing the Emacs-ish ChangeLog use.  At long last someone does :)

kevinushey:
Maybe not for this PR, but what do you think about making `Rcpp_eval()` and `Rcpp_fast_eval()` behave the same; ie, letting `Rcpp_eval()` use the new implementation (for R >= 3.5.0)?

Our `Rcpp_eval()` implementation basically invokes something of this form at the R level:

```
tryCatch(evalq(<expr>, <envir>), warning = identity, error = identity, interrupt = identity)
```

The intention of this code is to ensure that R longjmps are 'blocked' at this scope, so that longjmps never go 'up' the C++ stack. How does having the `tryCatch()` handlers active interact with the longjmp unwind protection here?

lionel-:
> what do you think about making Rcpp_eval() and Rcpp_fast_eval() behave the same

One potential problem is that `Rcpp::condition` is not internal so there's probably existing code that tries to handle it at C++ level. We could assess the feasibility of this by changing the type of the thrown condition to `Rcpp::internal::condition` in a branch and running the revdeps on it.

> How does having the tryCatch() handlers active interact with the longjmp unwind protection here?

There is no interaction for errors and interrupts because the tryCatch prevents the R code from jumping in these cases. The stack is still unwound with the existing mechanism (even if the error was a normally returned condition and not the result of catching an actual error/interrupt jump). All other longjump causes are detected and a `Rcpp::internal::LongjumpException` is thrown.

kevinushey:
> One potential problem is that Rcpp::condition is not internal so there's probably existing code that tries to handle it at C++ level. We could assess the feasibility of this by changing the type of the thrown condition to Rcpp::internal::condition in a branch and running the revdeps on it.

Fair enough. We might want to give that a shot later (in a separate PR).

> There is no interaction for errors and interrupts because the tryCatch prevents the R code from jumping in these cases. The stack is still unwound with the existing mechanism (even if the error was a normally returned condition and not the result of catching an actual error/interrupt jump). All other longjump causes are detected and a Rcpp::internal::LongjumpException is thrown.

Okay, awesome! Thanks for clarifying.

lionel-:
By the way, if I'm not mistaken we could replace the `evalq()` with the `quote()` primitive and evaluate the tryCatch in the user-supplied environment (with the `tryCatch()` closure inlined in the expression in case the environment does not inherit from base or masks `base::tryCatch()`).

This should improve performance a bit and give `Rcpp_eval()` the semantics of `Rf_eval()` (which matters for stack-sensitive expressions like `return()`, `sys.xxx()`, etc). Code that assumes a specific call stack trace would have to be adapted though.

kevinushey:
> By the way, if I'm not mistaken we could replace the evalq() with the quote() primitive and evaluate the tryCatch in the user-supplied environment (with the tryCatch() closure inlined in the expression in case the environment does not inherit from base or masks base::tryCatch()).

> This should improve performance a bit and give Rcpp_eval() the semantics of Rf_eval() (which matters for stack-sensitive expressions like return(), sys.xxx(), etc). Code that assumes a specific call stack trace would have to be adapted though.

Unfortunately, I'm fairly sure this is code of that form out there already in R packages so we might not be able to make that change, unless we had a convincing example of commonly-written R/Rcpp code that does the wrong thing in the current scheme.

lionel-:
Ok, so it seems that merging `Rcpp_fast_eval()` and `Rcpp_eval()` is not feasible as it would have the same problem by cleaning the call stack trace even further.

kevinushey:
Fair enough. Perhaps we can just document for new users that `Rcpp_fast_eval()` is preferred for R packages targeting R >= 3.5.0.

lionel-:
Yup, probably best to wait until R 3.5 is out and the protected API enabled by default in Rcpp.

eddelbuettel:
(Mostly just nodding along...)

eddelbuettel:
Hi @lionel quick question: was some of the material in the PR activated for stack unwinding?

With r-devel changing so much I had not pushed to win-builder in some time, and this evening I just got an error ending on 

```
Executing test function test.stackUnwinds  ... 
This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
terminate called after throwing an instance of 'Rcpp::internal::LongjumpException'
```
on 32 bit mode on r-devel. Full link is [here](https://win-builder.r-project.org/kGrw98mgNoEG/) for the usual time that win-builder keeps it.

lionel-:
> was some of the material in the PR activated for stack unwinding?

Only in the `test.stackUnwinds` unit test. From the message it seems like the `LongjumpException` is not caught by the wrapping macros. I just tried on R-hub (windows-x86_64-devel) and it seems to work properly (some unit tests fail but not due to exception catching) https://artifacts.r-hub.io/Rcpp_0.12.14.5.tar.gz-ca8f8d1eaf454fdea479be9b68bbeec4/Rcpp.Rcheck/00check.log

I'm not sure how the platform can impact the wrapping macros or exception catching. I am on vacations for one week and then I'll be busy preparing a webinar, so I won't have time to look into this further until January 25th.

One easy way around this is to disable the test for now. It'd also be better to extract the test in its own file so the protection stuff is not enabled for the other tests in `runit.misc.R` / `misc.cpp`. I'll send a PR to that effect.

eddelbuettel:
Ok, noted on the timing--bad luck, and my fault for not trying windows r-devel earlier, despite the current changes. 

As it is about release time for Rcpp, I will indeed just condition out the test.  As it works on Linux, I will only run it there

Don't send a PR just to isolate one test into another, that's not worth it.  Enjoy the vacation...

I agree that

eddelbuettel:
Turns out there may be more trouble.  

Your test is inactive, but I still have trouble with windows tests on either i386 or x86_64.  Very, very weird. Still in the middle of it.

eddelbuettel:
Ok, a few hours of fiddling mostly with R Hub (as I don't have a Windows machine) and I think I can say we are at this.  Notes are a little sketchy.  I refer to versions (which are committed) and in some cases the commit sha1s (should have been more consistent).  Order is by time, not order in which I tested.
I _think_ just proved that #789 is indeed the bad boy.


- 0.12.14.3 was pivoted just before Lionel's pr789
    - passes win builder
    - did not try rhub

- 0.12.14.3.1 was just after Lionel's pr789 
    - branch 'local/before790' at 1630a55b
    - added skip of Lionel's test
    - http://builder.r-hub.io/status/Rcpp_0.12.14.3.1.tar.gz-c3812d0bcbcd430c8f691c6abb3a3191
    - times out at rhub

- 0.12.14.4 was just after #790 by Kevin
    - added skip of Lionel's test
    - http://builder.r-hub.io/status/Rcpp_0.12.14.4.tar.gz-614f7dd1ed444823bb5aedf5404d44e9
    - times out at rhub 
    
- 0.12.14.7 was another more minimalistic attempt "from the current end" without unwinw
    - times out at still at rhub
        
- 0.12.14.7.1 was `git checkout -b local/beforePR797`
    - times out at rhub

kevinushey:
I can try testing on my Windows VM just to confirm what could be going on as well.

eddelbuettel:
I am totally lost at see now.  So I had bisected before/after @lionel 's #789 and thought I had evidence that was the change.

But then I created a new (local) branch, reverted the commits ... and it is still fried.  Fsck.

I called that one 0.12.14.7 just to be different.  Results / artfecacts at

HTML: https://builder.r-hub.io/status/Rcpp_0.12.14.7.tar.gz-e09ebc89ba764981a0615d2a614b5c9e                                                                                                                       
Text: https://builder.r-hub.io/status/original/Rcpp_0.12.14.7.tar.gz-e09ebc89ba764981a0615d2a614b5c9e                                                                                                              
Artifacts: https://artifacts.r-hub.io/Rcpp_0.12.14.7.tar.gz-e09ebc89ba764981a0615d2a614b5c9e

kevinushey:
On my Windows VM, running with 32bit R + Rcpp master, I see the following running with R-devel r74099:

```
> # On errors - Always unwound
>         unwound <- FALSE
> out <- tryCatch(testEvalUnwind(quote(stop("err")), unwound), error = identity)
terminate called after throwing an instance of 'Rcpp::internal::LongjumpException'
```

So it seems like the `LongjumpException` is not being caught for some reason. Other exception-throwing tests do still work fine, so it seems to be something unique to R + the `longjmp` exception handling.

I wonder if this is a more general R issue -- ie, I wonder if we could reproduce this independently of Rcpp, with R-devel?

eddelbuettel:
Note that *all my troubles are with R-release* which is an even more annoying kettle of fish.

lionel-:
I have conditioned out the tests and tested it on win-builder and it passes: https://win-builder.r-project.org/WKI449t96Y6B/00check.log

I think the freeze on r-hub is due to something else.

eddelbuettel:
I just pushed the branch that I hoped would save me but doesn't.

eddelbuettel:
@Lionel good thinking.  I also thought I should try win-builder once more.

lionel-:
(I have still no idea why the condition isn't caught on win-builder)

eddelbuettel:
@lionel:  So to be plain, you left all code as is but just skipped the test?   That is what I started from ...

lionel-:
I'm going to send the PR as I've restructured the tests a bit and added one more.

eddelbuettel:
Please don't send anything yet. 

I am extremely frustrated with the state of thing right now, and the last thing I want is another PR.

eddelbuettel:
Let's dial down and analyse what broke, or if you know it, share it.

Then let's establish a new baseline where this works everywhere.  We will call that 0.12.15.

PRs after that.

lionel-:
Sorry I sent it immediately but feel free to close it. The steps to fix the win-builder issue are:

- Move the stack unwind tests from `misc.cpp` to another file. The `#define RCPP_PROTECTED_EVAL` should move to that file as well.

- Disable the unit tests in the R file.

eddelbuettel:
Crikey, it looks like I wasted two or three complete hours chasing my own tail under the assumption R Hub was sane ... when it looks like R Hub's windows box was out sync.

I did submit a new build with this PR.  Maybe it will indeed fit the bill just fine.   One of the two at win-builder came back "no vignettes/ directory" so that may be spurious, the other concluded.

kevinushey:
I just put together a sample package (independent of Rcpp) that attempts to use the new R longjmp / unwind mechanism. It reliably crashes on my Windows VM.

https://github.com/kevinushey/longjmp

If you want to test it out:

```R
devtools::install_github("kevinushey/longjmp")
longjmp::longjmp()
```

Works fine on macOS, works fine on 64bit R on Windows, crashes on 32bit R on Windows as the exception is not caught. @lionel- does this seem like a minimal enough example that we might want to let some members of R Core know?

kevinushey:
I suspect the issue on Windows may come down to the use of dwarf exceptions with the 32bit Windows toolchain versus SEH on the 64bit Windows toolchain.

We may want to just also guard the usage of Rcpp's fast eval with something like:

```
#ifdef _WIN32
# ifdef _WIN64
// use fast eval
# else
// use slow eval
#endif
#endif
```

eddelbuettel:
I have the feeling we had seen 32 and 64 bit failures.  But maybe not -- all I triggered was via _joint_ build on, respectively, win-builder or r-hub.

krlmlr:
I sent a few instances of b52f690 to win-builder, will post results.

eddelbuettel:
I am currently doing that too.  Sucketh big time because each iteration is 30+ minutes.

The _worst_ is that I have two complete sets of what is now HEAD which failed once and passed once.  The same tarball ...

eddelbuettel:
More current thread also over there at #801.

krlmlr:
Not if you send with different versions ;-) And `devtools::build_win()` is a big time-saver for me.

eddelbuettel:
As I said over there, the *same* tarball got *both* a fail and a pass at two different points in time.

What does `build_win()` do?  I have a two-line shell script (well, plus arg checking and help) that uses `ncftp` to copy a tarball in batch to both instances at win-builder.

krlmlr:
`build_win()` probably does the same as your shell script.

I've sent five tarballs that are identical apart from the version number. Enough for a majority vote ;-)

eddelbuettel:
Good experimental design :)

kevinushey:
I e-mailed Luke and his response was effectively "ask @lionel-" ... so color me frustrated as well.

I think we should remove the use of `R_UnwindProtect()` altogether for this release until we hear back from @lionel- when he gets back from vacation.

eddelbuettel:
I am about done _editing_ this locally.  Let me test it once or twice here, then push the branch.

I basically left all he added, but `#ifdef`-ed it away.  This should work.  Unless I tripped myself up.

eddelbuettel:
Ok, new branch is [here](https://github.com/RcppCore/Rcpp/tree/feature/condition_unwind_protect) and I would definitely appreciate another few sets of eyes.

"Works here", submitted to win-builder and r-hub.

lionel-:
> Works fine on macOS, works fine on 64bit R on Windows, crashes on 32bit R on Windows as the exception is not caught. @lionel- does this seem like a minimal enough example that we might want to let some members of R Core know?

Interesting. Maybe we should look at this together at the work week?

In any case this doesn't explain the trouble caused by this branch because I'm pretty sure that `R_UnwindProtect()` was not called at all.

Thanks for all your work sorting this out @eddelbuettel.

eddelbuettel:
No worries.  Your patch meant to do the right thing, and we'll get there. I just had to pay extra today and yesterday because I don't usually run CI on Windows, and had not submitted "occassionally" as I usually do as R-devel changed to much of late.  Lesson learned.

lionel-:
FWIW I have sent Rcpp 0.12.14 to win-builder and got a timeout while rebuilding the vignettes: https://win-builder.r-project.org/112Ax81s7Dib/00check.log

Also R-hub times out in the tests with that same version. This seems to rule out this PR as the cause of undeterministic failures.

eddelbuettel:
No, that is precisely the issue.  One of the vignettes *runs all the unit tests* and the fact that it doesn't come back absolutely *is* related to this issue.   If you feel like it, try 0.12.13, or try what it is now in master and waiting at CRAN as 0.12.15.

eddelbuettel:
If you care, this is what I got this morning before submitting this to CRAN as 0.12.15:

#### r-devel

```
From: Uwe.Ligges@R-Project.org
To: edd@debian.org
Cc: Uwe.Ligges@R-Project.org
Subject: winbuilder: Package Rcpp_0.12.15.tar.gz has been checked and built
Date: Tue, 16 Jan 2018 13:35:30 +0100

Dear package maintainer,
 
this notification has been generated automatically.
Your package Rcpp_0.12.15.tar.gz has been built (if working) and checked for Windows.
Please check the log files and (if working) the binary package at:
https://win-builder.r-project.org/w4r8LSs3H6eY
The files will be removed after roughly 72 hours.
Installation time in seconds: 89
Check time in seconds: 717
Status: 4 NOTEs
R Under development (unstable) (2018-01-15 r74124)
 
All the best,
Uwe Ligges
(CRAN maintainer of binary packages for Windows)
```

#### r-release

```

From: Uwe.Ligges@R-Project.org
To: edd@debian.org
Cc: Uwe.Ligges@R-Project.org
Subject: winbuilder: Package Rcpp_0.12.15.tar.gz has been checked and built
Date: Tue, 16 Jan 2018 13:37:07 +0100

Dear package maintainer,
 
this notification has been generated automatically.
Your package Rcpp_0.12.15.tar.gz has been built (if working) and checked for Windows.
Please check the log files and (if working) the binary package at:
https://win-builder.r-project.org/zGdIH49Vy33X
The files will be removed after roughly 72 hours.
Installation time in seconds: 85
Check time in seconds: 720
Status: 5 NOTEs
R version 3.4.3 (2017-11-30)
 
All the best,
Uwe Ligges
(CRAN maintainer of binary packages for Windows)
```

lionel-:
> One of the vignettes runs all the unit tests and the fact that it doesn't come back absolutely is related to this issue. 

But the version I sent was anterior to this PR. Maybe the 4th version component causes too many tests to be run while building the vignette?

eddelbuettel:
Here are my logs from yesterday.  Note that r-hub essentially always breaks.  The key is that 
- before your PR winbuilder passes on both
- after your PR r-rel passes, r-dev fails
- this was a bit a dump of conciousness, I did not run all in order or write entries in order
- we did conclude that win-builder also had some flakyness (but cf @krlmlr getting "five for five")
- we all concluded that the PR was not good at this time (cf @kevinushey comments)

```
- 0.12.14.1 just after 0.12.14
    - now in ~/git/rcpp-is-fucked-up/000_release_0.12.14/
    - sent to win-builder @ 10:32
    - 40-ish mins later PASS on r-devel and r-release
        
- 0.12.14.3 was pivoted just before Lionel's pr789
    - now checked out in ~/git/rcpp-is-fucked-up/010_before_789_after_785/Rcpp
    - PASSed win builder on 01/14
    - times out at r-hub, indicating r-hub may not be quite right

- 0.12.14.3.1 (and 0.12.14.35) was just after Lionel's pr789 
    - branch 'local/before790' at 1630a55b
    - added skip of Lionel's test
    - http://builder.r-hub.io/status/Rcpp_0.12.14.3.1.tar.gz-c3812d0bcbcd430c8f691c6abb3a3191
    - times out at rhub
    - built again at 1630a55b in ~/git/rcpp-is-fucked-up/020_after_789_before_790
    - calling this 0.12.14.35
    - sent to win-builder @ 11:28
    - 30+ mins later PASS r-release
    - 60+ mins later FAIL r-devel (note that we had NOT remove his tests)

- 0.12.14.4 was just after #790 by Kevin
    - added skip of Lionel's test
    - http://builder.r-hub.io/status/Rcpp_0.12.14.4.tar.gz-614f7dd1ed444823bb5aedf5404d44e9
    - times out at rhub 
    
- 0.12.14.7 was another more minimalistic attempt "from the current end" without unwinw
    - times out at still at rhub
        
- 0.12.14.7.1 was `git checkout -b local/beforePR797`
    - times out at rhub
      
- 0.12.14.8 ("current" as of 01/14 and 01/15)
    - failed win-builder (r-rel,r-dev) on 01/14
    - PASSed win-builder (r-rel,r-dev) on 01/15

- 0.12.14.8 (with new #ifdef)
    - sent to win-builder at 14:40 or so
    - on r-hub three WARNING (normal) and one ERROR (with UBSAN, no actual error, also normal)
```

lionel-:
> we all concluded that the PR was not good at this time

Agreed. But we still need to figure out whether the PR really is the cause of the nondeterministic failures. Aren't you concerned that the failure also occurs with the previous CRAN release?

eddelbuettel:
> Aren't you concerned that the failure also occurs with the previous CRAN release?

Well, it doesn't.  First entry:

```
- 0.12.14.1 just after 0.12.14
    - now in ~/git/rcpp-is-fucked-up/000_release_0.12.14/
    - sent to win-builder @ 10:32
    - 40-ish mins later PASS on r-devel and r-release
```

PASS on r-devel and r-release is where we started.

I do not think that you and I are doing right now is productive use of my time, or your vacation.  _We all_ has determined that the issue was with #789.  You can fix, or throw your arms in the air.  But please do not make me re-visit every aspect now.   We all missed that this comes up a lot on Windows, but also see @kevinushey above (or maybe in the #801 thread) stating that _it also took his RStudio on OS X down_ though not each time.

Maybe we should just wait til this stabilizes in R-devel.

PR #793, opened 2017-12-21
Add a 'SystemRequirements' entry to DESCRIPTION

kenahoo:
Although it should be obvious to anyone who deliberately installs `Rcpp` that it requires a C++ compiler, I think it's helpful to call it out explicitly in the DESCRIPTION file.  This helps automated systems keep track of which packages have external dependencies, even if it might require a user to figure out how to actually satisfy it.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,6 +17,7 @@ Description: The 'Rcpp' package provides R functions as well as C++ classes whic
 Depends: R (>= 3.0.0)
 Imports: methods, utils
 Suggests: RUnit, inline, rbenchmark, knitr, rmarkdown, pinp, pkgKitten (>= 0.1.2)
+SystemRequirements: A C++ compiler and any libraries it needs for proper operation
 VignetteBuilder: knitr
 URL: http://www.rcpp.org, http://dirk.eddelbuettel.com/code/rcpp.html, https://github.com/RcppCore/Rcpp
 License: GPL (>= 2)


eddelbuettel:
Sorry.  What goes in here is driven by CRAN, `R Core` and not `R CMD check`.

If you think you need to educate the world about Rcpp, write a blog post.

If you think you need to file a pull request, read [Contributing](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) first and discuss with us.

kenahoo:
I don't understand what you mean.  I'm only "driven by" an attempt to comply with the recommendations in https://cran.r-project.org/doc/manuals/r-release/R-exts.html about where external dependencies should be declared.  Not sure what you mean by the `R CMD check` reference.

I also have no idea what you mean about educating the world.  Or CRAN or R Core.

I made this suggestion because I created an Alpine Linux dockerfile for R with `devtools`, `Rcpp`, `roxygen2`, and `testthat` installed, and all their dependencies.  In doing so, there were many dozens of packages (system & R) installed, but exactly 3 undeclared dependencies that I had to puzzle through manually:

 * `curl-dev` system package depends on `musl-dev` system package
 * `xml2` R package depends on `bash` system package
 * `Rcpp` R package depends on `g++` system package or another C++ compiler

Just thought you might want to know, I guess.

eddelbuettel:
Alpine is a minimal reduced C library that does not provide what libc provides.  In any event, this has nothing to do with Rcpp -- but feel free to bring this to `r-devel`.

PR #796, opened 2018-01-06
Update api.cpp for ANDROID

kendonB:
People are compiling Rcpp on Android so they can work on the bus (and the toilet):

https://github.com/termux/termux-packages/issues/250#issuecomment-302719003

This patch should help with installing Rcpp on that platform. Credit @its-pointless:
https://github.com/its-pointless/gcc_termux/blob/master/R-patches/Rcpp/api.cpp.patch

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -33,7 +33,7 @@ using namespace Rcpp;
 #endif
 
 #if defined(__GNUC__) || defined(__clang__)
-    #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX) || defined(__MUSL__) || defined(__HAIKU__)
+    #if defined(_WIN32) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__CYGWIN__) || defined(__sun) || defined(_AIX) || defined(__MUSL__) || defined(__HAIKU__) || defined(__ANDROID__)
         // do nothing
     #else
         #include <execinfo.h>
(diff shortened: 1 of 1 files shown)


eddelbuettel:
I wondered about that. I followed that end-of-December post and also installed the basic Debian image on my phone, but I got disenchanted when I could not upgrade through to Debian testing.  However, on that platform everything already seems to work with the plain source.  One of the other branches of the `#define` must be covering this.

I'll just wait for the Travis check but I think I can merge this as is as I see no possible harm from the PR and potential upside -- so thumbs up and thanks!

eddelbuettel:
I actually have `termux` installed too.  Guess I might not have known about its `apt` capabilities. 

Are you running R there? When I do `pkg search r-base` nothing comes back as found.

kendonB:
`pkg search r-cran`. Install info is scattered through `https://github.com/termux/termux-packages/issues/250` and `https://github.com/its-pointless/gcc_termux`

eddelbuettel:
Still nuttin'. Using `https://termux.net stable InRelease`.  Do I need to switch?

Also, if you have an email address you'd like me to use in ChangeLog please send (or DM) it (or reply here).  Else the (useless) default from GH goes in...

Gottan run an errand now...

kendonB:
You likely have to add the its-pointless repo: see the readme here: https://github.com/its-pointless/gcc_termux

eddelbuettel:
I guess I am even fiurther behind as I have no `/etc/apt/`. Hm.  Will read up -- thanks for the pointers!

kendonB:
Mine is in ``~/../usr/etc/apt``

eddelbuettel:
Oh wow that worked and I even see r-cran-rquantlib, the fattest of all fat packages ...

eddelbuettel:
Yay. R prompt. Rcpp installed.  [ Annoying that he renames the package, and also mungles the depends on `libicu`. ]

Many thanks for this.  Very cool.

PR #798, opened 2018-01-09
close #797: improve support for NVCC

### LINKED ISSUE #797: Support NVCC better?
Cf email discussion on rcpp-devel in thread titled "NVCC compatibility" started by @cdeterman.

One issue, apparently found first by @jtilly and noted [here in his RcppThrust README](https://github.com/jtilly/RcppThrust) is that [this line](https://github.com/RcppCore/Rcpp/blob/4af2b9acd40e198f657f67e028cbcceef3b68c1d/inst/include/Rcpp/vector/proxy.h#L98) need to drop `const`.   Can we just add a non-const version too?

(Typo fixed. Added New Year's Resolution to only copy addresses and never ever type free hand, cf `fortunes::fortune("dirk can type")`)

Enchufa2:
For the record, find attached a Dockerfile which verifies that nvcc no longer complains (apart from the warnings).

[Dockerfile](https://github.com/RcppCore/Rcpp/files/1616937/Dockerfile.txt)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2018-01-09  Iñaki Ucar  <i.ucar86@gmail.com>
+
+        * inst/include/Rcpp/vector/proxy.h: Improve support for NVCC, the CUDA
+        compiler (pull request #798, fixed issue #797)
+
 2018-01-06  Kendon Bell  <kmb56@berkeley.edu>
 
         * src/api.cpp: Allow for compilation on ANDROID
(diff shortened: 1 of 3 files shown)


eddelbuettel:
Thanks, launching a prrd run now for some revdep checks,

eddelbuettel:
Reverse depends check did not throw up any new issues, so I'll merge this.

Thanks a lot for the very prompt and clean PR!

Issue #804, opened 2018-01-18
Calling vector constructors with R_xlen_t or size_t values > 2³¹ fails in 64-bit R on Windows


krlmlr:
- The value of `(size_t)0x81000000` look correct when cast to `double`
- I can't create a `RawVector` of that size, presumably because it's coerced to a 32-bit integer
- Note the difference between `sizeof.long` and `sizeof.pointer`

Related: #459, #460, CC @kendonB.

``` r
.Machine$sizeof.long
#> [1] 4
.Machine$sizeof.pointer
#> [1] 8
f1 <- Rcpp::cppFunction("double f1() { R_xlen_t x = 0x81000000; return x; }")
f2 <- Rcpp::cppFunction("double f1() { size_t x = 0x81000000; return x; }")
f3 <- Rcpp::cppFunction("RawVector f3() { R_xlen_t x = 0x81000000; return RawVector(x); }")
f4 <- Rcpp::cppFunction("RawVector f4() { size_t x = 0x81000000; return RawVector(x); }")
f1()
#> [1] 2164260864
f2()
#> [1] 2164260864
f3()
#> Error in f3(): negative length vectors are not allowed
f4()
#> Error in f4(): negative length vectors are not allowed
```

<details><summary>Session info</summary>

``` r
devtools::session_info()
#> Session info -------------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.4.3 (2017-11-30)
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United States.1252  
#>  tz       Europe/Berlin               
#>  date     2018-01-18
#> Packages -----------------------------------------------------------------
#>  package   * version date       source        
#>  backports   1.1.2   2017-12-13 CRAN (R 3.4.3)
#>  base      * 3.4.3   2017-12-06 local         
#>  compiler    3.4.3   2017-12-06 local         
#>  datasets  * 3.4.3   2017-12-06 local         
#>  devtools    1.13.4  2017-11-09 CRAN (R 3.4.2)
#>  digest      0.6.13  2017-12-14 CRAN (R 3.4.3)
#>  evaluate    0.10.1  2017-06-24 CRAN (R 3.4.0)
#>  graphics  * 3.4.3   2017-12-06 local         
#>  grDevices * 3.4.3   2017-12-06 local         
#>  htmltools   0.3.6   2017-04-28 CRAN (R 3.4.0)
[...]

PR #811, opened 2018-02-07
allow long long on 64bit windows

kevinushey:
Fixes https://github.com/RcppCore/Rcpp/issues/804. Because the compiler toolchain on Windows provides long long regardless of whether C++98 or a newer standard is requested, this usage should be permissible on CRAN.

In addition, because `R_xlen_t` and `std::size_t` are actually defines for `long long` and `unsigned long long` on 64bit Windows, those constructors now work 'for free'.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2018-02-07  Kevin Ushey  <kevinushey@gmail.com>
+
+        * inst/include/Rcpp/longlong.h: Allow long long on 64bit Windows
+
 2018-01-21  Dirk Eddelbuettel  <edd@debian.org>
 
         * README.md: Added new section 'Support'
--- a/inst/include/Rcpp/longlong.h
+++ b/inst/include/Rcpp/longlong.h
@@ -38,4 +38,11 @@ __extension__ typedef unsigned long long int rcpp_ulong_long_type;
 #endif
 #endif
 
+// The Rtools toolchain provides long long on 64bit Windows
+#ifdef _WIN64
+typedef long long rcpp_long_long_type;
+typedef unsigned long long rcpp_ulong_long_type;
+#define RCPP_HAS_LONG_LONG_TYPES
+#endif
+
 #endif


krlmlr:
Thanks @kevinushey for looking into it!

eddelbuettel:
@kevinushey on win-builder r-devel I get:

```
* checking whether package 'Rcpp' can be installed ... WARNING
Found the following significant warnings:
  ../inst/include/Rcpp/longlong.h:43:14: warning: ISO C++ 1998 does not support 'long long' [-Wlong-long]
  ../inst/include/Rcpp/longlong.h:44:23: warning: ISO C++ 1998 does not support 'long long' [-Wlong-long]
See 'd:/RCompile/CRANguest/R-devel/Rcpp.Rcheck/00install.out' for details.
* checking installed package size ... 
```

Do we need additional C++11 protection around these?

kevinushey:
*grumbles loudly*

Let me think about whether there's an alternate mechanism by which we can get a 64bit integer type on Windows. I'd love for this to work even without C++11 but worst case scenario we can just add that in.

eddelbuettel:
I am 100% certain that we fought these dragons _many_ times before.  And at some point I just gave it a major *fsck this* as you cannot argue away the sillyness of the is C++ 1998 (hey, twenty years) requirement for an entirely irrelevant standard.

But long and short, unless users opt into C++11 (as eg my RcppBDT for Boost date_time did maybe six years ago) to get `long long` they simply will not get them on 'doze.  _C'est la vie_

I would be *delighted* to be proven wrong but don't expect it.

eddelbuettel:
Ok, win-builder came back and I suggest that we go fix this fix:

```diff
modified   inst/include/Rcpp/longlong.h
@@ -40,9 +40,11 @@ __extension__ typedef unsigned long long int rcpp_ulong_long_type;
 
 // The Rtools toolchain provides long long on 64bit Windows
 #ifdef _WIN64
+#ifdef RCPP_USING_CXX0X_OR_LATER
 typedef long long rcpp_long_long_type;
 typedef unsigned long long rcpp_ulong_long_type;
 #define RCPP_HAS_LONG_LONG_TYPES
 #endif
+#endif
```

Any objections?  I will be less 'global' that what @krlmlr had hoped but such is life when our good friends insist on C++98 for reasons way beyond our comprehension.

kevinushey:
I'm going to take another look at this so may follow up with another PR.

kevinushey:
(but if I fail I think that will be the right change)

krlmlr:
Can we just set `CXX_STD=CXX11` in `Makevars.win`? Worked for me in RSQLite and friends, I'm using `int64_t` there without problems.

eddelbuettel:
@krlmlr  Yes the 'use C++11' trick has worked (and been used) for many many years.

@kevinushey ok

krlmlr:
Do I understand correctly that we can't just add in `CXX_STD=CXX11` to `Makevars.win` in *Rcpp*? Or is it about packages having to opt in to use C++11?

Can we add some protection to avoid the underlying issue to occur in the future? I'd like to see an error or a warning if a vector constructor is called with an `R_xlen_t` and this is silently coerced to a 32-bit integer.

eddelbuettel:
We didn't do this _for Rcpp_ as we would then force down a newer compiler.  @jjallaire usually reminds us of the numerous old RHEL instances still trying to build it.  

We impose a new(er) compiler for RcppArmadillo and get occassional issues back, see its issue tickets.

As for `R_xlen_t` and 32-bit:  that is a) an R issue I believe them to handle well enough and b) on the other hand 32 bit machines are on their way out too...   The main issue, really, is Windows where we now have g++ 4,9.3 which while far from ideal is a _huge_ step up from the 4.6.2 dinosaur before it.

krlmlr:
I'm only suggesting to enable C++11 on Windows. Would that work?

eddelbuettel:
Maybe. But I don't like the idea of splitting standards.  And 4.9.3 is not _really_ C++11 compliant, it just pretends to be.  (For RcppCCTZ we ported something from the C++ Std Lib back.) 

And the patch by @kevinushey is functionally equivalent.

PR #813, opened 2018-02-14
Generate .Call(...) instead of .Primitive(.Call)(...) for cppFunction

### LINKED ISSUE #795: cppFuncton() emits .Primitive(".Call") instead of just .Call
See example below. This is a problem for joint profiling of R and native code, because I'm looking for `.Call` entries on the call stack, which are absent if the primitive is invoked directly (https://github.com/r-prof/gprofiler/issues/8). Would you accept a PR that swtiches to `.Call()`?

``` r
print(Rcpp::cppFunction("int x() { return 1; }"))
#> function () 
#> .Primitive(".Call")(<pointer: 0x7fd9fb532060>)
```

> Created on 2018-01-03 by the [reprex package](http://reprex.tidyverse.org) (v0.1.1.9000).

krlmlr:
Closes #795.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2018-02-14  Kirill Müller  <krlmlr@mailbox.org>
+
+        * R/Attributes.R: Code compiled with cppFunction() uses .Call instead
+        of .Primitive(".Call"), #795
+
 2018-02-07  Kevin Ushey  <kevinushey@gmail.com>
 
         * inst/include/Rcpp/longlong.h: Allow long long on 64bit Windows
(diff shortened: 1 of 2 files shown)


eddelbuettel:
This may become my favourite pull request ever:

```
R> library(Rcpp)
R> cppFunction("double plusTwo(double x) { return x+2; }")
R> plusTwo
function (x) 
.Call(<pointer: 0x7f692cdf0cb0>, x)
R> plusTwo(40)
[1] 42
R> 
```

We _all_ just looked at each other assuming the other knew why on earth we used `.Primitive()` in here too.

eddelbuettel:
Well okay, maybe not as famous as the one by @jjallaire that got us Attributes in the first place ;-)

PR #821, opened 2018-02-18
Fix typo

Marlin-Na:
import -> important


eddelbuettel:
Thanks for catching that.

Issue #823, opened 2018-02-23
RNG behaves differently in nested calls to Rcpp functions


wch:
The global [`RNGScopeCounter`](https://github.com/RcppCore/Rcpp/blob/0069f0b0/src/api.cpp#L68), causes `GetRNGstate()` and `PutRNGstate()` to be called only when entering/exiting the lowest Rcpp function on the call stack. 

This can cause problems when  one Rcpp function calls another one, and the latter one uses the random number generator.

In the example below, the behavior of `f` is different when called directly from the console vs. called from another Rcpp function.

```R
library(Rcpp)

cppFunction('
  double rng() {
    return R::runif(1,10);
  }
')

cppFunction('
  Rcpp::RObject invoke(Rcpp::Function f) {
    return f();
  }
')

f <- function() {
  set.seed(100)
  a <- runif(1)
  rng()
  
  set.seed(100)
  rng()
  b <- runif(1)

  identical(a, b)
}

f()  # FALSE

invoke(f) # TRUE
```

cc: @jcheng5
From rstudio/shiny#1953.

eddelbuettel:
As I discussed over there, I consider the example invalid and construed because of `Rcpp::Function()`.

eddelbuettel:
I guess our basic guarantee is this one:

```r
R> set.seed(100)
R> c(rng(), runif(1))
[1] 0.307766 0.257673
R> set.seed(100)
R> c(runif(1), rng())
[1] 0.307766 0.257673
R> 
```

Using your `rng()` function (corrected using `R::runif(0,1)` to correspond to the R call -- ), we see that once a set is seed, it does matter whether your query the RNG via R or C.  That is a nice property.

It seems like piercing this guarantee with Rcpp::Function() to call back to R destroy it.  Oh well, so let's not do that then in apps relying on identical RNG streams.   One could always call out to other RNGs.

Slightly mod'ed example below:

```c++
#include <Rcpp.h>

// [[Rcpp::export]]
double rng() {
  return R::runif(0, 1);
}

// [[Rcpp::export]]
Rcpp::RObject invoke(Rcpp::Function f) {
  Rcpp::RObject z = f();
  return z;
}

/*** R
f <- function() {
  set.seed(100)
  a1 <- runif(1)
  b1 <- rng()

  set.seed(100)
  a2 <- rng()
  b2 <- runif(1)

  identical(c(a1, b1), c(a2, b2))
}

f()  

invoke(f) # not TRUE
*/
```

[...]

PR #825, opened 2018-02-27
unconditionally get/put RNG state

kevinushey:
Resolves https://github.com/RcppCore/Rcpp/issues/823. With this PR, we ensure that Rcpp functions called by other Rcpp functions synchronize the RNG state.

Although @wch's example showed this through the use of `Rcpp::Function`, in theory this could occur if an Rcpp function were to e.g. call `source()` or `eval()` or another API that happened to call another Rcpp function.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2018-02-26  Kevin Ushey  <kevinushey@gmail.com>
+
+        * src/api.cpp: Always set / put RNG state when calling Rcpp function
+
+
 2018-02-25  Dirk Eddelbuettel  <edd@debian.org>
 
         * vignettes/Rcpp.bib: Updated
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -7,6 +7,11 @@
   \itemize{
     \item Changes in Rcpp API:
     \itemize{
+      \item Rcpp now sets and puts the RNG state upon each entry to an Rcpp
+      function, ensuring that nested invocations of Rcpp functions manage the
+      RNG state as expected
+    }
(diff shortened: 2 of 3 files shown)


kevinushey:
I wanted to leave the counter there just because it's part of the API of Rcpp (in particular, they're registered routines) and I want to avoid touching anything around that interface just out of an abundance of caution.

eddelbuettel:
That seems ... weird.  The entire block is https://github.com/RcppCore/Rcpp/blob/master/src/api.cpp#L66-L83 and the counter is not part of the interface.  We could simply make it `const`.  I just don't see why you want to increment/decrement it.  AFAIK no other code touches it.

And FWIW the silversearcher agrees:

```sh
edd@rob:~/git/rcpp(master)$ ag  RNGScopeCounter
src/api.cpp
68:            unsigned long RNGScopeCounter = 0;
73:            if (RNGScopeCounter == 0) GetRNGstate();
74:            RNGScopeCounter++;
75:            return RNGScopeCounter;
80:            RNGScopeCounter--;
81:            if (RNGScopeCounter == 0) PutRNGstate();
82:            return RNGScopeCounter;
edd@rob:~/git/rcpp(master)$ 
```

kevinushey:
Sorry; I meant that `enterRNGScope()` and `exitRNGScope()` are registered routines (just because we wanted to make sure that all packages using Rcpp were looking at the same counter, I presume).

I'll get rid of the counter.

eddelbuettel:
And I presume you and @wch tested this and ensured that it addressed the corner case seen with shiny?

PR #826, opened 2018-02-28
comment-out R::pythag

eddelbuettel:
Per email from BDR

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,11 @@
+2018-02-28  Dirk Eddelbuettel  <edd@debian.org>
+
+        * inst/include/Rcpp/Rmath.h (R): Rf_pythag has been remove in R 2.14.0
+        so comment-out the R::pythag wrapper (per request of Brian Ripley)
+
 2018-02-26  Kevin Ushey  <kevinushey@gmail.com>
 
         * src/api.cpp: Always set / put RNG state when calling Rcpp function
-
 
 2018-02-25  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2018-03-01  Dirk Eddelbuettel  <edd@debian.org>
+
+        * inst/include/Rcpp/sugar/functions/complex.h (Rcpp): Remove RCPP_HYPOT
+        macro and use ::hypot() throught as it is provided with C99
+
+        * inst/include/Rcpp/sugar/undoRmath.h: Also uncomment pythag here
+
 2018-02-28  Dirk Eddelbuettel  <edd@debian.org>
 
         * inst/include/Rcpp/Rmath.h (R): Rf_pythag has been remove in R 2.14.0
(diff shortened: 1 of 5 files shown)


eddelbuettel:
No comments by anyone?  Should be harmless -- may just merge in a few ...

kevinushey:
We might (?) need a few more changes on our side:

```
inst/include/Rcpp/sugar/functions/complex.h
28:# define RCPP_HYPOT ::Rf_pythag

inst/include/Rcpp/sugar/undoRmath.h
110:#undef pythag

inst/include/Rcpp/Rmath.h
222:    inline double pythag(double a, double b)        { return ::Rf_pythag(a, b); }
```

I think the bit in `undoRmath.h` can stay, but we might want to tweak what we have in `complex.h`:

```
kevin@cdrv:~/r/pkg/Rcpp [rng-state]
$ ag -Q "RCPP_HYPOT"
inst/include/Rcpp/sugar/functions/complex.h
26:# define RCPP_HYPOT ::hypot
28:# define RCPP_HYPOT ::Rf_pythag
83:             y.r = ::log( RCPP_HYPOT( x.r, x.i ) );
90:         if( (mag = RCPP_HYPOT(z.r, z.i)) == 0.0)
144:    t1 = 0.5 * RCPP_HYPOT(x + 1, y);
145:    t2 = 0.5 * RCPP_HYPOT(x - 1, y);
```

If I understand correctly, R removed that function because C99 now defines `::hypot()` and so R doesn't need to provide its own definition. (If I understand correctly, `Rf_pythag()` was R's own definition of `::hypot()`, for platforms that didn't provide it back in the day)

kevinushey:
That said, if everything compiles and checks okay we can probably leave it as is?

eddelbuettel:
There is indeed a bit more.  You first catch is good:  RCPP_HYPOT is conditionally defined with a now-dead fallback.  The `undef` is harmless.  The `inline` we can just comment out as well.

The use of `RCPP_HYPOT` may need a second look as RCPP_HYPOT now has no fallback. Question is then ... would `#ifdef HAVE_HYPOT` ever not be true?  And if so, shall we just emit a warning there?

`man 3 hypot` points to C99, POSIX.1-2001, POSIX.1-2008.  Good enough?

Also:

```sh
edd@rob:~$ grep HAVE_HYPOT ~/svn/r-devel/config.status 
S["RMATH_HAVE_HYPOT"]="# define HAVE_HYPOT 1"
D["HAVE_HYPOT"]=" 1"
edd@rob:~$ 
```

eddelbuettel:
The cleanest may just be to rewrite `complex.h` and do away with `RCPP_HYPOT` in favour of `::hypot`. Let me pile that on later today.

kevinushey:
I think `HAVE_HYPOT` is effectively always true since (as you pointed out) it's defined by the C99 standard and I think R requires C99 or greater during compilation of R + extensions.

eddelbuettel:
@kevinushey That was an excellent catch.  I maded some changes, when you have moment, could you peruse?

kevinushey:
LGTM -- thanks for taking care of this!

PR #828, opened 2018-03-06
Optionally suppress call info in Rcpp::stop and Rcpp::warning

michaelweylandt:
Add `bool include_call` arguments to `Rcpp::stop` and `Rcpp::warning` to
control inclusion of the call stack in Rcpp produced errors and warnings. 

See discussion at http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2018-March/009906.html (and thread)

Currently, the following produces a sub-par error message (`Error in (function (x)  : My error message.`): 

```{r}
library(Rcpp)

sourceCpp(code='

#include "Rcpp.h"

// [[Rcpp::export]]
Rcpp::NumericVector internal_function_name(Rcpp::NumericVector x){
    Rcpp::stop("My error message.");
    return x + 2;
}')

add2 <- function(x){
    if(!is.numeric(x)){
        x <- as.numeric(x)
    }

    do.call(internal_function_name, list(x))
}

add2(1:5)
```

With the attached PR, this will give a cleaner error message (`Error: My error message.`): : 

```{r}
library(Rcpp)

sourceCpp(code='

#include "Rcpp.h"

// [[Rcpp::export]]
Rcpp::NumericVector internal_function_name(Rcpp::NumericVector x){
    Rcpp::stop("My error message.", false); // include_call = false
    return x + 2;
}')

add2 <- function(x){
    if(!is.numeric(x)){
        x <- as.numeric(x)
    }

    do.call(internal_function_name, list(x))
}

add2(1:5)
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2018-03-05  Michael Weylandt  <michael.weylandt@gmail.com>
+
+        * inst/include/Rcpp/exceptions/cpp{98,11}/exceptions.h: Add
+        include_call argument to Rcpp::stop and Rcpp::warning to suppress
+        inclusion of call stack in resulting error / warning.
+
 2018-03-01  Dirk Eddelbuettel  <edd@debian.org>
 
         * inst/include/Rcpp/sugar/functions/complex.h (Rcpp): Remove RCPP_HYPOT
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -13,6 +13,10 @@
       \item The \code{R::pythag} wrapper has been commented out; the underlying
       function has been gone from R since 2.14.0, and \code{::hypot()} (part of
       C99) is now used unconditionally for complex numbers.
+      \item \code{Rcpp::stop} and \code{Rcpp::warning} gain a new \code{include_call}
+      argument which functions similarly to the \code{call.} argument to
+      \code{base::stop} and \code{base::warning} at the R level. By default,
+      this is \code{true} for back-compatability.
     }
     \itemize{
       \item The \code{long long} type can now be used on 64-bit Windows (Kevin
--- a/inst/include/Rcpp/exceptions/cpp11/exceptions.h
+++ b/inst/include/Rcpp/exceptions/cpp11/exceptions.h
@@ -43,13 +43,22 @@ class __CLASS__ : public std::exception {                                       
 };
 
 template <typename... Args>
-inline void warning(const char* fmt, Args&&... args ) {
+inline void warning(const char* fmt, Args&&... args) {
     Rf_warning("%s", tfm::format(fmt, std::forward<Args>(args)... ).c_str());
 }
 
 template <typename... Args>
-inline void NORET stop(const char* fmt, Args&&... args) {
-    throw Rcpp::exception( tfm::format(fmt, std::forward<Args>(args)... ).c_str() );
+inline void warning(const char* fmt, Args&&... args, bool include_call = true) {
+    if(include_call){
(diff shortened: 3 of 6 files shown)


eddelbuettel:
We mention in the file [Contributing.md](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) that prior discussion is preferred ... and that may have caught the issue apparent here:  you simply cannot introduce new arguments after a varargs argument (ie what would be `...` in R).  

On my box, with C++11 by default (due to the `g++` version) this blows up immediately; I also suspect it the cause of the Travis error.

So no go as this stands.

michaelweylandt:
Ahh -- I'm not up-to-speed on C++ and variadic templates. 

Thoughts on how to proceed? Some ideas: 

1. Implement in C++98 mode only. 
2. Have two functions: `stop` and `stopNoCall` (same for `warning`)
3. ??
4. Something exceedingly clever with templates and overloading  
5. Don't implement this feature, but document how to suppress call info.

eddelbuettel:
Yes, variadic templates for the C++11 side may be the way to go.  @dcdillon is poking around already.

C++98 only does not work; we need one API.

Ditto for `stop()` and `stopQuiet()` (or whatever name): we'd need to double everything up in C++98. Seems excessive.

For now, just use the `throw` method allowing you to pass an explicit `false`.

michaelweylandt:
The `stop` + `stopQuiet` approach doesn't seem too bad: we can cut down the redundancy of two implementations with a macro, but it is a bit heavy handed and makes the `Rcpp` namespace bigger. 

Here's an approach which I think works for both C++98 and C++11 as well as being back-compatible, but I haven't coded it up or necessarily thought through all the details.  

1. Declare a helper type `include_call` which takes a single `bool` argument to the constructor. 
2. Add a template specialization which puts the `include_call` before the optional formatting arguments. 
3. Include the call unless the `include_call(false)` template specialization is found. 

Code snippets: 

Unchanged: 
- `Rcpp::stop("Error");`
- `Rcpp::stop("Loop terminated on iteration k = %d", k);`
- `Rcpp::warning("My warning");`
- `Rcpp::warning("A very bad thing t = %s", 1);`

New (with suppressed call)
- `Rcpp::stop("Error", Rcpp::include_call(false));`
- `Rcpp::stop("Loop terminated on iteration k = %d", Rcpp::include_call(false), k);`
- `Rcpp::warning("My warning", Rcpp::include_call(false));`
- `Rcpp::warning("A very bad thing t = %s", Rcpp::include_call(false), 1);`

Pros: 
1. The use of the `include_call` type (as opposed to a bool in my first attempt), lets us use specialization without breaking existing code. (Looking at it again, my first PR might break existing examples of the form `Rcpp::stop("Thing %s", false)` but I think it would be caught at compile time)
2. Putting the `include_call` at the beginning lets us keep the trailing variadic template in the C++11 version. 

Cons: 
1. `Rcpp::include_call(false)` is verbose. 

Bike-sheddable: 
1. Should `include_call` be a boolean constructor or should we just have a `dont_include_call` singleton. (`include_call(true)` is the same as skipping `include_call` so it's a bit useless.)
2. The name `include_call` is terrible...

I'm happy to put this PR on hold and move to an issue if you'd prefer. (Alternatively, if RcppCore doesn't see a way forward, I can punt on this entirely and put together a RcppGallery post demonstrating how to achieve the same effect without changes to Rcpp.)

michaelweylandt:
Fair enough. I'm busy the next two days, so probably won't get to the gallery post till Thursday. 

Closing for now.

PR #829, opened 2018-03-07
alternate way to grab 'long long' type

kevinushey:
This PR is another attempt to grab 64-bit integer types in C++98 mode in a way that won't trigger pedantic compiler warnings. Fortunately, seems to work on my Windows box using the latest Rtools toolchain.

### DIFF
--- a/inst/include/Rcpp/longlong.h
+++ b/inst/include/Rcpp/longlong.h
@@ -38,10 +38,18 @@ __extension__ typedef unsigned long long int rcpp_ulong_long_type;
 #endif
 #endif
 
-// The Rtools toolchain provides long long on 64bit Windows
-#ifdef _WIN64
-typedef long long rcpp_long_long_type;
-typedef unsigned long long rcpp_ulong_long_type;
+// The Rtools toolchain provides long long on 64bit Windows,
+// but using those types directly in C++98 mode can trigger
+// compiler warnings:
+//
+//   warning: ISO C++ 1998 does not support 'long long' [-Wlong-long]
+//
+// fortunately, glibc provides these types as aliases for int64_t
+// and uint64_t, which we can use as extensions in C++98 mode
+#if defined(_WIN32) && defined(_GLIBCXX_HAVE_INT64_T) && defined(_GLIBCXX_HAVE_INT64_T_LONG_LONG)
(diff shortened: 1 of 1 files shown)


kevinushey:
For context, the Rtools toolchain has a `c++config.h` file with the contents:

```
/* Define if int64_t is available in <stdint.h>. */
#define _GLIBCXX_HAVE_INT64_T 1

/* Define if int64_t is a long. */
/* #undef _GLIBCXX_HAVE_INT64_T_LONG */

/* Define if int64_t is a long long. */
#define _GLIBCXX_HAVE_INT64_T_LONG_LONG 1
```

which I'm making use of in choosing the `int64_t` and `uint64_t` types as 'aliases' for `long long` and `unsigned long long`.

kevinushey:
This also makes me think there might be a chance that we could sneak in `long long` support on other platforms with C++98. I feel like venturing there will open some old scars but it _sure would sure be nice_ to just have `long long` everywhere.

eddelbuettel:
That is a bit of a "riskier" approach as it touches the old part.   Oh I see you just mod'ed that.

kevinushey:
I think this should be ready of merge. Just to summarize the changes:

1. We explicitly enable `long long` and `unsigned long long` when the compiler reports it's compiling in C++11 (or above) mode. (Previously I think this was caught implicitly based on whether C++0x extensions were enabled, which might change in the future)

2. We also enable `long long` for older `gcc` compilers where `long long` is just an alias is `int64_t` (which is true with the Windows toolchain, currently). Fortunately, glibc reports this to us with the `_GLIBCXX_HAVE_INT64_T_LONG_LONG` define.

I think this effectively lets us use `long long` on 64bit Windows (and other 64bit platforms) in a CRAN-compliant way, even if C++11 support has not been explicitly enabled.

If we find out this doesn't work as expected or we end up with warnings on some other platform we can of course back it out, but thus far things look good on my Windows VM, macOS machine and Ubuntu 16.04 VM (with gcc-5).

kevinushey:
Thanks!

eddelbuettel:
Shall we call it a day and ship this now as Rcpp 0.12.16?

kevinushey:
Sounds good to me!

Issue #839, opened 2018-03-21
Rcpp modules do not support named arguments


JorisGoosen:
Rcpp modules work nicely when you have a cpp class that you want to expose to R.
It is, however, not possible to use named arguments in the same way as one could do in R.

It would be practical to allow this and I have made an implementation of that in my fork. It adds a `method_named_args` to the module-definition functions. It assumes that the user has a class method that receives a single `std::map<std::string, Rcpp::RObject>` and will contain all named arguments with the specified name. Any unnamed arguments will be given a name such as `"__unnamed_arg_1"`.

This means the user will have to manually process the arguments in the same logical manner as R would do, aka use the unnamed arguments in the order they come in on the locations that would be expected but this is a small price to pay I would think.

eddelbuettel:
That sounds like something we could look into.  Could you post here a mock (or real) extension of one of the sample modules that ship with Rcpp?

JorisGoosen:
snippets from RcppModuleWorld that is extended in my fork:

```
class RcppModuleWorld {                   // a simple class with a setter and getter
public:
    RcppModuleWorld() : msg("hello"){}
  
    std::string named_args(Rcpp::map_named_args args) 
    { 
        std::string out = (args.count("hello") > 0 ? Rcpp::as<std::string>(args["hello"]) : "") + " ";
        out += args.count("world") > 0 ?  Rcpp::as<std::string>(args["world"]) : (args.count("1") > 0 ?  Rcpp::as<std::string>(args["1"]) : ""); //first check named argument world and otherwise use first unnamed arg ("1") if it exists
        return out;
    }

private:
    std::string msg;
};

std::string free_named_args(RcppModuleWorld * modworld, Rcpp::map_named_args args) { return modworld->named_args(args); }


RCPP_MODULE(RcppModuleWorld) {
    using namespace Rcpp;

    class_<RcppModuleWorld>( "RcppModuleWorld")
        .default_constructor()
        .method_named_args("named_args",    &RcppModuleWorld::named_args)
        .method_named_args("free_named_args",    &free_named_args)
	;
}
```

This works and i've added this to the unittest as well. I can set up a PR for this.

eddelbuettel:
I maybe dense but could you also add the new usage pattern from the R side, and explain why it is useful?

eddelbuettel:
I see it in the unit test you added.

```r
worldtest <- new(RcppModuleWorld)
[...]

PR #840, opened 2018-03-22
Added support for methods with named arguments. 

JorisGoosen:
The user does need to do some manual processing of these but it should work.

They can be registered as .method_named_args() and should accept a std::map<std::string, Rcpp::RObject>.
It also converts any unnamed arguments to named arguments with simply the integer of the unnamed entry as fieldname (because you cannot have named arguments such as 1 or 2 in R anyway. The user can then process them however seen fit.

See issue https://github.com/RcppCore/Rcpp/issues/839


eddelbuettel:
What systems did you test it on?  It bombs at Travis over 

```
g++  -I/usr/share/R/include -DNDEBUG -I../inst/include/     -fpic  -g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -g  -c Date.cpp -o Date.o
In file included from ../inst/include/Rcpp/Module.h:390:0,
                 from ../inst/include/Rcpp.h:66,
                 from Date.cpp:31:
../inst/include/Rcpp/module/class.h: In member function ‘Rcpp::map_named_args Rcpp::class_<Class>::convert_pairlist_to_named_args(SEXP)’:
../inst/include/Rcpp/module/class.h:279:89: error: ‘to_string’ is not a member of ‘std’
             std::string argName = !Rf_isNull(TAG(it)) ? Rcpp::Symbol(TAG(it)).c_str() : std::to_string(++unnamedArgs); //1-based like R likes
                                                                                         ^
make: *** [Date.o] Error 1
```

`to_string` is a C++11 method. We can, in general, not assume C++11 as we have to support much older systems too.

JorisGoosen:
I tested it on mac osx with clang and used R CMD INSTALL to test it but this did not complain. It did before about "override" being C++11 so I had removed those already.

eddelbuettel:
We also need to update the Rd file:

```
* checking for code/documentation mismatches ... WARNING
S4 class codoc mismatches from documentation object 'C++Class-class':
Slots for class 'C++Class'
  Code: constructors .Data docstring enums fields generator methods
        methods_named module parents pointer typeid
  Docs: constructors .Data docstring enums fields generator methods
        module parents pointer typeid
```

eddelbuettel:
This simpler internal method also seems to work, I looked into this earlier while you did the second commit:

```c++
            std::string argName = !Rf_isNull(TAG(it)) ?
                Rcpp::Symbol(TAG(it)).c_str() :
                Rcpp::internal::coerce_to_string<INTSXP>(++unnamedArgs); //1-based like R likes
```

JorisGoosen:
Aah that does look better (shorter anyway) but if you are fine with the standard streams then Ill just leave that.

eddelbuettel:
Department of bad news here.  I ran full reverse depends and packages 

    DNAtools
    filling
    nauf
    optiSel
    optiSolve
    projpred
    relSim

all seg.fault when trying to load their modules.   This is may be something you can work out, but would be a blocker for the current state of the PR.

JorisGoosen:
Aah yeah, I see. I tried getting filling to run and it seems to be the trouble that nabor uses an Rcpp Module and isn't recompiled yet. It then tries to give the class-info and presumably uses the new code to refer to an old compiled structure?

Recompiling nabor solves the problem for me, Im not sure how backwards binary compatibility would be arranged in this case.

eddelbuettel:
Confirming that 'filling' passes once dependecy 'nabor' is reinstalled.  

And same with 'DNAtools': it passes once dependency 'multicool' is reinstalled.  

But forcing users to reinstall is a bit of burden.

jjallaire:
Anything that changes the ABI needs to a very compelling reason to be merged, as Windows users in particular suffer a lot of pain when the ABI changes (as a result of the fact that R will succeed in package installation even when the underlying package DLL is in use). So we usually reserve anything that changes the ABI for critical changes (or alternatively queue up all ABI changes in a single release).

So I'd so definitely don't merge it.

kevinushey:
I agree as well. I don't want to discount the work that @JorisGoosen has done here, but Rcpp Modules are effectively in maintenance mode and I don't think bringing in this PR is worth the potential disruption in the next Rcpp release.

@JorisGoosen: if you wanted to pursue this further, I'd recommend forking the associated code into its own package and continuing development there.

There have been some efforts at alternate module implementations (e.g. [RcppR6](https://github.com/richfitz/RcppR6)) where code generation rather than template meta-programming is used to power the exposure of C++ objects to R, and IMHO that is a more compelling avenue to pursue for the future.

eddelbuettel:
I think we have come to the conclusion that this cannot go in as is, so I'll close the PR to mark it as such.

PR #845, opened 2018-04-17
Replace back-quotes with straight-quotes in normal text

rstub:
* back-quotes used as single opening quotes by TeX are not necessary since csquotes.sty is used to handle smart quotes
* back-quotes can confuse the markdown parser (c.f. section '2.1 Package Creation')

BTW, the closing quote and the last letter of the phrase are awfully close. It might make sense from a typographic point of view to use \textsl{'...'}. If you are interested, I can add this to this PR.

### DIFF
--- a/vignettes/Rcpp-attributes.Rmd
+++ b/vignettes/Rcpp-attributes.Rmd
@@ -557,7 +557,7 @@ moving code from a standalone \proglang{C++} source file to a package:
 ## Package Creation 
 
 To create a package that is based on \pkg{Rcpp} you should follow the
-guidelines in the `\textsl{Rcpp-package}' vignette. For a new package this
+guidelines in the '\textsl{Rcpp-package}' vignette. For a new package this
 is most conveniently done using  the `Rcpp.package.skeleton` function.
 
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Thanks for catching the remaining backquotes. A leftover from latex source I guess.  I'll have to think about / look at the quotes together with `\textsl`. There is a bit of redundancy there.  Maybe do away with one of them?

rstub:
Using either quote or `\textsl` would make a lot of sense. There are already several places where only `\textsl` is used. For only quotes, I have found two references to `'Rcpp Attributes'` in Rcpp-FAQ.Rmd.

coatless:
Down with `\textsl{}`!

In all reality, I think that is the only _latex_ specific code still embedded in the vignettes. If it can be removed by typesetting using quotes, then that's an ideal way to go.

eddelbuettel:
Could do. I like `\textsl{}` better than `\textit{}` which markdown's `_` gives us. The only thing i care about is the pdf output, though, and to me slanted looks best.

PR #852, opened 2018-05-25
execute embedded R script in specified env

### LINKED ISSUE #851: env argument not used to actually source the R code
If you set the `env` variable in `sourceCpp`, `Rcpp` correctly makes the R functions and modules available in the specified environment. However, the R snippet in the `/*** R` block is not executed in this environment, as I would expect.

## Reproducible example

`test.rcpp`

```cpp
#include <Rcpp.h>
using namespace Rcpp ;

// [[Rcpp::export]]
int answer(){
    return 42 ;
}

/*** R
    # call answer and check you get the right result
    x <- answer()
    x
*/
```

Next, in R:

```R
library(Rcpp)
testEnv <- new.env()
sourceCpp("test.cpp", env = testEnv)
ls(testEnv)
```

Expected output:

```
> ls(testEnv)
[1] "answer" "x"
```

Actual output:

```
> ls(testEnv)
[1] "answer"
```

## Context

I work at DataCamp, where we're building a course on Rcpp together with Romain Francois. Our R execution backend heavily relies on the concept of environments: we run the solution code in a solution environment and the student code in a student environment, so that we can check the equality of objects. Because of the above, it is currently very hard for Rcpp exercise to do this. For full context, you can check out [this issue](https://github.com/datacamp/testwhat/issues/118)

cc @sumedh10

filipsch:
Fixes #851 

As mentioned in the issue, I have no idea about how to include a unit test to verify the new behavior; I am open to suggestions or somebody else adding it. Thanks!

I checked locally and it works:

`test.rcpp`

```cpp
#include <Rcpp.h>
using namespace Rcpp ;

// [[Rcpp::export]]
int answer(){
    return 42 ;
}

/*** R
    # call answer and check you get the right result
    x <- answer()
    x
*/
```

in R:

```R
> library(Rcpp)
> testEnv <- new.env()
> sourceCpp("test.cpp", env = testEnv)

>     # call answer and check you get the right result
>     x <- answer()

>     x
[1] 42
> ls(testEnv)
[1] "answer" "x"  
```

### DIFF
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -196,7 +196,7 @@ sourceCpp <- function(file = "",
     if (embeddedR && (length(context$embeddedR) > 0)) {
         srcConn <- textConnection(context$embeddedR)
         setwd(rWorkingDir) # will be reset by previous on.exit handler
-        source(file=srcConn, echo=TRUE)
+        source(file = srcConn, local = env, echo = TRUE)
     }
 
     # cleanup the cache dir if requested


eddelbuettel:
For unit tests, please have a look at `inst/unitTests/runit.*.R` and `inst/unitTests/cpp/*`.  Not all that complicated---R functions calling C++ code built via Rcpp Attributes.  Just like you did in your reprex.  Or am I missing something here while I barely started the first coffee cup of the day?

filipsch:
@eddelbuettel Thanks for the pointers, I'll see how far I get with those and will report here!

filipsch:
@eddelbuettel Added a test, that fails if the `local = env` portion is not there.

eddelbuettel:
In that case could you actually have it fail and catch the failure?  

We have plenty of other tests catching exceptions and errors; RUnit supports that.

filipsch:
@eddelbuettel in this particular case, I think it is hard to do: It would involve me overriding the `source()` function to ignore the `local` argument, not taking in the `env` parameter as such. Feels like a lot of trickery.

eddelbuettel:
Maybe. Methinks you could mod your example from #851 to provide the "wrong" `answer()` function and check for the wrong result.  Or maybe not.  Have not thought to hard about this.

Took a number of years for someone to come along and declare this a problem :)

filipsch:
I'm not sure if I understand. This fix makes sure that embedded R code is executed in the R environment in which the RCpp-generated functions are made available, instead of the default `globalenv()`.
To check that this errs, I would have to programmatically update the R source of `sourceCpp()` and not take in the `env` argument. I don't see another way. Doing this feels very contrived.
Not sure if there's anymore I can contribute :)

eddelbuettel:
My thought was that you don't _have to_ have the R code of your C++ snippet exectuted via the `sourceCpp()` functionality you so like (and which _is_ great!).  Use `sourceCpp()` to compile a new function in  your new environment -- the whole point of #851 as I understand it -- and then _in the normal RUnit R code execution_ look for it.  It should fail as it should be tucked away in the new environment.  No?

filipsch:
It's actually the other way around. The point of #851 is to source the _R code_ in my new environment. Making available the Rcpp functions works fine already.
I've added a spec accordingly that fails as expected.
Thanks for your pointers!

Issue #861, opened 2018-06-06
segfault: terminate called after throwing an instance of 'Rcpp::eval_error'


mtmorgan:
I'm not really sure where the responsibility for this lies, but
```
suppressPackageStartupMessages({
    library(conflicted)
    library(BiocGenerics)  # not too heavy, exports a tweaked evalq
    library(xml2)
})
read_html("https://bioconductor.org")
```
results in a segfault
```
terminate called after throwing an instance of 'Rcpp::eval_error'
  what():  Evaluation error: evalq found in 2 packages. You must indicate which one you want with ::
 * BiocGenerics::evalq
 * base::evalq.
```
Normally (e.g., typing `evalq` at the command line), 'conflicted' would report the conflict between the two definitions of evalq as an error.

Rcpp seems to run in to trouble at inst/include/Rcpp/api/meat/Rcpp_eval.h:118 and 134
```
    Shield<SEXP> res(internal::Rcpp_eval_impl(call, R_GlobalEnv));
    ...
            throw eval_error(CHAR(STRING_ELT(conditionMessage, 0)));
```
the call stack before the segfault is
```
(gdb) where
#0  Rcpp::Rcpp_eval (expr=0x150b520, env=0x64b9d8)
    at ../inst/include/Rcpp/api/meat/Rcpp_eval.h:98
#1  0x00007fffea3fb5c3 in get_rcpp_cache () at barrier.cpp:100
#2  0x00007fffea3fb6ac in rcpp_get_stack_trace () at barrier.cpp:119
#3  0x00007fffea107834 in rcpp_get_stack_trace ()
    at /home/mtmorgan/R/x86_64-pc-linux-gnu-library/3.5-Bioc-3.8/Rcpp/include/Rcpp/routines.h:125
#4  0x00007fffea108b4d in exception_to_r_condition (ex=...)
    at /home/mtmorgan/R/x86_64-pc-linux-gnu-library/3.5-Bioc-3.8/Rcpp/include/Rcpp/exceptions.h:303
#5  0x00007fffea0f7752 in _xml2_read_connection_ (conSEXP=0x12bd038, 
    chunk_sizeSEXP=0x12bd2d8) at RcppExports.cpp:19
#6  0x00007ffff7816a59 in R_doDotCall (
    ofun=0x7fffea0f7491 <_xml2_read_connection_(SEXP, SEXP)>, nargs=2, 
    cargs=0x7ffffffdec20, call=0x974c18)
[...]

PR #863, opened 2018-06-06
evaluate call in base env (#861)

kevinushey:
This is the simpler version of my prior patch for #861 (as recommended by @lionel-). If this builds cleanly I think it should be good to merge


lionel-:
Thanks! It should also have no adverse effects on tracebacks, unlike the inlining solution.

eddelbuettel:
Thanks.  You're fire today :)  Thanks for all that.

I guess I need to figure _what_ I'll turn on to rev.dep.  I guess I should start with Lionel's PR with the extra flag set.

eddelbuettel:
Passes fine in full rev.dep (one level) so merging (after adjusing by hand here for merge conflicts).

lionel-:
@eddelbuettel There was an issue with the merge conflict resolution, this PR only contains Changelog changes.

eddelbuettel:
Boo.

eddelbuettel:
@lionel- @kevinushey @thirdwing :  As Lionel pointed out here, something went amiss with the PRs. I only ever resolved conflicts (on the website) pertaining to `ChangeLog` so I am not sure how this happens -- but between us we seem to have managed to loose some code.   Any idea how to check what's missing, besides glancing at the PR reports here and master?

kevinushey:
I should be able to fix things up on master.

kevinushey:
https://github.com/RcppCore/Rcpp/commit/2cb0d5e2f30b3f12f6ea5b6093dc08479ccfa717

eddelbuettel:
Maybe I was blind (it was late when I came across @lionel- 's message) but I didn't even see the correct fragment in master.  I guess I may been dense and been on another branch...

Thanks for the fix, @kevinushey.

PR #867, opened 2018-06-10
Switching from Rcpp_eval to Rcpp_fast_eval. close #866

### LINKED ISSUE #866: Internal usage of Rcpp_fast_eval in Rcpp
This has been discussed on the Rcpp mailing list: http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2018-June/010029.html

```
Hi all,

I've followed with interest the development of the new evaluation API.
Now that it's finally merged, I was testing it. Perhaps I'm mistaken,
but shouldn't we expect a performance improvement in code such as the
following?

Rcpp::sourceCpp(code='
  #include <Rcpp.h>
  using namespace Rcpp;

  // [[Rcpp::export]]
  void old_api(Function func, int n) {
    for (int i=0; i<n; i++) func();
  }'
)

Rcpp::sourceCpp(code='
  #define RCPP_PROTECTED_EVAL
  #include <Rcpp.h>
  using namespace Rcpp;

  // [[Rcpp::export]]
  void new_api(Function func, int n) {
    for (int i=0; i<n; i++) func();
  }'
)

func <- function() 1
system.time(old_api(func, 1e5))
system.time(new_api(func, 1e5))

Regards,
Iñaki
```

thirdwing:
@Enchufa2 

This PR ports most of the `Rcpp_eval` to `Rcpp_fast_eval`.

I might need a little more time to double check.

### DIFF
--- a/src/barrier.cpp
+++ b/src/barrier.cpp
@@ -28,8 +28,6 @@
 #include <algorithm>
 #include <Rcpp/protection/Shield.h>
 
-namespace Rcpp { SEXP Rcpp_eval(SEXP, SEXP); }
-
 // [[Rcpp::register]]
 SEXP get_string_elt(SEXP x, R_xlen_t i) {
     return STRING_ELT(x, i);
@@ -97,7 +95,7 @@ SEXP get_rcpp_cache() {
         SEXP getNamespaceSym = Rf_install("getNamespace"); // cannot be gc()'ed  once in symbol table
         Rcpp::Shield<SEXP> RcppString(Rf_mkString("Rcpp"));
         Rcpp::Shield<SEXP> call(Rf_lang2(getNamespaceSym, RcppString));
-        Rcpp::Shield<SEXP> RCPP(Rcpp_eval(call, R_GlobalEnv));
+        Rcpp::Shield<SEXP> RCPP(Rf_eval(call, R_GlobalEnv));
 
         Rcpp_cache = Rf_findVarInFrame(RCPP, Rf_install(".rcpp_cache"));
         Rcpp_cache_know = true;
@@ -139,7 +137,7 @@ SEXP init_Rcpp_cache() {
     SEXP getNamespaceSym = Rf_install("getNamespace"); // cannot be gc()'ed  once in symbol table
     Rcpp::Shield<SEXP> RcppString(Rf_mkString("Rcpp"));
     Rcpp::Shield<SEXP> call(Rf_lang2(getNamespaceSym, RcppString));
-    Rcpp::Shield<SEXP> RCPP(Rcpp_eval(call, R_GlobalEnv));
+    Rcpp::Shield<SEXP> RCPP(Rf_eval(call, R_GlobalEnv));
     Rcpp::Shield<SEXP> cache(Rf_allocVector(VECSXP, RCPP_CACHE_SIZE));
 
     // the Rcpp namespace
--- a/src/barrier.cpp
+++ b/src/barrier.cpp
@@ -28,7 +28,7 @@
 #include <algorithm>
 #include <Rcpp/protection/Shield.h>
 
-namespace Rcpp { SEXP Rcpp_fast_eval(SEXP, SEXP); }
+namespace Rcpp { SEXP Rcpp_eval(SEXP, SEXP); }
 
 // [[Rcpp::register]]
 SEXP get_string_elt(SEXP x, R_xlen_t i) {
@@ -97,7 +97,7 @@ SEXP get_rcpp_cache() {
         SEXP getNamespaceSym = Rf_install("getNamespace"); // cannot be gc()'ed  once in symbol table
         Rcpp::Shield<SEXP> RcppString(Rf_mkString("Rcpp"));
         Rcpp::Shield<SEXP> call(Rf_lang2(getNamespaceSym, RcppString));
-        Rcpp::Shield<SEXP> RCPP(Rcpp_fast_eval(call, R_GlobalEnv));
+        Rcpp::Shield<SEXP> RCPP(Rcpp_eval(call, R_GlobalEnv));
 
         Rcpp_cache = Rf_findVarInFrame(RCPP, Rf_install(".rcpp_cache"));
         Rcpp_cache_know = true;
@@ -139,7 +139,7 @@ SEXP init_Rcpp_cache() {
     SEXP getNamespaceSym = Rf_install("getNamespace"); // cannot be gc()'ed  once in symbol table
     Rcpp::Shield<SEXP> RcppString(Rf_mkString("Rcpp"));
     Rcpp::Shield<SEXP> call(Rf_lang2(getNamespaceSym, RcppString));
-    Rcpp::Shield<SEXP> RCPP(Rcpp_fast_eval(call, R_GlobalEnv));
+    Rcpp::Shield<SEXP> RCPP(Rcpp_eval(call, R_GlobalEnv));
     Rcpp::Shield<SEXP> cache(Rf_allocVector(VECSXP, RCPP_CACHE_SIZE));
 
     // the Rcpp namespace
--- a/src/barrier.cpp
+++ b/src/barrier.cpp
@@ -28,7 +28,7 @@
 #include <algorithm>
 #include <Rcpp/protection/Shield.h>
 
-namespace Rcpp { SEXP Rcpp_eval(SEXP, SEXP); }
+namespace Rcpp { SEXP Rcpp_fast_eval(SEXP, SEXP); }
 
 // [[Rcpp::register]]
 SEXP get_string_elt(SEXP x, R_xlen_t i) {
@@ -97,7 +97,7 @@ SEXP get_rcpp_cache() {
         SEXP getNamespaceSym = Rf_install("getNamespace"); // cannot be gc()'ed  once in symbol table
         Rcpp::Shield<SEXP> RcppString(Rf_mkString("Rcpp"));
         Rcpp::Shield<SEXP> call(Rf_lang2(getNamespaceSym, RcppString));
-        Rcpp::Shield<SEXP> RCPP(Rcpp_eval(call, R_GlobalEnv));
+        Rcpp::Shield<SEXP> RCPP(Rcpp_fast_eval(call, R_GlobalEnv));
 
         Rcpp_cache = Rf_findVarInFrame(RCPP, Rf_install(".rcpp_cache"));
         Rcpp_cache_know = true;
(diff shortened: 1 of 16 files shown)


eddelbuettel:
Passing Travis is an excellent first step :)

coatless:
Looks good. Though, the switch over is missing in `src/ barrier`:

https://github.com/RcppCore/Rcpp/blob/5d5cdb85504ae6e12916d963bdd621a41730d513/src/barrier.cpp#L30-L32

https://github.com/RcppCore/Rcpp/blob/5d5cdb85504ae6e12916d963bdd621a41730d513/src/barrier.cpp#L99-L101

https://github.com/RcppCore/Rcpp/blob/5d5cdb85504ae6e12916d963bdd621a41730d513/src/barrier.cpp#L141-L143

Enchufa2:
@thirdwing Why did you switch back to `Rcpp_eval` in `src/barrier.cpp` in the second commit?

thirdwing:
@Enchufa2 I have some linking errors on one machine after changing `barrier.cpp`, so I said I need a little more time.

eddelbuettel:
No issues in rev.dep, results committed in rcpp-logs repo under https://github.com/RcppCore/rcpp-logs/commit/8d355ce3a2fe9e76279500ebeb979051c731b32e

kevinushey:
This PR looks good to me as well. @lionel- any reservations around using `Rcpp_fast_eval()` by default internally in Rcpp? (Do you think we should be using it by default with R 3.5.x as well?)

lionel-:
If `RCPP_PROTECTED_EVAL` is unset (or if it is set but R version is < 3.5) `Rcpp_fast_eval()` devolves to `Rcpp_eval()` so this should be safe to merge.

@thirdwing Have you tested this with protect-unwind enabled? You can enable it by adding `#define RCPP_PROTECTED_EVAL` on top of Rcpp.h.

> Do you think we should be using it by default with R 3.5.x as well?

I was thinking we should first get a Rcpp release with protect-unwind disabled by default, then enable it in our packages to get some user exposure, and only then turn it on by default in Rcpp.

eddelbuettel:
I will once again run a second rev.dep with the `#define` on, probably overnight.  

Agreed that this should be safe to merge as is.

lionel-:
> I will once again run a second rev.dep with the #define on, probably overnight.

Will this be on R 3.5? If R 3.4 the define will not have an effect.

> Agreed that this should be safe to merge as is.

On second thought this might cause trouble. Switching to fast eval means errors will be rethrown as is instead of being wrapped in a `Rcpp::eval_error`. Also the error message will not start with "Evaluation error: ". Since users tend to check for errors based on the error message (and more rarely on the error class), this might cause revdep issues.

I think it's the right move to eventually switch to fast eval, but it might not be as smooth a transition as we could hope for. So it might be more prudent to first enable protect-unwind and only then switch to fast eval throughout Rcpp, so we can more easily assess the impact of enabling protect-unwind on revdep checks.

eddelbuettel:
> If R 3.4 the define will not have an effect.

Crap.  Time to bite the bullet I suppose and rebuild the rev.dep checker.

Enchufa2:
With R 3.5.0 and `RCPP_PROTECTED_EVAL` defined, I see

```
undefined symbol Rcpp::Rcpp_eval(SEXPREC*, SEXPREC*)
```

when installing Rcpp. If I substitute `Rcpp_eval` calls in `src/barrier.cpp` with `Rcpp_fast_eval` calls, the issue goes away.

eddelbuettel:
R 3.5.0 _without_ the `#define` passed just fine.

I just started R 3.5.0 _with_  the `#define` (and the fix `barrier.cpp`).

eddelbuettel:
Using `Rf_eval` sounds like a good plan there.

thirdwing:
@Enchufa2 Can you try the code again with `RCPP_PROTECTED_EVAL` defined?

Enchufa2:
I'm still running revdeps in 3.5.0 with the flag enabled in Rcpp, but as I
said, the machine is not very powerful and even building the environment
was painful. I had to exclude some packages due to memory problems (4GB)
while compiling (e.g., all Stan-based stuff), and I excluded Bioc-based
stuff too.

So at this rate, probably Dirk's checks will end before mine. :) But so
far, it looks good: 650 succeeded, and a few failures due to missing
dependencies (so my fault).

eddelbuettel:
All "mostly" good here too with the 
- R 3.5.0
- `#define` for `RCPP_PROTECTED_EVAL`
- `barrier.cpp` edited to use `Rcpp_fast_eval` (didn't think of Rf_eval then)
and everything passes but I got errors on reticulate, tidyxl and v8. Those matter, so I will check again
later with `Rf_eval`.  

We _could_ still merge as `RCPP_PROTECTED_EVAL` is not generally on.

eddelbuettel:
Regression altert: packages
- reticulate
- tidyxl
- V8

all fail their tests with this version of the PR (ie `Rf_eval` in `src/barrier.cpp`) as well as with the one from one commit back (using `Rcpp_fast_eval).   Any chance you could examine this, @lionel- ?

lionel-:
Happy to, could you link to or send me the logs please?

Did you run the checks on master or on this branch?

eddelbuettel:
This branch, once without the define, once with. That second run had an imperfect `barrier.cpp` but I ran the three failures again with the branch as it is now.

Not my machine, so not easy for access. I will tar them up and put them on my webserver.  Good enough?  I'll DM you the URL.

Enchufa2:
> Regression alert: packages

Same here.

eddelbuettel:
I am leaning towards merging as is, despite the three known regressions as it is still opt-in with the `#define`.  

Any thoughts?

thirdwing:
I am looking into the checking error in `reticulate`. Please give me one more day.

src/barrier.cpp
```diff
@@ -97,7 +95,7 @@ SEXP get_rcpp_cache() {
         SEXP getNamespaceSym = Rf_install("getNamespace"); // cannot be gc()'ed  once in symbol table
         Rcpp::Shield<SEXP> RcppString(Rf_mkString("Rcpp"));
         Rcpp::Shield<SEXP> call(Rf_lang2(getNamespaceSym, RcppString));
-        Rcpp::Shield<SEXP> RCPP(Rcpp_eval(call, R_GlobalEnv));
+        Rcpp::Shield<SEXP> RCPP(Rf_eval(call, R_GlobalEnv));
```
lionel-: This should be evaluated in base or in the Rcpp namespace.

lionel-: Most (all?) of these `Rcpp_eval()` / `Rcpp_fast_eval()` actually :)

Probably best to make these changes in another PR. And merge / check that PR before this one.

eddelbuettel: Or modify this one and test it again?

lionel-: I was thinking a different PR would make things easier should this one be reversed.

lionel-:
I can reproduce the failures. tidyxl is ok (it's failing because it's checking the error string). Working on figuring out the other two between meetings (it's the RStudio work week).

I don't think we should merge this PR right now because master doesn't have the failures, so there is something wrong here.

src/barrier.cpp
thirdwing: I think you can send another PR and I will reverse this commit.

eddelbuettel: What do you mean?  This PR is three commits.  What do you suggest reversing?

thirdwing: I mean the changes in `barrier.cpp`.

eddelbuettel: Could do.  And test that again, or first against those three?   And wasn't there a link issue?

lionel-:
The V8 issue is because Rcpp code is run in callbacks: https://github.com/jeroen/V8/blob/c67abd98797e488adc941ce67097b06e6547f6dc/src/V8.cpp#L63

@jeroen I think it would be safer not to use Rcpp in such callbacks and use protect-unwind manually.

Going to investigate the reticulate error soon.

thirdwing:
@lionel- The error in `reticulate` should come from the changes in `inst/include/Rcpp/generated/Function__operator.h`

jjallaire:
In reticulate we are executing an R function while within a C Python stack
frame. Here is where we execute the R function:
https://github.com/rstudio/reticulate/blob/master/src/python.cpp#L1259-L1272

thirdwing:
@lionel- I think the error in `reticulate` is because `Rcpp_fast_eval` never `throw`.

lionel-:
@thirdwing `Rcpp_fast_eval()` does throw, this is how it gets to unwind the C++ stack.

When sourcing a Python script, the getter defined here https://github.com/rstudio/reticulate/blob/43664503884625f55cd4d452228a7a855ce86ce2/R/python.R#L1323 is called from the try-catch block linked by @jjallaire. The getter parses strings as R code but it is passed two non-syntactic values `__call__` and `__iter__`:

```
// rArgs:
[[1]]
<__main__.R>

[[2]]
__call__

Error in parse(text = as_r_value(code)) : <text>:1:1: unexpected input
1: _
    ^

// rArgs:
[[1]]
<__main__.R>

[[2]]
__iter__

Error in parse(text = as_r_value(code)) : <text>:1:1: unexpected input
1: _
    ^
```

So `Function doCall("do.call")` throws an R error and then a C++ error.  With `Rcpp_eval()` the R error is caught by the R `tryCatch()` and an exception inheriting from `std::exception` is caught by the C++ try/catch. With `Rcpp_fast_eval()` the error is not caught on the R side but protect-unwind kicks in and throws a C++ LongjumpException. Since the latter does not inherit from `std::exception`, it hits the catch-all block. However this difference doesn't matter here because it only impacts the Python error string of `errorDict`.

The reason we observe a different behaviour is because if you don't catch the error on the R side, it gets printed before initiating the error longjump. Hence catching a longjump from the C++ side without catching it on the R side will always produce the side effect of printing the error message as part of the unwinding.

I think in general these longjumps should not be interrupted. The good practice would be to catch the `LongjumpException` (which is currently namespaced in `Rcpp::internal`, we might want to make it public) and rethrow it immediately. For the reticulate error it might be harder to rethrow since the libpython stack is on the way. Maybe we should add a getter for the SEXP token which would be easier to pass around, and make public a constructor for a LongjumpException that'd take the token as argument. This way you can easily pass the token in the C contexts and resume the longjump once back in Rcpp context (or maybe use the R API to resume the longjump in you're no longer in Rcpp context and it is safe to do so).

For the cases where you really want to interrupt the longjump, this gets trickier if Rcpp switches to fast_eval everywhere. But we could provide some kind of `Rcpp_catch_unwind()` function that takes a C/C++ function pointer that would be called from an R `tryCatch()` context. This way we pay the `tryCatch()` overhead only once and still get to properly catch the R errors.

lionel-:
The good news is that it appears fast_eval is doing the right thing in all cases :)

lionel-:
Oh and the last piece of the puzzle is that unhandled errors are recorded by testthat in a calling handler. So even though the code path eventually succeeds the error is detected and reported to R CMD check. This is another reason not to handle errors on the C++ side without handling them on the R side as well. And either resume the jump or evaluate Rcpp code in a `Rcpp_catch_unwind()` context as argued above. This way the R and C++ semantics are aligned.

eddelbuettel:
So what is the consensus then for next steps?  Merge this, have a follow-up PR?

lionel-:
I think this can be merged and I'll send a follow-up PR with an API to deal with handling of R errors at the C++ level. I'll also look into the eval envs.

PR #869, opened 2018-06-13
Resolve segfault when running embeddedR test

filipsch:
As @lionel- suggested, having the new environments be a child of the base environment fixed the nasty side effects described by @eddelbuettel and @coatless in https://github.com/RcppCore/Rcpp/issues/851#issuecomment-395932776


eddelbuettel:
Worked like a charm in [reverse depends](https://github.com/RcppCore/rcpp-logs/commit/c91fdb9b673cfefd501d738db697ab6c91375f98), and now part of Rcpp 0.12.17.3 -- so thanks!

PR #872, opened 2018-06-19
Remove editor headers

lionel-:
Follow up to #842 

The Emacs file-local settings are not consistent across files and take precedence over the the new `editorconfig` settings, causing confusion while editing.

### DIFF
--- a/inst/examples/ConvolveBenchmarks/convolve10_cpp.cpp
+++ b/inst/examples/ConvolveBenchmarks/convolve10_cpp.cpp
@@ -1,5 +1,3 @@
-// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; tab-width: 8 -*-
-
 // this version expands convolve8_cpp by making Vec mimic the structure of
 // NumericVector. It peforms well, so this is is not the structure of
 // NumericVector that is the problem. So what is it then ?
--- a/inst/examples/ConvolveBenchmarks/convolve11_cpp.cpp
+++ b/inst/examples/ConvolveBenchmarks/convolve11_cpp.cpp
@@ -1,5 +1,3 @@
-// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; tab-width: 8 -*-
-
 // This version uses nona to indicate that xb does not contain any missing
 // value. This is the assumption that all other versions do.
 
--- a/inst/examples/ConvolveBenchmarks/convolve12_cpp.cpp
+++ b/inst/examples/ConvolveBenchmarks/convolve12_cpp.cpp
@@ -1,5 +1,3 @@
-// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; tab-width: 8 -*-
-
 // This is a rewrite of the 'Writing R Extensions' section 5.10.1 example
 
 #include <Rcpp.h>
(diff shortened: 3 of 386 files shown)


eddelbuettel:
Please file issue tickets before making PRs as the file [Contributing.md] suggests. 

I am likely going to ignore this -- we have no cared about cosmetics such as this for many years, and I don't think we care now either.

lionel-:
This is not cosmetics. These headers mess up the Emacs settings from editorconfig. They cause wrong display and wrong indentation.

eddelbuettel:
I use Emacs with editorconfig too and have no issues.  In any event this has ALWAYS been a mess as we never had a consistent policy let alone behaviour here given the sometimes somewhat "volatile" action around here.

I am unlikely to be in favour of PRs touching hundreds of files. I would be in favour of figuring out something saner for all / most (and I do like editorconfig) and then move towards as we change files.

_Maybe_ several months in with experience all around we can discuss bulk changes.  But not as a start.

lionel-:
> I use Emacs with editorconfig too and have no issues.

Probably because you have edited files whose settings are congruent with those of editorconfig. Try any file with `tab-width: 8` and tabs written by someone who assumed a tab size of 4 (random example, inst/include/Rcpp/StringTransformer.h).

Anyway, I can now work on Rcpp based on this branch and rebase it out before sending other PRs. Feel free to reopen anytime.

eddelbuettel:
Tabs equal four was an old setting; we tried changing but never went anywhere.  The last (uncoordinated) mass commit was by @kevinushey a few years ago and along those lines.  I think we concluded that mass commits are generally not a good idea.

I also think we now feel that spaces is preferable to tabs, but I still don't think it is worth messing up file history by carpet bombing all files. I tend to M-x untabify when I come across it.

kevinushey:
Is there a way to enforce this in Emacs without touching each file? (E.g. some way to say in `.dir_locals.el` either "don't read file-specific indentation settings" or perhaps "re-load `.editorconfig` after parsing those settings")

lionel-:
There are various ways around this. The less intrusive I thought of is to rehook locally editorconfig after hack-local-variables (the function that applies local vars) from the local vars themselves. In `.dir-locals.el`:

```elisp
((nil . ((eval . (add-hook 'hack-local-variables-hook 'editorconfig-apply nil 'local)))))
```

By the way, I agree that incremental changes are preferable, but I don't think this is a case of carpet bombing. First because the changes in this PR are localised to the headers and so wouldn't get in the way of someone blaming/praising the git history. And second because pure line substractions actually don't show up in blame at all. (I am making the assumption that commit history is the reason to be wary about mass edits.)

eddelbuettel:
I like files and directories at the OS level; I keep the same ones for years and tar up and copy.  Giving EVERY header file a new timestamp is just not cool.

If we must, we could remove editorconfig from the repo and move it to gitignore.  Much less trouble.

lionel-:
Got it.

> If we must, we could remove editorconfig from the repo and move it to gitignore. Much less trouble.

I think editorconfig is awesome and the way to go. Let's just remove the headers when they cause trouble while editing a file if that's ok.

eddelbuettel:
That's what I generally do.  I also (when I remember) remove TABs with spaces, and generally remove the only-used-by-Romain-an-nobody-else extra level of spaces.

eddelbuettel:
And I would very much in favour of the three of us sticking our heads together and defining a sane (and simple) policy.  I'd rather not have it too descriptive (and I just learned about things editorconfig cannot do :-/ as eg the 'no-namespace-indentation' that $WORK prefers).

PR #876, opened 2018-07-09
AIX doesn't have tm_gmtoff

ayappanec:
AIX doesn't have tm_gmtoff due to which the compilation is failing. This PR will fix this issue.


eddelbuettel:
Hm, we have not had any AIX reports in some. I wonder why that did not come up before.  I see two other files with five total uses for `#ifdef _AIX` so this will probably not cause harm -- but still strange we did not need it before.

ayappanec:
Not sure. I am just trying R in AIX and i found this issue.

PR #877, opened 2018-07-10
Improvements for unwind-protect

lionel-:
- Add a plugin for enabling unwind-protect. Using `// [[Rcpp::plugins(unwindProtect)]]` is safer than `#define` because the plugin ensures unwind-protect is enabled in all compilation units including RcppExports.cpp. This avoids linking issues.

- Make `LongjumpException` public. Users should catch and rethrow it before `catch (...)` so longjumps are not ignored.

- Move `unwindProtect()` to Rcpp/unwindProtect.h. Include it early in RcppCommon to make it available in other header files. This will be useful for calling into R APIs for initialising static variables etc (more on that later).

- Rename `RCPP_PROTECTED_EVAL` to `RCPP_USE_UNWIND_PROTECT`. Then `RCPP_USING_UNWIND_PROTECT` is defined in headers.h if the R version is sufficient. The latter define can be used in the rest of header files.

- Refactor unit tests so that the unwound indicator is an environment rather than a logical vector. I got into trouble with the latter because the bytecode compiler replaces scalar logical with globally shared `TRUE` or `FALSE` values. Mutating these values corrupts the runtime.

- Pass `...` from `unitTestSetup()` to `sourceCpp()` for easy debugging.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+
+2018-07-10  Lionel Henry  <lionel@rstudio.com>
+
+        * inst/include/Rcpp/exceptions.h: Move LongjumpException from the
+        Rcpp::internal namespace to the public Rcpp namespace. If you have a
+        catch-all statement like `catch (...)`, be sure to catch and rethrow
+        Rcpp::LongjumpException to prevent the R longjump from being ignored.
+        * inst/include/Rcpp/macros/macros.h (VOID_END_RCPP): idem
+        * src/attributes.cpp: idem
+
 2018-07-05  Lionel Henry  <lionel@rstudio.com>
 
         * inst/include/Rcpp/api/meat/Rcpp_eval.h: Rename `RCPP_PROTECTED_EVAL`
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+
+2018-07-05  Lionel Henry  <lionel@rstudio.com>
+
+        * R/unit.tests.R (unitTestSetup): Pass extra arguments to sourceCpp()
+        for easier debugging.
+
 2018-06-22  Kevin Ushey  <kevinushey@gmail.com>
 
         * inst/include/Rcpp/api/meat/Rcpp_eval.h: Ensure R_BaseEnv is used
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,13 +1,22 @@
+2018-07-05  Lionel Henry  <lionel@rstudio.com>
+
+        * inst/include/Rcpp/api/meat/Rcpp_eval.h: Rename `RCPP_PROTECTED_EVAL`
+        to `RCPP_USE_UNWIND_PROTECT` because the new API is now more general
+        than just evaluation of R code.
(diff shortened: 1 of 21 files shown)


lionel-:
@eddelbuettel Would you like to rename `Rcpp::Rcpp_fast_eval()` here? If yes do you want to try `Rcpp::eval()`? Alternatively `Rcpp::evalProtect()` might not be too bad a name.

lionel-:
oops R CMD check fails on Windows with this branch, will investigate.

eddelbuettel:
Hi @lionel- -- there is a lot to unpack here. I am wondering if we should split it over several smaller PRs.

Lots of it looks really good and like something we can / should do (the plugin, moving to a normal header included included early, ...) but it seems we are wrapping this up with turning more things on automatically which may still be risky.

I was also thinking about a release "soon" so maybe doing this in parts may be safer?

lionel-:
>  it seems we are wrapping this up with turning more things on automatically which may still be risky

Sorry I didn't understand that. What do you mean?

> I was also thinking about a release "soon" so maybe doing this in parts may be safer?

I have put off all new/risky things for later PRs that I will send after the release. I think almost all of this should be in the next release:

* Making LongjumpException public is essential for compatibility of unwind-protect with code that uses `catch (...)`, so users can avoid swallowing the LJ exception.
* Enabling unwindProtect with the plugin ensures the correct Makevars flag is used to avoid UB issues.
* It is safer to remove the scalar logical mutation from the unit tests. Though the unit tests do work for the current version of R.

On the other hand including `unwindProtect()` earlier in the build process is not essential right now. Would you prefer to remove this commit?

Oh and also I meant to remove the default argument of `Rcpp_fast_eval()`. The global env is often not the right default and I think we should force users to think about where they want to evaluate (their own namespace, the base env, the global env?). Any objection? It is important to do this before release so we don't modify the API later on.

Likewise, it would be preferable to rename fast_eval to make it consistent with the rest of the Rcpp API before release.

lionel-:
> oops R CMD check fails on Windows with this branch, will investigate.

It was not Windows but on 3.5 in general, my bad for not checking the last commit properly. Now fixed.

eddelbuettel:
> Sorry I didn't understand that. What do you mean?

Maybe I misread above at first.  Does it, or does it not, turn the new behaviour on by default?  Last time we still had three important packages fail.  Can you clarify?

lionel-:
The plugin is for enabling unwind-protect, which is disabled by default. The three package fails were actually normal failures but I've prepared some tools to help fix those. I will send a new PR after the release.

Any objection to including the following commit to this PR? https://github.com/lionel-/Rcpp/commit/2ba5095b20465b8c3cd17fa876cca28aa8a67d68 It removes the default argument for fast_eval(). This commit passes R CMD check on mac, Linux, Windows, with R 3.4 and R 3.5.

eddelbuettel:
Yes toss it in. May as well test all at once.

eddelbuettel:
"No change to worse", ie no new rev.dep failures in this PR relative to the master branch, so merging in.

PR #878, opened 2018-07-12
sort case-insensitive to make RcppExports more deterministic 

jackwasey:
`compileAttributes` uses `list.files` which gives a locale-dependent ordering of the matched files. This differs between C and en-us.UTF-8, with C ordering putting upper-case file names first. This makes the code output to `RcppExports.*` different depending on the locale.

It was not obvious to me how to make this locale independent using base R, but at least ignoring case when calling `list.files` eliminates some of the discrepancies produced with these two widely used collations. People calling `compileAttributes()` from locales which are not ordered A-Z will see differences in the output order (and thus order of functions in the generated code).

### DIFF
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -422,7 +422,7 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
         dir.create(rDir)
 
     # get a list of all source files
-    cppFiles <- list.files(srcDir, pattern = "\\.((c(c|pp)?)|(h(pp)?))$")
+    cppFiles <- list.files(srcDir, pattern = "\\.((c(c|pp)?)|(h(pp)?))$", ignore.case = TRUE)
 
     # derive base names (will be used for modules)
     cppFileBasenames <- tools::file_path_sans_ext(cppFiles)


kevinushey:
A number of packages that have required locale-independent sorting have implemented a `sort_c` function, e.g. see what roxygen2 does here:

https://github.com/klutometis/roxygen/blob/bbf259ddee033a873426ded1e83cdd344b986595/R/util-locale.R#L1-L14

We could potentially do something similar in Rcpp -- explicitly set LC_COLLATE, sort the list of files, and then restore the old LC_COLLATE.

kevinushey:
But if setting `ignore.case` is enough to handle the most common cases then I think that's sufficient.

jackwasey:
I'll leave it to your judgement: I agree this is a quick win which, at least for me, covers my fairly vanilla Ubuntu and Mac setups, which were producing different RcppExports code. I'm sure those case-insensitive windows users would also have bad interactions when also developing also on case-sensitive operating systems.

eddelbuettel:
This is an easy win. We could always look into setting locale to C, sorting, then resetting as @kevinushey suggested.  But until then this works for me.

hadley:
Could we please change the locale here? That will also ensure consistency from system to system.

eddelbuettel:
Quoting @jjallaire [in the same thread](https://github.com/apache/arrow/pull/2596#issuecomment-424409293)  : 

> A recent change in Rcpp (v0.12.18 from a couple of months ago) changed the sorting behavior to be more stable across locales/operating systems: RcppCore/Rcpp#878
>
> Net is that this should be a one-time issue and that the order will be better preserved across environments in the future.

hadley:
That link is to this issue?

The problem with only ignoring the case is that you will still get a different default ordering in locales that don't order letters the same way as English. This means that collaborators with different locales will still generate different RcppExports.

kevinushey:
We could set the locale to C temporarily and then sort the discovered files, e.g.

```R
cppFiles <- list.files(srcDir, pattern = "\\.((c(c|pp)?)|(h(pp)?))$", ignore.case = TRUE)

locale<- Sys.getlocale(category = "LC_COLLATE")
Sys.setlocale(category = "LC_COLLATE", locale = "C")
cppFiles <- sort(cppFiles)
Sys.setlocale(category = "LC_COLLATE", locale = locale)
```

jackwasey:
@kevinushey this looks good to me.

PR #882, opened 2018-07-20
when debugging, abort with error on cast when incompatible with STRSXP

jackwasey:
As with #860, if `NDEBUG` is unset, then abort on a failed cast to `STRSXP`. #860 missed this code path. (Although I think this is a benign, the current master 2d4eb219bdff6bb4675e46b doesn't pass its own tests, so I can't complete an R CMD check after this patch.)

### DIFF
--- a/inst/include/Rcpp/r_cast.h
+++ b/inst/include/Rcpp/r_cast.h
@@ -123,7 +123,12 @@ namespace Rcpp {
                 return Rf_ScalarString( PRINTNAME( x ) );
             default:
                 const char* fmt = "Not compatible with STRSXP: [type=%s].";
+#ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Thanks, that PR looks fine.

What do you mean by "the current master 2d4eb21 doesn't pass its own tests" ? It obviously works at Travis, as it does on my box.  Can you detail what goes wrong at your end?

jackwasey:
Thanks. BTW, I was unable to reproduce the RUnit test failures in a clean docker environment.

eddelbuettel:
Thanks for the follow-up.  Also have (as you may have seen) a pending PR for the `min()` and `max()` issue.

jackwasey:
Yes, I spotted that: impressive speed, much appreciated.

PR #884, opened 2018-07-20
consider empty vectors in min() and max() (fixes #883)

### LINKED ISSUE #883: max of empty IntegerVector gives large negative number
Not sure if this is designed to be somewhat consistent with R (which gives `-Inf` and a warning for `max(integer(0))`.

```
Rcpp::cppFunction('int getiv() { IntegerVector IV = IntegerVector::create(); return max(IV); }')
getiv()
# [1] -368909088
```

Don't know where the number came from, other than perhaps a random memory location on my machine. If it is picking random memory locations, I guess it could cause some kind of memory access violation? I don't see an obviously correct choice of output in this case.

eddelbuettel:


eddelbuettel:
Reviews / thumbs up-or-down welcome!   @kevinushey @jjallaire @thirdwing @nathan-russell ...

kevinushey:
LGTM!

coatless:
Why not use `R_NaN` instead of `R_NegInf`?

eddelbuettel:
Because of this:

```r
R> v <- numeric(0); c(min(v), max(v))
[1]  Inf -Inf
Warning messages:
1: In min(v) : no non-missing arguments to min; returning Inf
2: In max(v) : no non-missing arguments to max; returning -Inf
R> 
```

In the integer case it does become `NA` from the `static_cast`. Have not tried `complex` or other vectors.

eddelbuettel:
Running another round of reverse depends now, and only hickup is in BigVAR. It has in file [src/ExperimentalBigVARFunctionsX.cpp](https://github.com/wbnicholson/BigVAR/blob/master/BigVAR/src/ExperimentalBigVARFunctionsX.cpp)  a function (indented here)

```C++
colvec ThreshUpdateOO(const mat& ZZ, double lam,const mat& Y,double eps, List groups, \
       List fullgroups, List compgroups,List M2f_, List eigvalF_, List eigvecF_, \
       colvec& B,int n, int k1) {
```

with a test

```C++
    if(max(groups)==0)
```

and we now get 

```sh
ccache g++ -std=c++11 -std=gnu++11 -I"/usr/share/R/include" -DNDEBUG  -I"/usr/local/lib/R/site-library/Rcpp/include" -I"/usr/local/lib/R/site-library/RcppArmadillo/include" -I"/usr/local/lib/R/site-library/Rc\
ppEigen/include"   -DNDEBUG -fpic  -g0 -O3 -Wall -pipe -Wno-unused  -fext-numeric-literals -march=native -c ExperimentalBigVARFunctionsX.cpp -o ExperimentalBigVARFunctionsX.o                                   
In file included from /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/functions.h:50:0,  
                 from /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/sugar.h:31,                  
                 from /usr/local/lib/R/site-library/Rcpp/include/Rcpp.h:74,                              
                 from /usr/local/lib/R/site-library/RcppArmadillo/include/RcppArmadillo.h:34,            
                 from ExperimentalBigVARFunctionsX.cpp:1:                                                
/usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/max.h: In instantiation of ‘Rcpp::sugar::Max<RTYPE, NA, T>::operator Rcpp::sugar::Max<RTYPE, NA, T>::STORAGE() const [with int RTYPE = 19; bool NA = true; T = Rcpp::Vector<19>; Rcpp::sugar::Max<RTYPE, NA, T>::STORAGE = SEXPREC*]’:                   
ExperimentalBigVARFunctionsX.cpp:590:20:   required from here                                            
/usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/max.h:36:32: error: invalid static_cast from type ‘double’ to type ‘Rcpp::sugar::Max<19, true, Rcpp::Vector<19> >::STORAGE {aka SEXPREC*}’       
             if (n == 0) return(static_cast<STORAGE>(R_NegInf));                                         
                               ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                          
/usr/lib/R/etc/Makeconf:168: recipe for target 'ExperimentalBigVARFunctionsX.o' failed                   
make: *** [ExperimentalBigVARFunctionsX.o] Error 1                                                       
ERROR: compilation failed for package ‘BigVAR’
```

but I am a little confused how that would have worked before `max()` over a `Rcpp::List` comparing to a scalar?  Did that ever make sense?  

Is this a regression we need to fix?

kevinushey:
Looks like `max()` doesn't actually do what one might expect it to do when applied to a `List`:

`````
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
SEXP max(List data) {
   return Rcpp::max(data);
}

/*** R
max(list(1, 2, 3, 4, 5))
*/
`````

```
> Rcpp::sourceCpp('~/scratch/Untitled.cpp')
> max(list(1, 2, 3, 4, 5))
[1] 1
```

IMHO it's fine that this is a compilation error and perhaps provide a patch (or request a change) from `BigVAR`.

**Comment by kevinushey** [comment]
The R level APIs also don't understand lists:

```R
> base::max(list(1, 2, 3, 4, 5))
Error in base::max(list(1, 2, 3, 4, 5)) : 
  invalid 'type' (list) of argument
```

so IMHO further evidence that not supporting `List`s is fine.

eddelbuettel:
Agreed. Did the same / similar test with `max()` / `min()` around `Rcpp::List` for both 0.12.17 and 0.12.18 -- results were incorrect for 0.12.17 as you show, and 0.12.18 fails to compile.  With an added `as<>()` it compiles and still returns the wrong result :)

So I concur: It was sloppy code by us before allowing this compile, but this particular snipped may actually be an issue in `BigVAR`. I also contacted its maintainer, @wbnicholson.

wbnicholson:
Thanks for letting me know about this issue.  I will remove the max call from my package and will re-upload a new version to cran.

eddelbuettel:
Awesome (as also discussed via email).  

Turns out that this was the only 'regression'  so I will wrap up Rcpp 0.12.18 for CRAN this weekend.

Issue #879, opened 2018-07-17
`Rcpp::exposeClass` doesn't work with different `class` and `CppClass` arguments


mlysy:
Here is a minimal example.  First consider a simple C++ class defined in `foo.h`:

```cpp
#ifndef foo_h
#define foo_h 1

// class to set/get a scalar double
class foo {
 private:
  double x;
 public:
  double get_x() {return x;}
  void set_x(double _x) {x = _x; return;}
  foo(double _x) {set_x(_x);}
};

#endif
```

Now let's create an **R** package exporting the class as an Rcpp module:

```r
options(useFancyQuotes = FALSE) # Rcpp::exposeClass uses dQuote
pkg_path <- getwd() # where to create package
pkg_name <- "ModuleLoadTest" # name of package
pkg_dir <- file.path(pkg_path, pkg_name) # package root directory
# create package set up to export Rcpp modules
Rcpp::Rcpp.package.skeleton(name = pkg_name,
                            path = pkg_path,
                            force = TRUE,
                            example_code = FALSE, module = TRUE)
# remove module sample code
file.remove(list.files(file.path(pkg_dir, c("src", "R")),
                       full.names = TRUE))
# add foo header file to package (assuming foo.h is in current directory)
file.copy(from = "foo.h", to = file.path(pkg_dir, "src", "foo.h"))
```

Finally to wrap the class into a module.  First with the **R** and C++ classes both named `foo`, which works as expected:

```r
Rcpp::exposeClass(class = "foo",
                  constructors = list("double"),
                  fields = character(),
                  methods = c("get_x", "set_x"),
                  header = '#include "foo.h"',
                  CppClass = "foo",
                  file = file.path(pkg_dir, "src", "fooModule.cpp"),
                  Rfile = file.path(pkg_dir, "R", "fooClass.R"))
devtools::install(pkg_dir) # install package
# simple unit test for getter/setter
bar <- ModuleLoadTest::foo(0) # create object
replicate(n = 10, {
  x <- rnorm(1)
  bar$set_x(x)
  bar$get_x() - x
}) # should return an array of zeros
```

However, when I change `class = "foo"` to `class = "fooR"` in the above, and rerun `devtools:install` (also checked with `R CMD install` and with CRAN and GitHub versions of Rcpp), I get the following output/error message:
```r
Installing ModuleLoadTest
'/Library/Frameworks/R.framework/Resources/bin/R' --no-site-file --no-environ  \
[...]

PR #886, opened 2018-07-24
bugfixes for Rcpp::exposeClass

mlysy:
Attempts to fix issue #879 and a couple of other minor issues with `Rcpp::exposeClass`, namely:

* `loadRcppClass` (called by `exposeClass`) handles the `Class` and `CppClass` arguments correctly now, such that the R and C++ class names for a given module needn't be the same.
* The `rename` argument to `exposeClass` now functions properly.
* When `exposeClass` is called in the root directory of a package, the generated wrapper code is put in `src` and `R` subdirectories.  This behavior is now ignored if `file` or `Rfile` contain absolute paths, in which case the files are written there.

A basic unit test for these changes (`inst/unitTests/runit.exposeClass.R`) is provided.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2018-07-24  Martin Lysy  <mlysy@uwaterloo.ca>
+
+        * R/loadRcppClass.R: Search in R module for 'Class' instead of 'CppClass'.
+        * R/exposeClass.R: Fixed 'rename' argument to work as expected.
+        * inst/unitTests/runit.exposeClass.R: Added unit tests for the above.
+
 2018-07-23  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/bib/Rcpp.bib: More updates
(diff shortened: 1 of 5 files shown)


eddelbuettel:
Wow. Thanks!  Will study.  Could you possibly rebase with master?  Your ChangeLog shows you branched after July 12 so you are missing a little bit.  It's not critical -- I could always squash-merge as well.  (And will hand-adjust ChangeLog.)

mlysy:
Sure! (I just distractedly updated only the ChangeLog, here comes the rest...)

eddelbuettel:
Interesting changes, eg the `Class` vs `CppClass`.  Were we operating on the wrong variable? Entirely possible, but also possible this may have side effects :)  I trust you tested this?

mlysy:
Hopefully this the correct rebase, modulo `ChangeLog`.

kevinushey:
Awesome! Thanks for the PR.

While I'm not that familiar with the modules codebase, all of the changes here look sensible and correct, so a thumbs up from me.

eddelbuettel:
Same here. Will run a rev.dep check just in case.

eddelbuettel:
Sorry, got distracted for a bit.  Reverse dependency check threw up nothing new, so merging this now.

PR #887, opened 2018-07-28
Feature/deprecate ldflags

eddelbuettel:


eddelbuettel:
Nothing too tricky here but a review or a thumbs up would still be appreciate.

Issue #892, opened 2018-08-18
no_init() triggers rchk errors


krlmlr:
Example package: https://github.com/krlmlr/RcppNoInitTest.

When using `no_init()` for declaring a variable with assignment, [rchk](https://github.com/kalibera/rchk) (with clang 6.0 on Ubuntu 18.04) detects an error. This goes away if `no_init()` is used in a copy constructor. 

```c++
IntegerVector ret_bad = no_init(1); // rchk detects problem
IntegerVector ret_ok(no_init(1)); // rchk detects no problem
```

I have created a test package to replicate the problem.

- https://github.com/krlmlr/RcppNoInitTest/commit/6856fbc0f1c3f4e96294914450ca0e0ce0a04ec8: First syntax, rchk sees a problem, see `.bcheck` file
- https://github.com/krlmlr/RcppNoInitTest/commit/6e77a951e0d5b65e272f9007ef1fbc7a7e4f4970: Second syntax, this particular rchk problem goes away (but still plenty of other problems)
- https://github.com/krlmlr/RcppNoInitTest/commit/1c3af740109bb3e626448d6900bc4f7b9a4784c8: First syntax with _Rcpp_ patch https://github.com/krlmlr/Rcpp/commit/5dbfca11da325a8e72a49b90ccdd1e8714f4089c, this particular rchk problem seems resolved
- https://github.com/krlmlr/RcppNoInitTest/commit/38aa1175660d07e95b63ab0ab890b403b7203729: Resolving more false positives with _Rcpp_ patch https://github.com/krlmlr/Rcpp/commit/4c4721aab4777fb60592511bbe571b98b1270bee

This is important to me, because _dplyr_ on CRAN shows these `rchk` errors: https://raw.githubusercontent.com/kalibera/cran-checks/master/rchk/results/dplyr.out.

I'll submit a PR for discussion pointing to this issue. Please advise.

eddelbuettel:
Taken care of by #893. Thanks!

PR #893, opened 2018-08-18
Rewrite no_init() and other functions to reduce false positives

krlmlr:
Reference: #892.

### DIFF
--- a/inst/include/Rcpp/exceptions.h
+++ b/inst/include/Rcpp/exceptions.h
@@ -243,7 +243,7 @@ namespace internal {
     inline bool is_Rcpp_eval_call(SEXP expr) {
         SEXP sys_calls_symbol = Rf_install("sys.calls");
         SEXP identity_symbol = Rf_install("identity");
-        SEXP identity_fun = Rf_findFun(identity_symbol, R_BaseEnv);
+        Shield<SEXP> identity_fun(Rf_findFun(identity_symbol, R_BaseEnv));
         SEXP tryCatch_symbol = Rf_install("tryCatch");
         SEXP evalq_symbol = Rf_install("evalq");
 
--- a/inst/include/Rcpp/protection/Shield.h
+++ b/inst/include/Rcpp/protection/Shield.h
@@ -26,6 +26,8 @@ namespace Rcpp{
     }
 
     inline void Rcpp_unprotect(int i){
+        // Prefer this function over UNPROTECT() in Rcpp so that all
+        // balance checks errors by rchk are contained at one location (#892)
         UNPROTECT(i);
     }
 
--- a/inst/include/Rcpp/protection/Shield.h
+++ b/inst/include/Rcpp/protection/Shield.h
@@ -25,7 +25,7 @@ namespace Rcpp{
         return x ;
     }
 
-    inline SEXP Rcpp_unprotect(int i){
+    inline void Rcpp_unprotect(int i){
         UNPROTECT(i);
     }
 
--- a/inst/include/Rcpp/protection/Shield.h
+++ b/inst/include/Rcpp/protection/Shield.h
@@ -25,12 +25,16 @@ namespace Rcpp{
         return x ;
     }
 
+    inline SEXP Rcpp_unprotect(int i){
+        UNPROTECT(i);
+    }
+
     template <typename T>
     class Shield{
     public:
         Shield( SEXP t_) : t(Rcpp_protect(t_)){}
         ~Shield(){
-            if( t != R_NilValue ) UNPROTECT(1) ;
+            if( t != R_NilValue ) Rcpp_unprotect(1) ;
         }
 
         operator SEXP() const { return t; }
--- a/inst/include/Rcpp/vector/no_init.h
+++ b/inst/include/Rcpp/vector/no_init.h
@@ -37,6 +37,8 @@ public:
 
     template <int RTYPE>
     operator Vector<RTYPE, PreserveStorage>() const {
+        // Explicitly protect temporary vector to avoid false positive
+        // with rchk (#892)
         SEXP x = PROTECT(Rf_allocVector(RTYPE, size));
         Vector<RTYPE, PreserveStorage> ret(x);
         UNPROTECT(1);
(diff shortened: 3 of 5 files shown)


inst/include/Rcpp/protection/Shield.h
```diff
@@ -25,12 +25,16 @@ namespace Rcpp{
         return x ;
     }
 
+    inline SEXP Rcpp_unprotect(int i){
```
eddelbuettel: This function wants to be `void` or else `g++` yells at you.

eddelbuettel: You can cherry-pick it from 0df99c8f247f4af9a0e8b7d362086056e435b08a

inst/include/Rcpp/vector/no_init.h
```diff
@@ -35,9 +35,12 @@ class no_init_vector {
         return size;
     }
 
-    template <int RTYPE, template <class> class StoragePolicy >
-    operator Vector<RTYPE, StoragePolicy>() const {
-        return Rf_allocVector(RTYPE, size) ;
+    template <int RTYPE>
+    operator Vector<RTYPE, PreserveStorage>() const {
```
kevinushey: Why change the template parameters here? Do we need separate operators for each of the storage policies?

kevinushey:
My understanding is that rchk was primarily designed to diagnose C rather than C++ code. I would prefer not making these kinds of changes just to appease the tool unless we really have a good reason.

inst/include/Rcpp/vector/no_init.h
krlmlr: I think `no_init()` should always return a properly protected vector, that's why I removed this template parameter. But I'm not sure.

inst/include/Rcpp/protection/Shield.h
krlmlr: Thanks, done.

krlmlr:
Thanks. I'll address the matrix case and also try to simplify the code more.

I still think we need to return an object with protection, if we keep the `StoragePolicy` template argument we might end up with unprotected objects again.

I don't understand "keep `Rf_allocVector()`": the code still uses `Rf_allocVector()` but just protects it explicitly.

kevinushey:
> We've been asked about these when submitting dplyr 0.7.6

Since CRAN is enforcing these checks I agree we should do what we can to reduce false positives.

inst/include/Rcpp/exceptions.h
```diff
@@ -243,7 +243,7 @@ namespace internal {
     inline bool is_Rcpp_eval_call(SEXP expr) {
         SEXP sys_calls_symbol = Rf_install("sys.calls");
         SEXP identity_symbol = Rf_install("identity");
-        SEXP identity_fun = Rf_findFun(identity_symbol, R_BaseEnv);
+        Shield<SEXP> identity_fun(Rf_findFun(identity_symbol, R_BaseEnv));
         SEXP tryCatch_symbol = Rf_install("tryCatch");
```
kevinushey: Does `rchk` warn about the unprotected use of 'install' here?

inst/include/Rcpp/vector/no_init.h
```diff
@@ -35,9 +35,14 @@ class no_init_vector {
...
+        // Explicitly protect temporary vector to avoid false positive
+        // with rchk (#892)
+        SEXP x = PROTECT(Rf_allocVector(RTYPE, size));
```
kevinushey: What about using `Shield<SEXP>` here rather than raw protect / unprotect?

inst/include/Rcpp/vector/no_init.h
kevinushey: The intention is that the caller can indicate whether they want protection or want -- by default, they get a protected vector; if they don't want one then they can request conversion to the `NoPreserveStorage` variant.

IIUC, since this function was previously returning a raw SEXP, the associated constructor would be called on that return value and that constructor would take care of preserving the underlying storage when required.

eddelbuettel: As I recall, the "feature" of `no_init()` was to save the handful of cycles needed to set the memory to zero, but not to alter anything on protection or lack thereof.

inst/include/Rcpp/vector/no_init.h
krlmlr: Need to check if `rchk` is satisfied.

inst/include/Rcpp/vector/no_init.h
krlmlr: `no_init()` effectively returns an uninitialized vector (via constructing a temporary and calling `operator Vector<...>()` on it). To be on the very safe side, this uninitialized vector needs to be protected during its lifetime; the protection state of the target is of no relevance. This corner case is perhaps irrelevant.

`Vector::Vector(SEXP)` seems to coerce (in the very generic case, not affecting us, but may be hard to tell for `rchk`). Maybe `rchk` thinks (or sees) that this constructor is involved in `Vector<...> x = no_init(...)`?

Happy to revert to the original implementation (with the second template parameter), but I'll check first if `rchk` eats this.

inst/include/Rcpp/exceptions.h
krlmlr: I was confused. No, `rchk` knows that the `Rf_install()` result is protected already, but `Rf_install()` could be an allocation (in theory) which could clean up the result of `Rf_findFun()`.

eddelbuettel:
Err, don't you test locally before sending a commit ?

krlmlr:
Guilty as charged.

eddelbuettel:
I may not be entirely innocent of the practice either but I hate the red outcomes with such a vengeance that I try avoid this.  Nice use of `git commit --amend` though :)

eddelbuettel:
I will set up another reverse-depends check. Not expecting any trouble.  Should be ready to merge.

@kevinushey @jjallaire @thirdwing @nathan-russell  and anybody else: Comments? Concerns?

eddelbuettel:
I canceled and restarted that one, and (of course) just after I canceled it completed.  It had `exit`-ed fine with value of zero but was hanging there...

eddelbuettel:
No new issue in level-one reverse depends check, results at https://github.com/RcppCore/rcpp-logs/commit/1d9d5fafc197be32d8cd23fc1feb162d397e1441

PR #910, opened 2018-10-07
Fix typo EXPOSED -> EXPORT

rstub:


rstub:
Ooops, actually it is the other way round. I will update the PR.

rstub:
BTW, I also get some (pedantic) warnings:

```
expose_class.cpp:4:19: warning: extra ‘;’ [-Wpedantic]
 RCPP_EXPOSED_AS(A);
                   ^
```

Should I remove the `;`?

eddelbuettel:
Thanks for the PR.  We generally like to hear about things _first in an issue ticket_.  So what is this fixing?  Documentation issues only?   Apart from the `-pedantic` ?

rstub:
Only documentation issues: The docs use `RCPP_EXPORT_AS` and `RCPP_EXPORT_WRAP` where it should read `RCPP_EXPOSED_AS` and `RCPP_EXPOSED_WRAP`.

eddelbuettel:
So do I have this right that
i)  two instances of `RCPP_EXPORT_WRAP` are being correcteded to `RCPP_EXPOSED_WRAP` in `vignettes/Rcpp-extending.Rmd`, and
ii) I do not see a file `expose_class.cpp` -- is that your running example?

rstub:
i) yes, plus the same for `RCPP_EXPORT_AS` -> `RCPP_EXPOSED_AS`
ii) yes, that is my example code. Sorry for the confusion.

eddelbuettel:
1.  Sounds good.
2.  But caused by our macro, or caused by you typing an extra `;`?

rstub:
2) Caused by me copying the extra `; ` from the documentation. ;-) The examples for both `RCPP_EXPOSED_AS` and `RCPP_EXPOSED_WRAP` have this extra `;`. And looking at the source for these macros in `inst/include/Rcpp/macros/module.h`,  I think it is not possible to define these macros in a way that makes `; ` after it not superfluous. So my suggestion would be to remove the `; ` from the two relevant lines.

eddelbuettel:
Yep!  Can you add that change to the Rmd as well?

Thanks for catching this (in the process of yet another nice SO answer -- thanks for all those too!) !

Issue #916, opened 2018-10-25
Rcpp::String silently drops embedded NUL bytes


kevinushey:
For example:

`````
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
SEXP embeddedNul() {
  std::string hasNullByte("abc\0abc", 7);
  String converted(hasNullByte);
  return wrap(converted);
}

/*** R
embeddedNul()
*/
`````

Sourcing this gives me:

``` r
> embeddedNul()
[1] "abc"
```

I would hope for either:

1. An error, since we shouldn't silently lose data following the NUL byte; or
2. A conversion to e.g. a raw character vector, preserving the embedded NUL.

It's worth stating that attempting to do this on the R side would give an error, e.g.

``` r
data <- as.raw(c(97, 98, 99, 0, 97, 98, 99))
rawToChar(data)
```

giving

``` r
> data <- as.raw(c(97, 98, 99, 0, 97, 98, 99))
> data
[1] 61 62 63 00 61 62 63
> rawToChar(data)
Error in rawToChar(data) : embedded nul in string: 'abc\0abc'
```

kevinushey:
I believe this occurs because we use `Rf_mkCharCE()` throughout the String classes implementation:

```
$ ag -Q "mkCharCE"
inst/include/Rcpp/String.h
55:        String(): data(Rf_mkCharCE("", CE_UTF8)), buffer(), valid(true), buffer_ready(true), enc(CE_UTF8) {
369:            return valid ? data : Rf_mkCharCE(buffer.c_str(), enc);
398:                data = Rcpp_ReplaceObject(data, Rf_mkCharCE(Rf_translateCharUTF8(data), encoding));
400:                data = Rf_mkCharCE(buffer.c_str(), encoding);
472:                data = Rf_mkCharCE(buffer.c_str(), enc);
```

But since `buffer` itself is just a `std::string`, we should pass both the `c_str()` and the `length()` on and use `Rf_mkCharLenCE()`. I'll try making a PR.

PR #917, opened 2018-10-25
use 'Rf_mkCharLenCE() as appropriate

kevinushey:
Closes https://github.com/RcppCore/Rcpp/issues/916, taking option (1) -- signal an error if an embedded NUL byte is encountered, as R does.

### DIFF
--- a/inst/include/Rcpp/String.h
+++ b/inst/include/Rcpp/String.h
@@ -364,9 +364,15 @@ namespace Rcpp {
         }
 
 
+        inline SEXP get_sexp_impl() const {
+            if (buffer.find('\0') != std::string::npos)
+                throw embedded_nul_in_string();
+            return Rf_mkCharLenCE(buffer.c_str(), buffer.size(), enc);
+        }
+
         inline SEXP get_sexp() const {
             RCPP_STRING_DEBUG_1("String::get_sexp const (valid = %d) ", valid);
-            return valid ? data : Rf_mkCharCE(buffer.c_str(), enc);
+            return valid ? data : get_sexp_impl();
         }
 
         inline SEXP get_sexp() {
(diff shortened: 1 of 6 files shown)


inst/include/Rcpp/String.h
```diff
@@ -395,9 +395,12 @@ namespace Rcpp {
             enc = encoding;
 
             if (valid) {
-                data = Rcpp_ReplaceObject(data, Rf_mkCharCE(Rf_translateCharUTF8(data), encoding));
+                const void* vmax = vmaxget();
```
kevinushey: I noticed that translateCharUTF8() has, in the R sources:

https://github.com/wch/r-source/blob/61132afa07832f4e855da35773fa1341f1a8b481/src/main/sysutils.c#L992-L994

so it's recommended that we manage the stack after usage. Not sure if we should do this in this PR or just leave it out for now though.

kevinushey:
It just occurred to me that we _don't_ want the R error (ie the `longjmp`) so we'll have to do our own error checking before calling `Rf_mkCharLenCE()`.

eddelbuettel:
That sounds like a worthwhile change.

kevinushey:
Yes, sorry -- I'll try to get to it soon.

eddelbuettel:
No worries -- that was the right call.  No rush.

eddelbuettel:
I'm confused.  Did you amend the initial commit in this PR? I am only seeing one commit when there should have been two.

kevinushey:
Yes, I amended the original commit -- I tend to do this on feature branches just to keep the commit history clean. (Although perhaps it's better to just squash and merge rather than retroactively change history on the branch)

eddelbuettel:
Yes, we could settle on just squashing.  I also don't single / non-parallel branches merging back.

In any event here the downside was that your repo co-user, ie me, got lost :).

kevinushey:
Sorry for the confusion :-)

Ultimately, this should be ready to go now.

kevinushey:
It looks like the `h5` package is deprecated, so updated to CRAN may not be as timely as we might like. For that reason I'm going to just work around the issue and use the old-school `Rf_mkCharCE()` when `h5` is detected.

eddelbuettel:
Interesting.  Have we ever done that? Explicitly accomodate other packages?

(And yes, hdf5 package seem to be a in constant state of flux.  Too bad CRAN doesn't just patch...)

kevinushey:
> Interesting. Have we ever done that? Explicitly accomodate other packages?

Not that I can recall, but given that it's one package with a simple workaround, I don't mind having this live in Rcpp if it means existing code depending on `h5` can still work fine. What do you think?

PR #918, opened 2018-10-27
Add new class StreambufSwitcher

heavywatal:
It redirects std::cout/cerr to Rstreambuf during its lifetime. As a trial, I inserted a code to construct an object in `RcppExports.cpp` for package developers. It works as expected, but `std:cout` is detected and complained by `R CMD check`. The definition of `StreambufSwitcher` may need to be move to somewhere outside the scope of the check. I also don't know how to apply this feature to non-package usage like `sourceCpp()`. Therefore, please take this PR as a starting point of discussion.

### DIFF
--- a/inst/include/Rcpp/iostream/Rstreambuf.h
+++ b/inst/include/Rcpp/iostream/Rstreambuf.h
@@ -24,6 +24,7 @@
 
 #include <cstdio>
 #include <streambuf>
+#include <iostream>
 
 namespace Rcpp {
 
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Thanks for filing this. Maybe an initial issue ticket would have been even better, but we have a starting point.  This may not be the the final PR on this, but it gets the ball rolling.

kevinushey:
While this is reasonable, I don't think this will suffice for getting around CRAN checks because regardless R CMD check is still going to see the usages of `cout` / `cerr` in the compiled library, even if they're not used at runtime.

The only 'real' way around these CRAN checks is to ensure that any supporting libraries used / compiled alongside the package just don't use `std::cout` / `std::cerr`.

eddelbuettel:
The time has come close this for inactivity.  As proposed, it can't work.  No harm donw, and we're of course always open to new proposals.

PR #927, opened 2018-12-01
new C++2a plugin for g++ 8 or later

eddelbuettel:
Adds a C++2a plugin given the support in `g++` version 8 and later which is becoming more widespread (eg Ubuntu 18.10, presumably also current Fedora Core).


eddelbuettel:
If nobody is howling I'll merge that later.  Probably prime example of a trivial PR ...

kevinushey:
LGTM!

Issue #934, opened 2019-01-31
Reproducible (minimal ?) need for two compileAttributes() runs


eddelbuettel:
This comes courtesy of a colleague.  If one does

1. Create 'New package with Rcpp' followed by Build->Install+Restart -- OK
2. Add new C++ file using the examples, saved in `src/` as _.e.g._ `src/doubleMe.src, Install+Restart -- OK
3. Now delete the initial src/rcpp_hello_world.cpp, Install+Restart --> Link error

Two runs of `compileAttributes()` fix it, but the IDE does not suggest it.  Can we think of a heuristic to figure out when to run it twice?

coatless:
@kevinushey already produced a reprex c.f. https://github.com/RcppCore/Rcpp/issues/733#issuecomment-314505646

eddelbuettel:
Thanks for that -- I once knew that too.

coatless:
tl;dr of the story: the bug effects ~1% < of the developer base and is quickly solved by re-running hence there never was a push to fix it.

eddelbuettel:
I does frustrate users.  So I am wondering if we should just run the (still very fast) `compileAttributes()` twice.

eddelbuettel:
Closed via #939

PR #939, opened 2019-02-12
fix issue with 'stale' symbols being registered

### LINKED ISSUE #733: potential registration issue when using `_` prefix?
From @krlmlr:

> I'm getting bogus errors that seem to relate to the new registration code. I can't replicate them locally yet, but on Travis CI I'm getting test failures iff I use the all-new registration code (with the underscore prefix). I'd like to suggest to hold the release until we get a better understanding of the problem.
> 
> It's complicated because the failures occur in the DBItest repo which (effectively) installs the RSQLite repo from source. They don't occur inside the RSQLite repo. See [1] for a failing and [2] for a successful build of the same repo, the only difference is that the second build happened with the old Rcpp registration code [3]. In the failing build, the exported symbol declared last (_RSQLite_init_logging) cannot be loaded from the .so file. Does this ring a bell?
> 
> 
> -Kirill
> 
> 
> [1] https://travis-ci.org/rstats-db/DBItest/builds/251777532#L3897
> 
> [2] https://travis-ci.org/rstats-db/DBItest/builds/251929259
> 
> [3] https://github.com/rstats-db/RSQLite/commit/66070b4e

The relevant bit of the error logs, from what I can see:

```
Error in dyn.load(dllfile) : 
  unable to load shared object '/home/travis/build/rstats-db/DBItest/revdep-dev/RSQLite/src/RSQLite.so':
  /home/travis/build/rstats-db/DBItest/revdep-dev/RSQLite/src/RSQLite.so: undefined symbol: _RSQLite_init_logging
Calls: <Anonymous> -> load_dll -> library.dynam2 -> dyn.load
```

The only thing I can divine is that, normally, C-style name mangling will prepend an underscore to the symbol name, so we might expect the symbol to actually be called `__RSQLite_init_logging`. Indeed that's what I see on a local installation:

```
$ nm RSQLite.so  | grep RSQLite
000000000003f360 T _R_init_RSQLite
000000000003e7d0 T __RSQLite_rsqliteVersion
000000000003c0a0 T __RSQLite_rsqlite_bind_rows
000000000003acb0 T __RSQLite_rsqlite_clear_result
000000000003db00 T __RSQLite_rsqlite_column_info
00000000000363b0 T __RSQLite_rsqlite_connect
0000000000039700 T __RSQLite_rsqlite_connection_valid
0000000000039080 T __RSQLite_rsqlite_copy_database
0000000000038a30 T __RSQLite_rsqlite_disconnect
000000000003b300 T __RSQLite_rsqlite_fetch
000000000003ba20 T __RSQLite_rsqlite_get_placeholder_names
000000000003c750 T __RSQLite_rsqlite_has_completed
0000000000039d90 T __RSQLite_rsqlite_import_file
000000000003eda0 T __RSQLite_rsqlite_init_logging
000000000003e150 T __RSQLite_rsqlite_result_valid
000000000003cdd0 T __RSQLite_rsqlite_row_count
000000000003d480 T __RSQLite_rsqlite_rows_affected
000000000003a5d0 T __RSQLite_rsqlite_send_query
```

so it's possible this is something specific to the Travis environment in which @krlmlr was testing. (Note that the package builds, installs, and loads fine for me locally)

kevinushey:
Closes #733, #934.

The underlying issue here IIUC is that the machinery that attempts to discover used symbols was also crawling the various RcppExports files, and since those were 'stale' at the time when `compileAttributes()` is run, they would incorrectly contribute 'old' symbols which would then remain incorrectly registered.

The fix here removes the R-level `RcppExports.R` file (since we want to ensure that `tools::package_native_routine_registration_skeleton()` doesn't touch it), and also removes `RcppExports.cpp` from the set of files we crawl for exportable functions.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2019-02-11  Kevin Ushey  <kevinushey@gmail.com>
+
+	* R/Attributes.R: better handle pre-existing RcppExports.cpp,
+	RcppExports.R when calling compileAttributes()
+
 2018-12-26  Zhuoer Dong  <dongzhuoer@mail.nankai.edu.cn>
 
 	* vignettes/Rcpp-quickref.Rmd: Fix three bugs: use `Named()`, define
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -421,8 +421,14 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
     if (!file.exists(rDir))
         dir.create(rDir)
 
+    # remove the old RcppExports.R file
+    unlink(file.path(rDir, "RcppExports.R"))
+
     # get a list of all source files
     cppFiles <- list.files(srcDir, pattern = "\\.((c(c|pp)?)|(h(pp)?))$", ignore.case = TRUE)
+
+    # don't include RcppExports.cpp
+    cppFiles <- setdiff(cppFiles, "RcppExports.cpp")
 
     # locale independent sort for stable output
     locale <- Sys.getlocale(category = "LC_COLLATE")
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -15,6 +15,9 @@
     \item Changes in Rcpp Attributes:
     \itemize{
       \item A new plugin was added for C++20 (Dirk in \ghpr{927})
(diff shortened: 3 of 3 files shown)


jjallaire:
LGTM

PR #963, opened 2019-04-26
Feature/version string

eddelbuettel:
A simple extension to get the "API" version of the package out.

We have always have `packageVersion("Rcpp")` (or the earlier `packageDescription("Rcpp")$Version`) but as I keep increasing the 'dev version' during development, this leaks eg 1.0.1.2 right now when used by _e.g._ `Rcpp.package.skeleton()`.  Which then bites potential users who would only have CRAN's 1.0.1 or older.  

So I added two new `#define` statements for string versions in the usual `config.h`,  plus a simple function to export to the R side -- where we turn it into `package_version` objects.  Nothing major, but a little help.

@kevinushey This may also help for the RStudio package generator.


R/tools.R
```diff
@@ -48,3 +48,31 @@ asBuildPath <- function(path) {
...
+##' @param devel An logical value indicating if the development or
+##' release version number should be returned, default is release.
+##' @return A \code{package_version} object with either the release
+##' or development version.
+##' @author Dirk Eddelbuettel
+##' @seealso \code{\link{packageVersion}},
+##' \code{\link{Rcpp.package.skeleton}}
+##' @examples getRcppVersion()
+getRcppVersion <- function(devel = FALSE) {
+    rcpp <- .Call("getRcppVersionStrings")[if(devel) 2 else 1]
```
kevinushey: nit: use `.Call("getRcppVersionStrings", PACKAGE = "Rcpp")` to ensure the symbol gets looked up directly in the Rcpp library?

R/tools.R
```diff
@@ -48,3 +48,31 @@ asBuildPath <- function(path) {
...
+    .make_numeric_version(rcpp,
```
kevinushey: Why not just use `package_version()` here?

eddelbuettel: Good point. I think I was looking _inside_ that function and inadvertendly copied it, rather than calling it.

R/tools.R
eddelbuettel: Sure. Look at all the other `.Call()` we do -- it is not too consistent.

eddelbuettel:
@kevinushey All good?

Issue #975, opened 2019-07-09
Setting parenthesis characters as default value to function parameters yields a compiling error


pkrog:
Defining the following function:
```r
Rcpp::sourceCpp(code = '//[[Rcpp::export]]
void foo(const char* a = "(", const char* b= ")") {}')
```
gives:
```
file8586f8153c2.cpp: In function ‘SEXPREC* sourceCpp_63_foo(SEXP)’:
file8586f8153c2.cpp:12:10: error: call of overloaded ‘foo(Rcpp::traits::input_parameter<const char*>::type&)’ is ambiguous
   12 |     foo(a);
      |          ^
file8586f8153c2.cpp:2:6: note: candidate: ‘void foo(const char*, const char*)’
    2 | void foo(const char* a = "(", const char* b= ")") {}
      |      ^~~
file8586f8153c2.cpp:7:6: note: candidate: ‘void foo(const char*)’
    7 | void foo(const char* a);
      |      ^~~
make: *** [/usr/lib64/R/etc/Makeconf:176: file8586f8153c2.o] Error 1
g++ -std=gnu++11 -I"/usr/include/R/" -DNDEBUG   -I"/home/pr228844/R/x86_64-pc-linux-gnu-library/3.6/Rcpp/include" -I"/tmp/RtmpTogNDM/sourceCpp-x86_64-pc-linux-gnu-1.0.1" -D_FORTIFY_SOURCE=2  -fpic  -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt  -c file8586f8153c2.cpp -o file8586f8153c2.o
Error in Rcpp::sourceCpp(code = "//[[Rcpp::export]]\nvoid foo(const char* a = \"(\", const char* b= \")\") {}") : 
  Error 1 occurred building shared library.
```

Inverting the default values, leads to the same error:
```r
Rcpp::sourceCpp(code = '//[[Rcpp::export]]
void foo(const char* a = ")", const char* b= "(") {}')
```
Replacing one of the default values with another value, remove the error:
```r
Rcpp::sourceCpp(code = '//[[Rcpp::export]]
void foo(const char* a = "x", const char* b= "(") {}')
```

kevinushey:
I think this is a bug in the hand-rolled C++ parser used for attributes. Since this is somewhat of an edge case, it's unlikely we'll fix this quickly -- but pull requests are welcome!

eddelbuettel:
[...]

PR #977, opened 2019-07-18
Allow parenthesis and backslash characters inside default values

pkrog:
Solves #975 
The argument parser method `SourceFileAttributesParser::parseArguments()` has been modified in order to allow the use of parenthesis and backslash characters inside quoted strings and character values.

### DIFF
--- a/src/attributes.cpp
+++ b/src/attributes.cpp
@@ -1420,7 +1420,6 @@ namespace attributes {
 
         // Establish the text to parse for the signature
         std::string signature = parseSignature(lineNumber);
-		std::cerr << "-------------------------------- SourceFileAttributesParser::parseFunction 10" << std::endl << signature << std::endl;
         if (signature.empty()) {
             rcppExportNoFunctionFoundWarning(lineNumber);		// #nocov
             return Function();						// #nocov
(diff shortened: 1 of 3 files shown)


eddelbuettel:
Thanks for the PR -- we will take a good look.  I was about to roll up 1.0.2 but this may still fit in.

jjallaire:
This change looks correct to me. @kevinushey Could you also take a look?

src/attributes.cpp
```diff
@@ -1590,44 +1590,65 @@ namespace attributes {
...
-                    case '>':
-                        templateCount--;
-                        break;
-                    case '(':
-                        parenCount++;			// #nocov
-                        break;				// #nocov
-                    case ')':
-                        parenCount--;			// #nocov
-                        break;				// #nocov
+                endOfArg = true;
```
kevinushey: This looks correct to me but IMHO the `endOfArg` flag is a bit more confusing than just using `continue` to bypass  adding characters to the current argument. Is there a reason why you made that switch (perhaps I'm missing a nuance)? Or just style?

pkrog: Hi, I used it when the escape character was still handled with a `char prevChar` variable that was set at the end of the loop. So for consistency, I preferred to replace the `continue` with `endOfArg` in order for `prevChar` to be set even when a comma was met (it was not the case with the `continue` statement, `prevChar` was not updated).
Since I've now changed `prevChar` with the boolean variable `escaped` and `escaped` is set before testing for a comma that marks the end of an argument, we can go back to the `continue` statement. I will make the change.

pkrog: Thanks for pointing this you, by the way.

Issue #976, opened 2019-07-10
Modules suddenly not loading anymore


david-cortes:
I don’t know how to reproduce the error. But roughly speaking, I was compiling some test packages with RCPP_MODULE exposes classes, which at some point failed to compile and left the `.lock` file in my R folder. After deleting that, I cannot get any package with Rcpp modules to work.

E.g. if I create an example module
```r
Rcpp::Rcpp.package.skeleton("mod4", path="/home/me/mypath", module = TRUE)
```

then compile it, generating necessary files along the way, e.g. RcppExports.cpp
```cpp

RcppExport SEXP _rcpp_module_boot_NumEx();
RcppExport SEXP _rcpp_module_boot_yada();
RcppExport SEXP _rcpp_module_boot_stdVector();

static const R_CallMethodDef CallEntries[] = {
    {"_mod4_rcpp_hello_world", (DL_FUNC) &_mod4_rcpp_hello_world, 0},
    {"_rcpp_module_boot_NumEx", (DL_FUNC) &_rcpp_module_boot_NumEx, 0},
    {"_rcpp_module_boot_yada", (DL_FUNC) &_rcpp_module_boot_yada, 0},
    {"_rcpp_module_boot_stdVector", (DL_FUNC) &_rcpp_module_boot_stdVector, 0},
    {NULL, NULL, 0}
};

RcppExport void R_init_mod4(DllInfo *dll) {
    R_registerRoutines(dll, NULL, CallEntries, NULL, NULL);
    R_useDynamicSymbols(dll, FALSE);
}
```
zzz.R
```r
loadModule("NumEx", TRUE)
loadModule("yada", TRUE)
loadModule("stdVector", TRUE)
```

then
```r
library(Rcpp)
library(mod4)
loadModule("yada", TRUE)
a = new("yada")
```

I get
```
Error in getClass(Class, where = topenv(parent.frame())) : 
  “yada” is not a defined class
```

If I try
```r
a = Module("yada")
a$<autocomplete>
```

Then I get
```
Error in Module(module, mustStart = TRUE) : 
  Failed to initialize module pointer: Error in FUN(X[[i]], ...): no such symbol _rcpp_module_boot_yada in package .GlobalEnv
```

(This is despite that the necessary export is right there in `RcppExports.cpp`)

Issue affects all packages. Already tried reinstalling Rcpp.

eddelbuettel:
[...]

PR #982, opened 2019-07-25
Modules vignette review

riccardoporreca:
Review of the Rcpp-modules vignette, with major focus on the usage in packages, as a follow-up of https://github.com/RcppCore/Rcpp/issues/976#issuecomment-512537482

A (not so short) summary of the changes:

- Replaced sparse references to packages across a few existing examples in the _Rpp Modules_ section with a pointer to the specific section at beginning.

- Introduced a new sub-section about usaging `sourceCpp()` as an alternative to `cxxfuntion()` + `getDynLib()`.

- Package section reviewed with focus on recommended loading via `loadModule()`
    - A more practical example has been included
    - The section about the deprecated loading was left as-is
    - Namespace exports are also covered.

- Improvements and harmomization of existing examples in section _Rcpp Modules_
    - General review to make it easier for a user to follow the examples.
    - All examples now assume `fx` (clearly stated) and explicitly call `Module()`, to make them close to self-contained.
    - Include `World` class module in S4 dispatch example for completeness.
    - Make sure all examples (with R code) are runnable, and optionally enable running them
        - Non-included chunks creating the assumed relevant `fx` object based on the C++ code in previous named chunk(s).
        - The document has gained parameters `eval` and `results` in the YAML header to control evaluating the examples on demand (e.g. _Knit with parameters..._).
        - This is pretty non-intrusive and can be easily removed.
    - Running the examples revealed compilation errors with the final `std::vector` example, which I fixed with a **workaround to be discussed**.

- Fixed a few tyops + batch cosmetic alignment:
    - `\proglang` all the way.
    - `\textsl{Rcpp modules}` everywhere but in their own section.
    - Aligned and fixed usage of C++ elements in inline text.

I left a few explicit comments marked as `CHECK-PR` to have some pointers for the PR discussion.

### DIFF
--- a/vignettes/Rcpp-modules.Rmd
+++ b/vignettes/Rcpp-modules.Rmd
@@ -73,12 +73,6 @@ skip_final_break: true
 # Produce a pinp document
 output: pinp::pinp
 
-# CHECK-PR: Keep the option to run the examples? (look for params$eval)
-# Parameters defining the evaluation behavior (chunk options) 
-params:
-    eval: FALSE
-    results: "hide"
-
 header-includes: >
   \newcommand{\proglang}[1]{\textsf{#1}}
   \newcommand{\pkg}[1]{\textbf{#1}}
@@ -94,20 +88,6 @@ vignette: >
   %\VignetteEngine{knitr::rmarkdown}
   %\VignetteEncoding{UTF-8}
 ---
-
-```{r setup, include=FALSE}
-# disable evaluation globally
-knitr::opts_chunk$set(eval = FALSE)
-if (params$eval) {
-    library(Rcpp)
-    library(inline)
-    # convenience wrapper to extract code from chunks and call cxxfunction()
-    chunk_cxxfunction <- function(chunks) {
-        inc <- unlist(knitr::knit_code$get(chunks))
-        cxxfunction(signature(), plugin = "Rcpp", include = inc)
-    }
-}
-```
 
 # Motivation
 
@@ -269,7 +249,7 @@ DE 21 Sep 2013: there must a bug somewhere in the vignette processing
              hence shortened example to not show code again
 -->
 
-```{r}
+```{r, eval=FALSE}
 f1 <- cxxfunction( , "", includes = unifModCode,
                   plugin = "Rcpp" )
 getDynLib(f1)  ## will display info about 'f1' 
@@ -278,7 +258,7 @@ getDynLib(f1)  ## will display info about 'f1'
 The following listing shows some \textsl{manual} wrapping to access the code,
 we will see later how this can be automated:
 
-```{r}
+```{r, eval=FALSE}
 setClass("Uniform",
          representation( pointer = "externalptr"))
 
@@ -358,7 +338,7 @@ the need for a wrapper function using either \pkg{Rcpp} or the \proglang{R} API.
 On the \proglang{R} side, the module is retrieved by using the
 `Module` function from \pkg{Rcpp}
 
-```{r, eval=params$eval, results=params$results}
+```{r, eval=FALSE}
 inc <- '
 using namespace Rcpp;
 
@@ -389,7 +369,7 @@ as described in Section \ref{sec:modules-sourceCpp}.
 A module can contain any number of calls to `function` to register
 many internal functions to \proglang{R}. For example, these 6 functions:
 
-```{Rcpp yada-fun-def}
+```{Rcpp yada-fun-def, eval=FALSE}
 std::string hello() {
     return "hello";
 }
@@ -417,7 +397,7 @@ void bla2( int x, double y) {
 
 can be exposed with the following minimal code:
 
-```{Rcpp yada-fun-module}
+```{Rcpp yada-fun-module, eval=FALSE}
 RCPP_MODULE(yada) {
     using namespace Rcpp;
 
@@ -429,13 +409,10 @@ RCPP_MODULE(yada) {
     function("bla2"  , &bla2 );
 }
 ```
-```{r include=FALSE, eval=params$eval, results=params$results}
-fx <- chunk_cxxfunction(c("yada-fun-def", "yada-fun-module"))
-```
 
 which can then be used from \proglang{R}:
 
-```{r eval=params$eval, results=params$results}
+```{r eval=FALSE}
 yada <- Module("yada", getDynLib(fx))
 yada$bar(2L)
 yada$foo(2L, 10.0)
@@ -462,7 +439,7 @@ are:
 In addition to the name of the function and the function pointer, it is possible
 to pass a short description of the function as the third parameter of `function`.
 
-```{Rcpp norm-doc}
+```{Rcpp norm-doc, eval=FALSE}
 using namespace Rcpp;
 
 double norm(double x, double y) {
@@ -477,10 +454,7 @@ RCPP_MODULE(mod) {
 
 The description is used when displaying the function to the \proglang{R} prompt:
 
-```{r include=FALSE, eval=params$eval, results=params$results}
-fx <- chunk_cxxfunction("norm-doc")
-```
-```{r, eval=params$eval, results=params$results}
+```{r, eval=FALSE}
 mod <- Module("mod", getDynLib(fx))
 show(mod$norm)
 ```
@@ -491,7 +465,7 @@ show(mod$norm)
 of the \proglang{R} function that encapsulates the \proglang{C++} function, by passing
 a `Rcpp::List` after the function pointer.
 
-```{Rcpp mod_formals}
+```{Rcpp mod_formals, eval=FALSE}
 using namespace Rcpp;
 
 double norm(double x, double y) {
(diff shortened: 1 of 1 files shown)


riccardoporreca:
More on the issue with the **`std::vector` example**.

The following is  a minimal example
```cpp
#include <Rcpp.h>
typedef std::vector<double> vec;
// Free function wrappers to work around errors
// error: call of overloaded
// ‘method(const char [7], <unresolved overloaded function type>)’ is ambiguous
void vec_resize (vec* obj, int n) {
    obj->resize(n);
}
// error: no matching function for call to
// ‘Rcpp::class_<std::vector<double> >::method(const char [10], <unresolved overloaded function type>)’
void vec_push_back (vec* obj, double value) {
    obj->push_back(value);
}
RCPP_MODULE(mod_vec) {
    using namespace Rcpp;
    class_<vec>("vec")
    .constructor<int>()
    // exposing member functions
    .method("resize", &vec::resize)
    // .method("resize", &vec_resize)
    .method("push_back", &vec::push_back)
    // .method("push_back", &vec_push_back)
    ;
}
```
With compilation errors (Ubuntu 18.04, compiler details: `g++ -std=gnu++11`). This are probably errors introduced with C++11 where both [`push_back`](http://www.cplusplus.com/reference/vector/vector/push_back/) and [`resize`](http://www.cplusplus.com/reference/vector/vector/resize/) are overloaded compared to C++11
```cpp
void push_back (const value_type& val);
void push_back (value_type&& val);
void resize (size_type n);
void resize (size_type n, const value_type& val);
```
This does not seem to be supported by `.const_method` and `.nonconst_method` in **Rcpp**

eddelbuettel:
Lovely stuff.  I will make some time to review.   Trying to get some other things out the door first so may take a few days.  But first off ... big thanks for this!

vignettes/Rcpp-modules.Rmd
```diff
@@ -1059,74 +1136,183 @@ RCPP_MODULE(mod_vec) {
...
+Section \ref{sec:modules-sourceCpp} above
 
-Note: This approach is deprecated as of Rcpp 0.12.5, and now triggers a warning
-message.  Eventually this function will be withdrawn.
+Then, `loadModule` is called in the package's \proglang{R} code to expose all
+\proglang{C++} functions and classes as objects `hello`, `bla`, `bla2`, `World`
+into the package namespace:
+
+```{r}
+loadModule("yada", TRUE)
```
coatless: Perhaps switch this to a namespace function call?

```suggestion
Rcpp::loadModule("yada", TRUE)
```

vignettes/Rcpp-modules.Rmd
```diff
@@ -1059,74 +1136,183 @@ RCPP_MODULE(mod_vec) {
...
+
+# Using modules in other packages {#sec:package}
+
+## Namespace import
+
+When using \textsl{Rcpp modules} in a packages, the client package needs to
 import \pkg{Rcpp}'s namespace. This is achieved by adding the
 following line to the `NAMESPACE` file.
 
+<!-- CHECK-PR: is this chunk needed at all? -->
```
coatless: This was styling for the commands. See previous remark.

vignettes/Rcpp-modules.Rmd
```diff
@@ -1059,74 +1136,183 @@ RCPP_MODULE(mod_vec) {
...
 ```{r, echo=FALSE,eval=TRUE}
 options( prompt = " ", continue = " " )
 ```
 
-```{r, eval=FALSE}
+```{r}
```
coatless: I feel like there was also a need to add:

```r
import(methods)
```

vignettes/Rcpp-modules.Rmd
```diff
@@ -1018,6 +1085,14 @@ Rcpp::NumericVector vec_asR( vec* obj ) {
 void vec_set(vec* obj, int i, double value) { 
    obj->at( i ) = value; 
 }
+// CHECK-PR: workaround complilation error with vecL:resize (C++11)
```
coatless: I'll try to investigate this a bit later.

riccardoporreca:
@eddelbuettel, @coatless, I was happy to provide my small contribution after being a happy user for quite some time.

eddelbuettel:
This is really good work, and I like it a lot.  I didn't get to _running_ it though.

I think we should work out what we want the sections / questions you labeled as 'CHECK-PR' to say before we merge.

eddelbuettel:
I truly apologize for taking so long.  Life gets in the way at times.  

My verdict is mostly two thumbs way up.  I would just ask for one more pass, removing all the CHECK-PR as detailed below.   

I may still have overlooked some issues, but my quick notes are below.


### Comments:

- CHECK-PR on parameters:  I'd say remove, in-line with what all other vignettes do.  Not a
  bad idea though.  (We do conditional evaluation of chunks in the drat-data
  paper in the R Journal, so the idea is fine in principle.)

- So let's restore evaluation as before

- All the typo fixes are excellent. Thank you so much!

- Ditto with all the editing fixes. Very attentive.

- CHECK-PR on resize: Likely name clash and good idea to work around. So we
  should keep this and replace 'CHECK-PR' with something like 'FixForC++11'
  
- CHECK-PR on mustStart=TRUE:  Don't know

- CHECK-PR on sourceCpp: Excellent. Absolutely keep. Nicely done.

- CHECK-PR on chunk needed:  That looks like a left-over style option from
  JSS or something.  Not needed I'd say.

riccardoporreca:
@eddelbuettel, @coatless, thanks for your feedback.

Back to connected life after a long weekend, I have addressed the CHECK-PR points in 9832314 as suggested above, with details in the commit comment.

I have rebased the branch, although squash-merging might actually be a good idea.

vignettes/Rcpp-modules.Rmd
```diff
@@ -73,6 +73,12 @@ skip_final_break: true
 # Produce a pinp document
 output: pinp::pinp
 
+# CHECK-PR: Keep the option to run the examples? (look for params$eval)
+# Parameters defining the evaluation behavior (chunk options)
```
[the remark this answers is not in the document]
riccardoporreca: Restored original evaluation in 9832314 also based on @eddelbuettel's feedback.

vignettes/Rcpp-modules.Rmd
```diff
@@ -166,9 +186,9 @@ double norm(double x, double y) {
 ```
 
 See the corresponding vignette \citep{CRAN:Rcpp:Attributes} for details, but
-read on for \textsl{Rcpp modules} which contains to provide features not
+read on for \textsl{Rcpp modules} which provide features not
 covered by \textsl{Rcpp attributes}, particularly when it comes to binding
-entire C++ classes and more.
+entire \proglang{C++} classes and more.
```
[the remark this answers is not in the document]
riccardoporreca: There was actually a mixed usage of \proglang{}, not it should be consistent (at least within this vignette).

vignettes/Rcpp-modules.Rmd
```diff
@@ -1059,74 +1136,183 @@ RCPP_MODULE(mod_vec) {
...
 v$push_back(10)
 v$size()
+v$resize(30L)
 v$capacity()
 v[[ 0L ]]
 v$as.vector()
 ```
 
-# Using modules in other packages {#sec:package}
+<!-- CHECK-PR: sourceCpp is pretty convenient, OK to add as new section? -->
```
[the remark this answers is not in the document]
riccardoporreca: Reference to the vignette added in 9832314.

vignettes/Rcpp-modules.Rmd
```diff
@@ -1137,64 +1323,65 @@ with a default value of `TRUE`. With this default value, all content
...
-```{r, eval=FALSE}
+```{r}
 yada <- Module( "yada" )
 
 .onLoad <- function(libname, pkgname) {
     # placeholder
 }
 ```
 
+<!-- CHECK-PR: is this chunk needed at all? -->
```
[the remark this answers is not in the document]
riccardoporreca: Removed in 9832314 also based on @eddelbuettel's similar feedback

vignettes/Rcpp-modules.Rmd
riccardoporreca: This is surely a valid point, although this approach does not seem to be used across other Rcpp vignettes.

vignettes/Rcpp-modules.Rmd
riccardoporreca: I double-checked using the package produced by `Rcpp::Rcpp.package.skeleton()`, removed the DESCRIPTION and NAMESPACE imports of `methods` and things seem to still work.

vignettes/Rcpp-modules.Rmd
riccardoporreca: I actually just realized this fix was part is part of one of the skeleton files since a long time https://github.com/RcppCore/Rcpp/blob/4ec861775601ab9c6fcad3fd03d4b84ba88814b5/inst/skeleton/stdVector.cpp#L45-L46

vignettes/Rcpp-modules.Rmd
eddelbuettel: Thanks for checking that.  I will merge this PR but that whole aspect may be worth looking into in another ticket.  This changed over the years, and maybe describing just one simple (working) setup is best.

PR #995, opened 2019-10-01
disable inclusion of modules code with RCPP_NO_MODULES

### LINKED ISSUE #993: allow users to disable inclusion of modules code
The modules code in Rcpp (especially the 'code bloat') is responsible for a large chunk of the time spent during compilation, as the compiler has to do a lot of work to parse all of the generated functions.

What would we think about adding a define `RCPP_NO_MODULES` (similar to `RCPP_NO_SUGAR`) that would allow users to disable inclusion of the modules code, for cases where they know they don't need it?

kevinushey:
Closes #993.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2019-09-30  Kevin Ushey  <kevinushey@gmail.com>
+
+	* inst/include/Rcpp.h: add RCPP_NO_MODULES
+	* inst/include/Rcpp/api/meat/is.h: Idem
+	* inst/include/Rcpp/api/meat/meat.h: Idem
+
 2019-09-28  Dirk Eddelbuettel  <edd@debian.org>
 
 	* README.md: Add a stackoverflow tag to indicate where to ask questions,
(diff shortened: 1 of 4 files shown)


eddelbuettel:
What did you timing tests reveal in terms of net change? 10%? 20%? More?

kevinushey:
Using `roxygen2` as an example:

```
# regular build ----
$ time R CMD INSTALL --preclean roxygen2 &> /dev/null

real    0m13.648s
user    0m12.796s
sys     0m0.773s

# RCPP_NO_MODULES ----
$ time R CMD INSTALL --preclean roxygen2 &> /dev/null

real    0m9.479s
user    0m8.800s
sys     0m0.594s
```

So, almost 50% faster, which isn't too shabby.

eddelbuettel:
Ok, took a moment to go over rev.deps as some unrelated change (possibly in package `sf`) seems to have invalidated a bunch of the spatial packages and their tests failed -- nothing to do with us, mostly just needing a recompilation on that box it seems.  Will merge in due course.

PR #1002, opened 2019-10-19
fix #968 with a single argument XPtr constructor

### LINKED ISSUE #968: Convert SEXP to XPtr cause lost of tag value
Here is a minimal example:

```cpp
#include <Rcpp.h>
using namespace Rcpp;

typedef struct {
    int x;
    int y;
} mycomplex;

void myfinalizer(mycomplex* x) {
    free(x);
}

typedef XPtr<mycomplex, Rcpp::PreserveStorage, myfinalizer> XPtrMyComplex;

// [[Rcpp::export]]
SEXP init(int a, int b) {
    mycomplex* m = (mycomplex*) malloc(sizeof(mycomplex));
    m->x = a;
    m->y = b;
    XPtrMyComplex ans = XPtrMyComplex(m, true, Rf_ScalarRaw(42), R_NilValue);
    return ans;
}

// [[Rcpp::export]]
int access(SEXP robj) {
    XPtrMyComplex obj = as<XPtrMyComplex>(robj);
    mycomplex* p = obj.checked_get();
    return p->x + p->y;
}


/*** R
library(xptr)
x <- init(3, 4)
print(xptr_tag(x))
access(x)
print(xptr_tag(x))
*/
```

I get:

```
R> library(xptr)

R> x <- init(3, 4)

R> print(xptr_tag(x))
[1] 2a

R> access(x)
[1] 7

R> print(xptr_tag(x))
NULL
```

I am not sure if this behavior is expected. Also, I would appreciate if anyone can explain to me the semantics when doing this conversion: is a new external pointer object allocated or it is just wrapper around the original one?

Thanks!

stephematician:
`as<>` will no longer modify tags/protected data of an existing external pointer. Modifying the tags/protected data of an external pointer via the `XPtr(SEXP,SEXP,SEXP)` constructor is made more explicit by removing the two default arguments, and `XPtr(SEXP)` is implemented separately (as explicit) and does not modify the external pointer's tags/protected data.

Added a self-tag test which passed here. I was unable to test if downstream packages are affected by this change.


eddelbuettel:
Lovely!  

>  was unable to test if downstream packages are affected
I do and will.  Thanks for the PR!

eddelbuettel:
Do you want to drop a short paragraph in [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog), or do you want me to?

stephematician:
> Do you want to drop a short paragraph in [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog), or do you want me to?

I'll give it a go.

eddelbuettel:
Thanks, If you use The One Editor Everybody Should Use (kidding) it just `C-x 4 a` from the file you want to record.  Everybody else gets to copy and paste.  Two spaces around name and email.

stephematician:
I'm hoping a PR from a vim user is still acceptable - let me know if any other changes needed.

stephematician:
Oh no - it added my .swp file :/ Reject!

eddelbuettel:
That's what you get for using A Very Bad Editor ;-)

eddelbuettel:
Kidding aside the single worst Unix command known to man is `git add` in wildcard mode.

Of course, those of use a "that" editor have the undeescribably awesome Magit mode....

And please reopen and just a commit to delete the .swp file.

PR #1003, opened 2019-10-19
fix #968 with a single argument XPtr constructor

### LINKED ISSUE #968: Convert SEXP to XPtr cause lost of tag value

stephematician:
Cleaned up pull request 1002

as<> will no longer modify tags/protected data of an existing external pointer. Modifying the tags/protected data of an external pointer via the XPtr(SEXP,SEXP,SEXP) constructor is made more explicit by removing the two default arguments, and XPtr(SEXP) is implemented separately (as explicit) and does not modify the external pointer's tags/protected data.

Added a self-tag test which passed here. I was unable to test if downstream packages are affected by this change.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2019-10-19  Stephen Wade  <stephematician@gmail.com>
+
+	* inst/include/Rcpp/XPtr.h: XPtr constructor split up, a single
+	argument does not modify tags and protected data of the external pointer
+	* inst/unitTests/cpp/Xptr.cpp: Added test
+	* inst/unitTests/runit.XPtr.R: Idem
+
 2019-10-14  Dirk Eddelbuettel  <edd@debian.org>
 
 	* README.md: Added CRAN + BioConductor badges for reverse depends,
--- a/inst/include/Rcpp/XPtr.h
+++ b/inst/include/Rcpp/XPtr.h
@@ -63,13 +63,22 @@ public:
      *
      * @param xp external pointer to wrap
      */
-    explicit XPtr(SEXP x, SEXP tag = R_NilValue, SEXP prot = R_NilValue) {
+    explicit XPtr(SEXP x) {
         if (TYPEOF(x) != EXTPTRSXP) {
             const char* fmt = "Expecting an external pointer: [type=%s].";
             throw ::Rcpp::not_compatible(fmt, Rf_type2char(TYPEOF(x)));
         }
-
         Storage::set__(x);
+    };
+
+    /**
+     * constructs a XPtr wrapping the external pointer (EXTPTRSXP SEXP)
+     *
+     * @param xp external pointer to wrap
+     * @param tag tag to assign to external pointer 
+     * @param prot protected data to assign to external pointer
+     */
+    explicit XPtr(SEXP x, SEXP tag, SEXP prot) : XPtr(x) {
         R_SetExternalPtrTag( x, tag);
         R_SetExternalPtrProtected(x, prot);
     };
--- a/inst/unitTests/cpp/XPtr.cpp
+++ b/inst/unitTests/cpp/XPtr.cpp
@@ -47,6 +47,17 @@ int xptr_2( XPtr< std::vector<int> > p){
 }
 
 // [[Rcpp::export]]
+void xptr_self_tag( XPtr< std::vector<int> > p ){
+    XPtr< std::vector<int> > self_tag(wrap(p), wrap(p), R_NilValue) ;
+}
+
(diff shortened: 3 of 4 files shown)


eddelbuettel:
Somewhat, opening a new pull request is worse than cleaning the previous one (as it leads to issue inflation).  All you needed was one `git rm` and add and commit and push.  Too late now...

stephematician:
> Somewhat, opening a new pull request is worse than cleaning the previous one (as it leads to issue inflation). All you needed was one `git rm` and add and commit and push. Too late now...

Roger that, apologies that you had to be the one to suffer my first PR in a long time.

eddelbuettel:
No worries, after all it is you has to spend the rest of your life with the shame of a close Rcpp pull request ;-)  

And thanks again for the fix.  I am running reverse-depends on a seriously underpowered machine (but hey, free courtesy access on another continent so cannot really complain and am in fact rather grateful for having it) so it will be a few hours but zero issues through sixty packages so far.

eddelbuettel:
@artemklevtsov Do you want to send me (or include here) a two-line diff?

artemklevtsov:
```diff
# inst/unitTests/cpp/XPtr.cpp
+ return wrap(p) == (p).tag();
- return wrap(p) == R_ExternalPtrTag(p);
```
But I agree that `R_ExternalPtrTag` variant is more clear.

eddelbuettel:
Sorry, when I say `diff` I mean something I can pass to `patch`  and which fully and clearly specifies which _file(s_ and _line(s)_ are changed _how_.  This only does the latter and leaves the other two unclear.

eddelbuettel:
@stephematician Slight trouble.  On win-builder and with the older compiler I get

```
 D:/RCompile/CRANguest/R-devel/lib/Rcpp/include/Rcpp/XPtr.h: In constructor 'Rcpp::XPtr<T, StoragePolicy, Finalizer, finalizeOnExit>::XPtr(SEXP, SEXP, SEXP)':
  D:/RCompile/CRANguest/R-devel/lib/Rcpp/include/Rcpp/XPtr.h:81:56: warning: delegating constructors only available with -std=c++11 or -std=gnu++11
       explicit XPtr(SEXP x, SEXP tag, SEXP prot) : XPtr(x) {
```

and we have so far resisted the siren song of turning C++11 on in Rcpp itself.  

Maybe we should condition part of this and/or keep the older version around?

eddelbuettel:
> Maybe we should condition part of this and/or keep the older version around?

Which is what I just did and will test.  Your new change will be in effect if and only if we have C++11 or better (_i.e._ almost everywhere) and I stuck the previous `ctor` back in as a fallback for windoze and other old systems.

stephematician:
Ah - right delegating constructors. Apologies for my slow response. We could move the work that the (delegating) constructor was doing into a private member, and call that in the body of both `XPtr(SEXP)` and `XPtr(SEXP, SEXP, SEXP)`. That way there's no C++11-specific syntax, and we would gain consistency in XPtr across older and newer systems (although this seems like a real niche issue). I'll submit a PR along those lines, but otherwise I'm happy with the changes you've implemented.

eddelbuettel:
Oh, didn't think of that trick.  Could do; or else we keep it simple and leave it as is ...

PR #1009, opened 2019-10-28
two (short) new entries for Rcpp-FAQ vignette

eddelbuettel:
Two simple additions which provide a start.  

@kevinushey @jjallaire if you have comments / extensions maybe commit straight into branch?


vignettes/rmd/Rcpp-FAQ.Rmd
```diff
@@ -1720,3 +1737,11 @@ elsewhere.
 So if your autogenerated file fails, and a `symbols not found` error is reported
 by the linker, consider running `compileAttributes()` twice. Deleting
 `R/RcppExports.R` and `src/RcppExports.cpp` may also work.
+
+## Can we use exceptions and stop() across shared libraries?
+
+No, sadly.  A known limitation coming from the operating system, Windows in
```
jjallaire: I think this might in it's present form be too cautionary. In all Rcpp attribute generated code (including the export of C++ interfaces to other packages) we handle this gracefully. The only place I'm aware of in Rcpp where this bites is specifically in Rcpp Module methods on Windows only. I think we might be better off just adding docs to modules saying you can't call stop() from within methods (they should always return something even if just an error object).

eddelbuettel: Hm, I thought we still bound by the general restriction of not being able to throw/catch across distinct DLLs?  But maybe I am indeed to pessimistic in that view as it would blow up the use case of, eg, RQuantLib catching QuantLib errors. 

Easy to rewrite as "Mostly.  A known remaining limitation ..."   I can take a stab at it later.

jjallaire:
I'm not sure about the RQuantLib case but I'd say given we haven't really
heard about this save for modules that perhaps there is some other
mediation in place there?

eddelbuettel:
Yes I meant that in general it just works (at least on not-on-Windows).  I.e. at work we have a large-ish standard C++ codebase, exceptions are generally forbidden and it asserts and exits on error so for the  calling-from-R-case I redefine these as Rcpp::stop() and it works just fine. 

Just not sure about Windows which I have the benefit of mostly not seeing anymore ....

vignettes/rmd/Rcpp-FAQ.Rmd
kevinushey: Cross-DLL exceptions work as long as:

1. You are linking to a non-static C++ standard library;
2. You are using the same standard library in all compiled objects.

Unfortunately, 1. is not true on Windows as R opts to link to libstdc++ statically; there was some discussion about whether this could be changed for Rtools 4.0 but it sounds like that won't happen.

In some cases, 2. is also not true -- e.g. on macOS users might end up mixing the version of libc++ bundled with R in some dylibs, and the system version of libc++ in others. Usually this means that the exception type info isn't properly handled across module boundaries.

kevinushey: (and of course, Rcpp attributes help us as, if one calls an attributes-generated function, any C++ exceptions are effectively shielded by try-catch and so are not propagated into a separate DLL)

eddelbuettel: @jjallaire @kevinushey Thanks for the very helpful comments.  

I just rebased and force-pushed this branch adding one refining commit.  The diff in the commit now shows an expanded and refined answer to this new FAQ entry based on your comments.

eddelbuettel:
@kevinushey @jjallaire if/when you have a chance glance over the current state of the new paragraphs and especially the updated / rewritten second one reflecting your really useful feedback.

Otherwise looks like we are good to go. May toss this at CRAN soon as 1.0.3.

eddelbuettel:
Ok, I will take "no further comment" in the generous sense of "no further complaints" :grinning:

Merging now and going to prepare a 1.0.3 release.

PR #1011, opened 2019-11-02
Feature/add more shields

eddelbuettel:
As discussed in #1010, a few more unprotected temporaries were floating around.  This attempts to cover a few more.

@kevinushey give it a look when you have a moment

### DIFF
--- a/inst/include/Rcpp/Environment.h
+++ b/inst/include/Rcpp/Environment.h
@@ -2,8 +2,8 @@
 //
 // Environment.h: Rcpp R/C++ interface class library -- access R environments
 //
-// Copyright (C) 2009 - 2013    Dirk Eddelbuettel and Romain Francois
-// Copyright (C) 2014           Dirk Eddelbuettel, Romain Francois and Kevin Ushey
+// Copyright (C) 2009 - 2013  Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2014 - 2019  Dirk Eddelbuettel, Romain Francois and Kevin Ushey
 //
 // This file is part of Rcpp.
 //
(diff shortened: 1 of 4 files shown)


inst/include/Rcpp/Environment.h
```diff
@@ -246,9 +246,8 @@ namespace Rcpp{
                        we have to go back to R to do this operation */
                     SEXP internalSym = Rf_install( ".Internal" );
                     SEXP removeSym = Rf_install( "remove" );
-                    Shield<SEXP> call( Rf_lang2(internalSym,
-                            Rf_lang4(removeSym, Rf_mkString(name.c_str()), Storage::get__(), Rf_ScalarLogical( FALSE ))
-                        ) );
+                    Shield<SEXP> str(Rf_mkString(name.c_str()));
+                    Shield<SEXP> call(Rf_lang2(internalSym, Rf_lang4(removeSym, str, Storage::get__(), Rf_ScalarLogical(FALSE))));
```
kevinushey: It may also be worth pulling out `Rf_lang4()` from this call and protecting that as well, since IIUC that will be an unprotected R object being passed into `Rf_lang2()`, and so _in theory_ could be cleaned up by the GC.

kevinushey: Hmm, I might be wrong. IIUC, each of the `lang*` helper functions protects the arguments passed in, before the final call is created:

https://github.com/wch/r-source/blob/eafae2b2ed39b3383943cd6c36f92d78f966811e/src/include/Rinlinedfuns.h#L629-L677

So this pattern should in fact be safe.

eddelbuettel:
For completeness, the reverse dependsn check results are at https://github.com/RcppCore/rcpp-logs/commit/74e007b2d914f9f49e2e4b7de0f5fcf8a350e9d4

PR #1016, opened 2019-11-09
Update Attributes.R (closes #1017)

### LINKED ISSUE #1017: cppFunction won't compile
Hi,
There is an issue with `cppFunction` in that it won't compile if you require multiple libraries. A minimal reproducible example is below. (I appreciate that this example doesn't actually require the `BH` and `RcppArmadillo` header files, but the presence of the former means that the header file for latter isn't linked.)

```
library(Rcpp)
cppFunction('arma::mat timesTwo(arma::mat x) {
    return x * 2.0;}', depends = c("BH", "RcppArmadillo"))
```

Running the above example causes a compilation error. I have raised a pull request (#1016 ) that fixes this. The result of me running `sessionInfo()` is:
```
R version 3.5.2 (2018-12-20)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 19.04

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.8.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.8.0

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8     LC_MONETARY=en_GB.UTF-8   
 [6] LC_MESSAGES=en_GB.UTF-8    LC_PAPER=en_GB.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
[1] Rcpp_1.0.3

[1] compiler_3.5.2            tools_3.5.2               RcppArmadillo_0.9.800.1.0 BH_1.69.0-1 
```

Many thanks,

TJ

tjmckinley:
When passing multiple "depends" arguments to "cppFunction" e.g.
```
cppFunction(code, depends = c("package1", "package2"))
```
this was producing code with scaffolding:
```
// [[Rcpp:depends(package1)]]
// [[Rcpp:depends(package2)]]
#include <package1.h>
#include <Rcpp.h>
```
rather than
```
// [[Rcpp:depends(package1, package2)]]
#include <package1.h>
#include <package2.h>
#include <Rcpp.h>
```

### DIFF
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -244,7 +244,7 @@ cppFunction <- function(code,
 
     # process depends
     if (!is.null(depends) && length(depends) > 0) {             # #nocov start
-        depends <- paste(depends, sep=", ")
+        depends <- paste(depends, collapse=", ")
         scaffolding <- paste("// [[Rcpp::depends(", depends, ")]]", sep="")
         scaffolding <- c(scaffolding, "", .linkingToIncludes(depends, FALSE),
                          recursive=TRUE)                        # #nocov end


tjmckinley:
No worries. Thanks to all for producing such a useful package!

coatless:
@tjmckinley while you're at it... Mind fixing the code for `plugins` as well? c.f.

https://github.com/RcppCore/Rcpp/blob/975361f40ee5140e3c843be365a20070edc63bfb/R/Attributes.R#L258-L259

eddelbuettel:
@tjmckinley Thank you. Please also consider reading [Contributing](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) before sending PRs.

What is the motivation for the PR?  What it fixes is not a defect but an aesthetic preference.  Which would change something used as it currently is and was by lots of people. We generally don't do that without good reason.

tjmckinley:
My apologies @eddelbuettel. I will log an issue now with a reprex.

tjmckinley:
Hi @eddelbuettel. I've logged an issue (#1017) with a reprex. It's not purely aesthetic in this case since the code won't compile for the example since the header file for `RcppArmadillo` is not found. For example, running a function of the form
```
cppFunction(code, depends = c("package1", "package2"))
```
produces code with scaffolding:
```
// [[Rcpp:depends(package1)]]
// [[Rcpp:depends(package2)]]
#include <package1.h>
#include <Rcpp.h>
```
with no `#include <package2.h>`. This is caused simply by the `cppFunction()` code not parsing the `depends` argument correctly.

Many thanks,

TJ

tjmckinley:
By the way, many thanks @eddelbuettel, @coatless for producing this fantastic package. I can't do without it! Apologies also if I take a while to reply. I'm in Japan and hence the time difference causes a lag in response time.

tjmckinley:
> @tjmckinley while you're at it... Mind fixing the code for `plugins` as well? c.f.
> 
> https://github.com/RcppCore/Rcpp/blob/975361f40ee5140e3c843be365a20070edc63bfb/R/Attributes.R#L258-L259

Yes @coatless. Happy to do this if you and @eddelbuettel think this is an issue.

jjallaire:
LGTM as well.

eddelbuettel:
Looks good to me too after a closer look; merging.

PR #1022, opened 2019-11-12
new test for version number consistency (fixes #1021)

### LINKED ISSUE #1021: consider updating RCPP_VERSION etc. automatically at build time
To help avoid things like https://github.com/RcppCore/Rcpp/issues/1019. It'd be nice to automate this (e.g. with a `configure` script) so that these fields are always populated as appropriate based the version in the DESCRIPTION file.

I don't think we need a full-blown autoconf script; rather, we could just use a simple `configure` / `configure.win` script to build e.g. `config.h` from `config.h.in`.

If I can toot my own horn, I would propose using [configure](https://github.com/kevinushey/configure) to make this happen. Note that this would not require any explicit dependency on the configure package; the necessary infrastructure would be bundled into Rcpp itself.

Otherwise, it'd be simple to (say) have `configure` / `configure.win` scripts which merely sourced a helper R script that manages the configure process.

Let me know if this sounds worth doing.

eddelbuettel:
As mentioned in #1021, the entries in `config.h` can become disconnected from `packageVersio()`. This PR ties them back together via a simple set of unit tests.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2019-11-11  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/unitTests/runit.packageversion.R: New test
+	* inst/unitTests/cpp/rcppversion.cpp: Cpp portion of test 
+
 2019-11-10  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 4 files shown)


eddelbuettel:
We can still experiment with a helper script on top keeping `DESCRIPTION` and `config.h` tied, but at least now we have checks to go with missing balances :)

PR #1023, opened 2019-11-13
Small update to XPtr removing C++11 syntax following PR #1012

stephematician:
Removes the delegating constructor call in the XPtr constructor, thus the constructors no longer rely on C++11 syntax and still maintain the behaviour in PR #1003. The calls to Storage::__set are moved into the new checked_set (private) member of XPtr, and checked_set is called by both constructors.

I've tested this with newer Rtools (3.3 and 3.4 via rhub) and some linux builds (gcc 5), but not older Rtools - so can't 100% confirm it will build on older gcc.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2019-11-13  Stephen Wade  <stephematician@gmail.com>
+
+	* inst/include/Rcpp/XPtr.h: Delete delegating constructor (follow-up #1012)
+	* inst/unitTests/runit.XPTr.R (test.XPtr): Reinsert self-tag test on Windows
+
 2019-11-11  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/unitTests/runit.packageversion.R: New test
--- a/inst/include/Rcpp/XPtr.h
+++ b/inst/include/Rcpp/XPtr.h
@@ -58,20 +58,12 @@ public:
 
     typedef StoragePolicy<XPtr> Storage;
 
-#if defined(RCPP_USING_CXX11)
-
     /**
      * constructs a XPtr wrapping the external pointer (EXTPTRSXP SEXP)
      *
      * @param xp external pointer to wrap
      */
-    explicit XPtr(SEXP x) {
-        if (TYPEOF(x) != EXTPTRSXP) {
-            const char* fmt = "Expecting an external pointer: [type=%s].";
-            throw ::Rcpp::not_compatible(fmt, Rf_type2char(TYPEOF(x)));
-        }
-        Storage::set__(x);
-    };
+    explicit XPtr(SEXP x) { checked_set(x); };
 
     /**
      * constructs a XPtr wrapping the external pointer (EXTPTRSXP SEXP)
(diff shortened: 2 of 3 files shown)


eddelbuettel:
> but not older Rtools

Well, allow me to place the burden on you if you want another change.  Win-builder gives you gcc/g++ 4.9.3.

stephematician:
So, when I run R CMD build it fails to build vignettes. Pressing on with --no-build-vignettes prior to upload to rhub, with Rtools 3.4 (warnings about building the vignettes should be immaterial?), i get an ok build: https://builder.r-hub.io/status/Rcpp_1.0.3.1.tar.gz-53526561b00f4b3fa49f78d89ae8169d.

Winbuilder (again with --no-build-vignettes prior to upload) goes ok up until tests, but it times out at 30 minutes and cancels the build, but otherwise looks ok: https://win-builder.r-project.org/Rb53AnRqyag4/00check.log. Winbuilder also uses R 3.5.3 which, I suspect, means Rtools 3.4 too. Given this has gcc 4.9.3 which should support C++11, not sure this proves beyond doubt that older Rtools versions (that don't support C++11) will work.

For reference, the error I get with R CMD build is:

Rscript -e 'tools::texi2pdf("Rcpp-attributes.tex", texi2dvi="pdflatex")'
Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
  Running 'texi2dvi' on 'Rcpp-attributes.tex' failed.
LaTeX errors:
! Package pdfpages Error: Cannot find file `pdf/Rcpp-attributes.pdf'.

See the pdfpages package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
! Emergency stop.
 ...                                              
                                                  
l.10 ...-, fitpaper=true]{pdf/Rcpp-attributes.pdf}
                                                  
!  ==> Fatal error occurred, no output PDF file produced!

eddelbuettel:
> [rhub] t fails to build vignettes

Yes, I have given up on that one a while ago.  See how I [invoke rhub builds](https://github.com/eddelbuettel/littler/blob/master/inst/examples/c4c.r). The latex infrastructure is just not there. I even turned all vignettes into mere includes of prebuilt pdf files.  Still....

/cc gaborcsardi

> Winbuilder [...] times out at 30 minutes

Yes, sorry, I had that too preparing 1.0.3, forgot about that.  You _could_ just go into `tests/doRUnit.R` and force `Sys.setenv("RunAllRcppTests"="no")` after which it should work.

stephematician:
I went back to 1.0.2.2 (commit 1730b7f9) and can replicate the warnings from https://github.com/RcppCore/Rcpp/pull/1003#issuecomment-550353482 both on my work machine (3.6.1, Rtools 3.4) and in winbuilder with tests disabled; see [00install.out.txt](https://github.com/RcppCore/Rcpp/files/3844131/00install.out.txt).

I then went back to the future (commit 9e6b2e7); no warnings on my work machine and winbuilder (tests disabled) output looks ok: see [00install.out.txt](https://github.com/RcppCore/Rcpp/files/3844167/00install.out.txt)
 
Links to the (72 hours until expiry) winbuilder output:
Commit 1730b7f9, with tests disabled: https://win-builder.r-project.org/6GSflTLlvi3g
Commit 9e6b2e7, with tests disalbed: https://win-builder.r-project.org/Kxi85lnHiYkB

eddelbuettel:
Sounds good.  Can you do me one more favour and add to [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog), please?

stephematician:
I think the changes are already there?

https://github.com/RcppCore/Rcpp/pull/1023/files#diff-02f0b547c2779d25cff89672135f20e3

eddelbuettel:
My apologies -- I had looked earlier and then mislabel "one commit" as "one file changed, hence no ChangeLog".  All good.

PR #1025, opened 2019-11-18
Add invisible param to export attribute

### LINKED ISSUE #1024: Need a way to return invisible
In my use case, my Rcpp function returns an object but I need the generated R wrapper use `invisible(.Call(...))` instead of `.Call(...)`.

I don't find a clean way to do that. I'm wondering if it is possible to add an attribute like `Rcpp::invisible` or something to indicate that the R wrapper generator uses `invisible`.

renkun-ken:
Closes #1024.

This PR add `invisible` param to `export` attribute so that if `Rcpp::export(invisible=true)` is specified, the R wrapper will use `invisible(.Call(...))` instead of `.Call(...)`.

#### Checklist

- [x] Code compiles correctly
- [x] The corresponding [Rcpp Attributes](https://github.com/RcppCore/Rcpp/blob/master/vignettes/rmd/Rcpp-attributes.Rmd) vignette is updated accordingly.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 2019-11-18  Kun Ren  <mail@renkun.me>
 
 	* src/attributes.cpp: Add invisible param to export attribute (#1024)
-        * vignettes/rmd/Rcpp-attributes.Rmd: Add an section of Returning invisible object
+	* vignettes/rmd/Rcpp-attributes.Rmd: Add an section of Returning invisible object
 
 2019-11-13  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2019-11-18  Kun Ren  <mail@renkun.me>
+
+	* src/attributes.cpp: Add invisible param to export attribute (#1024)
+
 2019-11-13  Dirk Eddelbuettel  <edd@debian.org>
 
 	* .github/ISSUE_TEMPLATE.md: Another small edit
@@ -11,7 +15,7 @@
 2019-11-11  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/unitTests/runit.packageversion.R: New test
-	* inst/unitTests/cpp/rcppversion.cpp: Cpp portion of test 
+	* inst/unitTests/cpp/rcppversion.cpp: Cpp portion of test
 
 2019-11-10  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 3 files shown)


src/attributes.cpp
```diff
@@ -2413,11 +2434,14 @@ namespace attributes {
                 // determine the function name
                 std::string name = attribute.exportedName();
 
+                // determine if return invisible
+                bool invisible = function.type().isVoid() || attribute.invisible();
```
eddelbuettel: Maybe call this `invisibleOrVoid` which would make it clearer below that what you add "just" generalizes the `void` case?

eddelbuettel: Or, looking at other boolean toggles in the files:  `isInvisibleOrVoid` ?

eddelbuettel:
Looks fine in general, but I would of course like to know what @jjallaire thinks here.

I also added an item to the checklist to make sure the corresponding vignette reflects it.

src/attributes.cpp
renkun-ken: ` isInvisibleOrVoid` looks good.

renkun-ken:
I don't see any unit test about export params. Should I add tests about `[[ Rcpp::export(invisible=true) ]]`?

eddelbuettel:
Good question.  I don't think we have any, and should maybe table it.  Would be nice to ascertain that we can turn visible on/off, that we can renamed (test could use `exists()` after the build I suppose) but less sure about the RNG option.

Your PR is fine by me now.  I will just wait for @jjallaire to glance at it too.  

So thanks for the PR and the quick refinements!

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2019-11-18  Kun Ren  <mail@renkun.me>
+
+	* src/attributes.cpp: Add invisible param to export attribute (#1024)
+        * vignettes/rmd/Rcpp-attributes.Rmd: Add an section of Returning invisible object
```
eddelbuettel: There is a minor misalignement here but do not worry about it.  I will probably make a new minor revision after the merge and can address that then.  Thanks for taking care of the `ChangeLog` entries.

renkun-ken:
Thanks for your kind suggestions!

ChangeLog
renkun-ken: Sorry for forgetting about the second tab here.

ChangeLog
```diff
@@ -29,14 +34,14 @@
 
 	* R/Attributes.R: Correct how cppFunction() deals with multiple
 	depends arguments
-
```
kevinushey: nit: please avoid whitespace-only changes when possible; consider configuring your editor to only trim whitespace on the lines you've edited

eddelbuettel: That hit me the other day too.  So I just googled it for The One True Editor: https://www.emacswiki.org/emacs/DeletingWhitespace#toc11   

Will give that a spin myself.

renkun-ken: Sorry for not noticing the auto formatting. I'll disable auto formatting for the project.

PR #1027, opened 2019-11-21
test for and report unavailable package in linkingto (closes #1026)

### LINKED ISSUE #1026: failure if pkgPath is not valid
For some obscure reason, a variable pkgPath is returned as character(0) in my unit tests run on R-devel (win10) under CMD check (in interactive session they run OK, as well as on R-3.6.1). Anyway, if this line:
https://github.com/RcppCore/Rcpp/blob/d6594bae57d796f0385bf9552532422e59a6e56d/R/Attributes.R#L964
returns character(0) the .linkingToIncludes() ends up with an error "argument is of length zero" in the line `if (file.exists(pkgHeaderPath)) {`. It would be great to add a test for `length(pkgPath)` and if it is zero, to stop with a meaningfull error message "package 'pkg' is not found" or alike.

eddelbuettel:
This addresses #1026 and checks against an empty path for `LinkingTo` if a non-exsting package is referenced from `cppFunction(..., depends=...)`.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2019-11-20  Dirk Eddelbuettel  <edd@debian.org>
+
+	* R/Attributes.R: Test for and report unavailable 'LinkingTo' package
+
 2019-11-19  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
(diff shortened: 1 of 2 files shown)


eddelbuettel:
A quick test for #1026 has been added in 37c390ec which will go in a (large) PR in a sec.

PR #1029, opened 2019-12-01
Feature/simpler description

eddelbuettel:
Now that the vignettes are pre-made and included as pdf files wrapped by tiny Rnw wrappers, we no longer need knitr, rmarkdown and pinp in the Suggests.  (They are still used 'off-site' to produce the pdf files included in the tarball.)  This makes the dependency graph a little simpler.

#### Checklist

- [x] Code compiles correctly


eddelbuettel:
Noone objected so merging in a moment or two...

PR #1037, opened 2019-12-17
Feature/win debug dll (closes #1035)

### LINKED ISSUE #1035: Should we support R CMD SHLIB -d on Windows only?
Via email, a user alerted me to the fact that he cannot create a gdb-compliant file on Windows when using `sourceCpp()`.

And that appears to be the case.  The bottom of `R CMD SHLIB --help` shows

```
Windows only:
  -d, --debug           build a debug DLL
```

And seemingly (when using `sourceCpp()`) one cannot get by via `~/.R/Makevars`.   Looking at `R/Attributes.R` it appears that we'd "just" need a new argument to `sourceCpp()` and insert an `ifelse()` line around line 124 just as we do to support `--preclean` and `--dry-run`.   

Shall we do that? I guess there is no real reason not to...

eddelbuettel:
See the discussion and resolution in #1035 -- this addresses a narrow need for debugging when using windows and `sourceCpp()`.   

#### Checklist

- [x] Code compiles correctly
- [x] User confirmed issue as addressed

### DIFF
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -67,7 +67,7 @@ sourceCpp <- function(file = "",
     } else {
         if (windowsDebugDLL) {
             if (verbose) {
-                message("The 'windowsDebugDLL' toggle is ignored on "
+                message("The 'windowsDebugDLL' toggle is ignored on ",
                         "non-Windows platforms.")
             }
             windowsDebugDLL <- FALSE    # now we do not need to deal with OS choice below
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -65,8 +65,11 @@ sourceCpp <- function(file = "",
                  "is not permitted.")
         }
     } else {
-        if (verbose && windowsDebugDLL) {
-            message("The 'windowsDebugDLL' toggle is ignored on non-Windows platforms.")
+        if (windowsDebugDLL) {
+            if (verbose) {
+                message("The 'windowsDebugDLL' toggle is ignored on "
+                        "non-Windows platforms.")
+            }
             windowsDebugDLL <- FALSE    # now we do not need to deal with OS choice below
         }							# #nocov end
     }
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -63,12 +63,12 @@ sourceCpp <- function(file = "",
         if (grepl(' ', basename(file), fixed=TRUE)) {
             stop("The filename '", basename(file), "' contains spaces. This ",
                  "is not permitted.")
-        }                                                       # #nocov end
+        }
     } else {
         if (verbose && windowsDebugDLL) {
             message("The 'windowsDebugDLL' toggle is ignored on non-Windows platforms.")
-            windowsDebugDLL <- FALSE    # no we do not need to for OS below
-        }
+            windowsDebugDLL <- FALSE    # now we do not need to deal with OS choice below
+        }							# #nocov end
     }
 
     # get the context (does code generation as necessary)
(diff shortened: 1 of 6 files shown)


R/Attributes.R
```diff
@@ -62,7 +63,12 @@ sourceCpp <- function(file = "",
         if (grepl(' ', basename(file), fixed=TRUE)) {
             stop("The filename '", basename(file), "' contains spaces. This ",
                  "is not permitted.")
-        }                                                       # #nocov end
+        }
+    } else {
+        if (verbose && windowsDebugDLL) {
+            message("The 'windowsDebugDLL' toggle is ignored on non-Windows platforms.")
+            windowsDebugDLL <- FALSE    # now we do not need to deal with OS choice below
```
coatless: Shouldn't this be outside the if()? As it stands now, this will only trigger if `verbose = TRUE` is passed (it normally isn't.)

eddelbuettel: Nice catch.  I think you are correct.  Will fix.

eddelbuettel: New variant:

```R
        if (windowsDebugDLL) {                                                                                                                                                                                     
            if (verbose) message("The 'windowsDebugDLL' toggle is ignored on non-Windows platforms.")                                                                                                              
            windowsDebugDLL <- FALSE    # now we do not need to deal with OS choice below                                                                                                                          
        }                                                       # #nocov end                                                                                                                                       
    }       
```

Better...

inst/NEWS.Rd
```diff
@@ -9,6 +9,8 @@
     \itemize{
       \item Safer \code{Rcpp_list*}, \code{Rcpp_lang*} and
       \code{Function.operator()} (Romain in \ghpr{1014}, \ghit{1015}).
+      \item A number of \code{#nocov} markers were added (Dirk in
+      \ghprP1036}).
```
coatless:
```suggestion
      \ghpr{1036}). 
```

R/Attributes.R
```diff
@@ -62,7 +63,15 @@ sourceCpp <- function(file = "",
         if (grepl(' ', basename(file), fixed=TRUE)) {
             stop("The filename '", basename(file), "' contains spaces. This ",
                  "is not permitted.")
-        }                                                       # #nocov end
+        }
+    } else {
```
coatless:
```suggestion
     else {
```

coatless: This is what broke the build.

inst/NEWS.Rd
eddelbuettel: Hmpf. I could swear I can a check which should spotted that.  Will fix!

R/Attributes.R
```diff
@@ -67,7 +67,7 @@ sourceCpp <- function(file = "",
     } else {
         if (windowsDebugDLL) {
             if (verbose) {
-                message("The 'windowsDebugDLL' toggle is ignored on "
+                message("The 'windowsDebugDLL' toggle is ignored on ",
```
coatless: Good catch on this one.

eddelbuettel: Well it was a last-second edit to comply with the @jjallaire style of less than 80 columns. I find 100 or so easier and this is what I got for it :)

eddelbuettel:
Thanks for the review.  There isn't all that much else here, and you caught the one obvious thinko, so I think I'll merge this now.

inst/NEWS.Rd
```diff
@@ -17,18 +19,27 @@
       \ghit{1017}).
       \item Invisible return object are now supported via new option (Kun Ren
       in \ghpr{1025} fixing \ghit{1024}).
+      \item Unavailable packages referred to in \code{LinkingTo} are now
+      reported (Dirk in \ghpr{1027} fixing \ghit{1026}).
+      \item The \code{sourceCpp} function can now create a debug DLL on
+      Windows (Dirk in \ghpr{1037} fixing \ghit{ghit}).
```
coatless: Also, there's `\ghit{ghit}`

eddelbuettel: Bah.  Something must be wrong with the water.  Or the coffee.  Or both.

Changed to `\ghit{1035}`. Thanks again.

PR #1043, opened 2020-02-09
Make Rcpp::exception thread-safe

jpritikin:

See the [discussion thread on rcpp-devel](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2020-February/010384.html).

I don't think new tests are a great idea. It would be tricky to reliably fail by throwing the old thread-unsafe version of `Rcpp::exception`.

#### Checklist

- [X] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2020-02-09  Joshua Pritikin  <jpritikin@pobox.com>
+
+	* inst/include/Rcpp.h: Include exceptions_impl.h
+	* inst/include/Rcpp/exceptions.h: Make thread-safe
+	* inst/include/Rcpp/exceptions_impl.h: New home for code moved
+	from inst/include/Rcpp/routines.h and src/api.cpp
+	* src/rcpp_init.cpp: stack_trace no longer exported as an R API
+
 2020-01-05  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/include/Rcpp/exceptions.h: A few more #nocov tags
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,8 +3,7 @@
 	* inst/include/Rcpp.h: Include exceptions_impl.h
 	* inst/include/Rcpp/exceptions.h: Make thread-safe
 	* inst/include/Rcpp/exceptions_impl.h: New home for code moved
-	from inst/include/Rcpp/routines.h and src/api.cpp
-	* src/rcpp_init.cpp: stack_trace no longer exported as an R API
+	from src/api.cpp
 
 2020-01-05  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -7,7 +7,6 @@
   \itemize{
     \item Changes in Rcpp API:
     \itemize{
-      \item \code{Rcpp::exception} is now thread-safe
       \item Safer \code{Rcpp_list*}, \code{Rcpp_lang*} and
       \code{Function.operator()} (Romain in \ghpr{1014}, \ghit{1015}).
       \item A number of \code{#nocov} markers were added (Dirk in
@@ -16,6 +15,9 @@
       Dirk in \ghpr{1038}).
       \item Scalar operations with a rhs matrix no longer change the
       matrix value (Qiang in \ghpr{1040} fixing (again) \ghit{365}).
+      \item \code{Rcpp::exception} and \code{Rcpp::stop} are now
+      thread-safe. Be cautioned that most of Rcpp continues to be not
+      thread-safe.
     }
     \item Changes in Rcpp Attributes:
     \itemize{
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -7,6 +7,7 @@
   \itemize{
     \item Changes in Rcpp API:
     \itemize{
+      \item \code{Rcpp::exception} is now thread-safe
       \item Safer \code{Rcpp_list*}, \code{Rcpp_lang*} and
       \code{Function.operator()} (Romain in \ghpr{1014}, \ghit{1015}).
       \item A number of \code{#nocov} markers were added (Dirk in
@@ -14,7 +15,7 @@
       \item Finalizer calls clear external pointer first (Kirill Müller and
       Dirk in \ghpr{1038}).
       \item Scalar operations with a rhs matrix no longer change the
-      matrix value (Qiang in \ghpr{1040} fixing (again) \ghit{365}). 
+      matrix value (Qiang in \ghpr{1040} fixing (again) \ghit{365}).
     }
     \item Changes in Rcpp Attributes:
     \itemize{
--- a/inst/include/Rcpp/exceptions.h
+++ b/inst/include/Rcpp/exceptions.h
@@ -27,21 +27,22 @@
 #define RCPP_DEFAULT_INCLUDE_CALL true
 #endif
 
-#define GET_STACKTRACE() stack_trace( __FILE__, __LINE__ )
-
 namespace Rcpp {
 
+    // Throwing an exception must be thread-safe to avoid surprises w/ OpenMP.
     class exception : public std::exception {
     public:
         explicit exception(const char* message_, bool include_call = RCPP_DEFAULT_INCLUDE_CALL) :	// #nocov start
             message(message_),
-            include_call_(include_call){
-            rcpp_set_stack_trace(Shield<SEXP>(stack_trace()));
+            include_call_(include_call)
+        {
+            record_stack_trace();
         }
         exception(const char* message_, const char*, int, bool include_call = RCPP_DEFAULT_INCLUDE_CALL) :
             message(message_),
-            include_call_(include_call){
-            rcpp_set_stack_trace(Shield<SEXP>(stack_trace()));
+            include_call_(include_call)
+        {
+            record_stack_trace();
         }
         bool include_call() const {
             return include_call_;
@@ -50,9 +51,12 @@ namespace Rcpp {
         virtual const char* what() const throw() {
             return message.c_str();					// #nocov end
         }
+        inline void copy_stack_trace_to_r() const;
     private:
         std::string message;
         bool include_call_;
+        std::vector<std::string> stack;
+        inline void record_stack_trace();
     };
 
     // simple helper
@@ -340,7 +344,8 @@ inline SEXP exception_to_condition_template( const Exception& ex, bool include_c
 }
 
 inline SEXP rcpp_exception_to_r_condition(const Rcpp::exception& ex) {
-  return exception_to_condition_template(ex, ex.include_call());
+    ex.copy_stack_trace_to_r();
+    return exception_to_condition_template(ex, ex.include_call());
 }
 
 inline SEXP exception_to_r_condition( const std::exception& ex){
--- a/inst/include/Rcpp/exceptions.h
+++ b/inst/include/Rcpp/exceptions.h
@@ -27,6 +27,8 @@
 #define RCPP_DEFAULT_INCLUDE_CALL true
 #endif
 
+#define GET_STACKTRACE() R_NilValue
+
 namespace Rcpp {
 
     // Throwing an exception must be thread-safe to avoid surprises w/ OpenMP.
@@ -34,14 +36,12 @@ namespace Rcpp {
     public:
         explicit exception(const char* message_, bool include_call = RCPP_DEFAULT_INCLUDE_CALL) :	// #nocov start
             message(message_),
-            include_call_(include_call)
-        {
+            include_call_(include_call) {
             record_stack_trace();
         }
         exception(const char* message_, const char*, int, bool include_call = RCPP_DEFAULT_INCLUDE_CALL) :
             message(message_),
-            include_call_(include_call)
-        {
+            include_call_(include_call) {
             record_stack_trace();
         }
         bool include_call() const {
--- a/inst/include/Rcpp/exceptions_impl.h
+++ b/inst/include/Rcpp/exceptions_impl.h
@@ -1,6 +1,6 @@
 // exceptions_impl.h: Rcpp R/C++ interface class library -- exceptions
 //
-// Copyright (C) 2020  Dirk Eddelbuettel and Romain Francois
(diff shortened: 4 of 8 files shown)


eddelbuettel:
Well this falls short on
- issue ticket with prior discussion in order to establish consensus about what to change, rather than throwing a PR over the fence
- does not demonstrate lack of impact on reverse dependencies, and it is not fair of you to assume I have to do it for you

eddelbuettel:
Also, I did point out to you on rcpp-devel that we had extensive discussions around that.  Maybe take a moment and look at 
- #877  with 12 comments
- #873  with 24 comments
- #868 with 10 comments
- #867 with 44 comments
- #859 with 31 comments
- #789 with 60 comments
- #706 with 14 comments
which may be relevant.

jpritikin:
> does not demonstrate lack of impact on reverse dependencies, and it is not fair of you to assume I have to do it for you

There is no impact on reverse dependencies. I merely reorganized the existing code.

> issue ticket with prior discussion in order to establish consensus about what to change, rather than throwing a PR over the fence

I'm happy to open an issue, but why not just have the discussion here?

> Also, I did point out to you on rcpp-devel that we had extensive discussions around that.

Hm, I didn't see these because I searched issues instead of PRs. The only one that looks relevant is #868. It seems like this PR was incomplete because `rcpp_set_stack_trace` was not conditioned on `include_call`. So I am happy to add another commit to obey `include_call` if you think that is the right thing to do.

Basically, my approach here was to make `Rcpp::exception` thread-safe and otherwise change as little as possible. So what's the next step?

jpritikin:
> does not demonstrate lack of impact on reverse dependencies, and it is not fair of you to assume I have to do it for you

What's an efficient way to demonstrate a lack of impact on reverse dependencies?

kevinushey:
This requires a reverse depends check, which does imply running `R CMD check` on all client packages of Rcpp. @eddelbuettel usually runs these after PRs which may be disruptive. (There's also the [revdepcheck](https://github.com/r-lib/revdepcheck) package for doing these kinds of checks as well)

eddelbuettel:
I wrote a package for that (see [prrd](https://github.com/eddelbuettel/prrd)) which I use on somewhat underpowered VM I have a courtesy account, but it busy today rev.dep checking RcppArmadillo which I submitted.  Doing Rcpp take 18+ hours. 

How to do reverse dependency checks is otherwise not a secret, discussed on the R mailing lists, supported by base R and several add-on packages.

eddelbuettel:
> There is no impact on reverse dependencies. I merely reorganized the existing code.

"Beware of bugs in the above code; I have only proved it correct, not tried it.".   -- Donald Knuth

inst/include/Rcpp/exceptions.h
```diff
@@ -27,21 +27,22 @@
 #define RCPP_DEFAULT_INCLUDE_CALL true
 #endif
 
-#define GET_STACKTRACE() stack_trace( __FILE__, __LINE__ )
```
kevinushey: Although unlikely it's possible that some client package of Rcpp makes use of this macro, so this could fail in revdep checks.

inst/include/Rcpp/routines.h
```diff
@@ -157,12 +156,6 @@ inline attribute_hidden const char* short_file_name(const char* file) {
     return fun(file);
 }
 
-inline attribute_hidden SEXP stack_trace( const char *file = "", int line = -1){
```
kevinushey: We should keep this routine registered.

inst/include/Rcpp/exceptions_impl.h
```diff
@@ -0,0 +1,69 @@
...
+            function_name.resize(function_plus);
+        }
+        buffer.replace(last_open + 1, function_name.size(), demangle(function_name));
+        return buffer;
+    }
+    #endif
+#endif
+
+    // thread-safe; invoked prior to throwing the exception
+    inline void exception::record_stack_trace()
```
kevinushey: Should this also accept the filename + line where the exception occurred? (I realize this seems to be otherwise unused but we should avoid API changes if possible)

inst/include/Rcpp/exceptions_impl.h
```diff
@@ -0,0 +1,69 @@
+namespace Rcpp {
```
kevinushey: Missing GPL license header.

inst/NEWS.Rd
```diff
@@ -7,14 +7,15 @@
   \itemize{
     \item Changes in Rcpp API:
     \itemize{
+      \item \code{Rcpp::exception} is now thread-safe
```
kevinushey: We should also indicate that `Rcpp::stop()` is now thread safe, as many users may not realize that `Rcpp::stop()` works by throwing an `Rcpp::exception`.

I think it's worth re-iterating that R / Rcpp is _not_ thread-safe in general. I don't want this PR to give users the idea that it might now be safe to use R / Rcpp APIs within OpenMP or separate threads.

jpritikin:
> This requires a reverse depends check, which does imply running R CMD check on all client packages of Rcpp.

Wow. Okay, I kicked off the job.

inst/NEWS.Rd
eddelbuettel: We also generally do not rewrite history.  New `\item` entries go into new `\itemize` blocks in new sections. If this were to be merged, it would be part of Rcpp 1.0.4 due next month.

inst/include/Rcpp/exceptions.h
```diff
@@ -27,21 +27,22 @@
...
 
+    // Throwing an exception must be thread-safe to avoid surprises w/ OpenMP.
     class exception : public std::exception {
     public:
         explicit exception(const char* message_, bool include_call = RCPP_DEFAULT_INCLUDE_CALL) :	// #nocov start
             message(message_),
-            include_call_(include_call){
-            rcpp_set_stack_trace(Shield<SEXP>(stack_trace()));
+            include_call_(include_call)
+        {
```
eddelbuettel: Nitpick but we have about 100k lines of code.  Very few place the opening curly on a line by itself.

inst/include/Rcpp/exceptions_impl.h
jpritikin: Ah, via the `GET_STACKTRACE` macro. I didn't consider that. We'll see what happens with revdep.

eddelbuettel:
@lionel- I couldn't add you to 'request a code review' but it would be good if you could look at this too.  If you have a few moments in the next few days it would be appreciated!

inst/NEWS.Rd
jpritikin: @eddelbuettel What do you mean? You want the `\item` added at the bottom instead of at the top? I did add it to the version 1.0.4-to-be section.

eddelbuettel: Then I misread about which bkock in went into, my bad.  But yes, within a release I tend to do it chronologically so last line would be better.

inst/include/Rcpp/exceptions_impl.h
jpritikin: Hm, given that you've recommended that `stack_trace` can return R_NilValue, I think there is no need to store filename + line.

jpritikin:
> Also, how to do reverse dependency checks is otherwise not a secret, discussed on the R mailing lists, supported by base R and several add-on packages.

If you are so tired of answering, you might as some hints in the `Contributing.md` file.

eddelbuettel:
The file, like everything else here, is based on volunteer effort and time, and in fact fairly recent.  It appears to have helped as well because your PR, while clearly not ready yet or consensus-based, did in fact 'tick off' a couple of boxes I consider helpful. 

If you have a suggestion to make for the file, why not raise an issue ticket and suggest new wording?

Lastly, I am sorry but as I recall you too have multiple CRAN uploads under your belt and state each time you do one that you checked reverse depends (if there any) for your package.  In other words you were already supposed to know this, which is why I allowed myself to use short hand.

jpritikin:
> Lastly, I am sorry but as I recall you too have multiple CRAN uploads under your belt and state each time you do one that you checked reverse depends (if there any) for your package. In other words you were already supposed to know this, which is why I allowed myself to use short hand.

I do indeed have a primitive script to check reverse depends, but nowhere near as polished as the scripts you suggested. It is one thing to check when there are less than 10 revdeps, and other thing entirely when there are more than 2000. Cheers!

eddelbuettel:
1857 as of today.  We omit Suggests:.

And I am aware that this is a lot of work.  I have been doing this diligently for some time now: https://github.com/RcppCore/rcpp-logs  

It is to the benefit of everybody who depends on Rcpp, i.e. all of use here and of course you too.  And if "stuff breaks" I have to deal with it which is why I lean towards the conservative side when it comes to unannounced PRs going deep into internals.

And as a whole we could do better around CRAN and tooling but alas there is always so much to do and so little time.  Again, if _you_ think you can make the process better please do not hold back (and I of course do not speak for CRAN as nobody does).

lionel-:
LGTM!

jpritikin:
I tried to run revdepcheck overnight and my laptop ran out of memory. Sorry but it seems that I don't have access to big enough hardware to run the check. I would appreciate it if somebody else can do it. No hurry; This can happen at your convenience.

eddelbuettel:
Now that RcppArmadillo is updated on CRAN, I can turn this on.  Runs unattended in a (very remote to me) data center anyway once I launch.  But what about comments / requests for changes Kevin made yesterday? (I didn't have time to look in detail, was busy shepherding RcppArmadillo and cranking out a paper draft...)

jpritikin:
> But what about comments / requests for changes Kevin made yesterday?

I believe that I have addressed them.

inst/include/Rcpp/exceptions_impl.h
```diff
@@ -0,0 +1,88 @@
+// exceptions_impl.h: Rcpp R/C++ interface class library -- exceptions
+//
+// Copyright (C) 2020  Dirk Eddelbuettel and Romain Francois
```
eddelbuettel: Well you wrote it, so might as well be your name...

eddelbuettel:
And just to be sure: you squashed/rebased your several commits back into one?  It is a little hard to see what changed when in that case.  Possibly nicer for merged history yet harder for PR assessment...

inst/include/Rcpp/exceptions_impl.h
jpritikin: I mostly just copy & pasted code, but sure, if you say so.

jpritikin:
> And just to be sure: you squashed/rebased your several commits back into one? It is a little hard to see what changed when in that case. Possibly nicer for merged history yet harder for PR assessment...

Yeah, I wasn't sure which kind of history was preferred.

eddelbuettel:
Well I guess all we know now is that I just wasted the last few minutes preparing a test tarball.

eddelbuettel:
And I am in no position to assess what just changed.  _That goes against the entire spirit of the PR we all just spent 24 hours on_.  Why?

jpritikin:
> And I am in no position to assess what just changed. That goes against the entire spirit of the PR we all just spent 24 hours on. Why?

I changed the authorship in the comment. That's it. From now on, I'll add commits instead of squashing.

jpritikin:
If you want, I can unsquash the commits. But that would require another force push. Let me know.

eddelbuettel:
I would appreciate that as it will no longer hide your work and make it inaccessible.  It is also how the rest the world works.  If and when we're done you can still squash, or I could squash-merge, ...

eddelbuettel:
Jeebus.

Did you just push AGAIN?

eddelbuettel:
The :-1: was a discouragement. As I, "no, I would prefer if you not pushed again".

jpritikin:
> I would appreciate that as it will no longer hide your work and make it inaccessible. It is also how the rest the world works. If and when we're done you can still squash, or I could squash-merge, ...

> The -1 was a discouragement. As I, "no, I would prefer if you not pushed again".

What the heck? You are giving contradictory instructions within the span of 30 minutes. Yeah, I pushed again to unsquash the commits.

eddelbuettel:
I'll unsubscribe from this thread now. I actually have work to do.

eddelbuettel:
@kevinushey When you have a moment, can you take a look at the changes made in response to you code review comments?   I have "unresolved" these; commonly the person who raises an issue gets to declare it resolved (rather than the person responding, but many things were a little unusual here...)

I had started one full reverse-depends which was promptly partially invalidated by forced pushes, but maybe that won't happen again either.  In any, results from the rev.dep runs are in https://github.com/RcppCore/rcpp-logs/commit/dc22889d100ad937ece0c55dac0bc09c51bac645 and https://github.com/RcppCore/rcpp-logs/commit/68114634b87773b05ceb8933bbb5f7671e7bf83b. Result are pretty stunning -- essentially only three fails on the last run all of which seem internal to the affected packages (but as usual I do skip a few for various reasons).

So in sum no issue from here: looks good.  Shout-out also to @lionel- for taking a look, appreciate it,

ChangeLog
```diff
@@ -1,3 +1,11 @@
+2020-02-09  Joshua Pritikin  <jpritikin@pobox.com>
+
+	* inst/include/Rcpp.h: Include exceptions_impl.h
+	* inst/include/Rcpp/exceptions.h: Make thread-safe
+	* inst/include/Rcpp/exceptions_impl.h: New home for code moved
+	from inst/include/Rcpp/routines.h and src/api.cpp
+	* src/rcpp_init.cpp: stack_trace no longer exported as an R API
```
kevinushey: nit: this is no longer true I think?

inst/include/Rcpp/exceptions_impl.h
```diff
@@ -0,0 +1,88 @@
+// exceptions_impl.h: Rcpp R/C++ interface class library -- exceptions
+//
+// Copyright (C) 2020 Joshua N. Pritikin
```
kevinushey: We should preserve the old author names as well (adding yourself is okay since you've also made changes)

eddelbuettel:
I can take care of the cleanup on the header and ChangeLog and NEWS entries.

jpritikin:
Thanks all! Sorry again for the force pushes.

eddelbuettel:
It's ok, I think we're all coming closer to the same page regarding process now.

Do you want to do another commit with touch-ups?  Or do you want me to?

PR #1055, opened 2020-03-22
Exceptions impl fixes

ellert:
Some fixes for inst/include/Rcpp/exceptions_impl.h
- the header file should have an include guard like all the other headers.
- The RCPP_DEMANGLER_ENABLED should always have a value - missing #else
- demangler_one should have an inline qualifier. Without this ROOT's R interface complains:

Processing /builddir/build/BUILD/root-6.20.02/tutorials/r/Function.C...
IncrementalExecutor::executeFunction: symbol '_ZN4RcppL13demangler_oneEPKc' unresolved 
while linking [cling interface function]!
You are probably missing the definition of Rcpp::demangler_one(char const*)
Maybe you need to load the corresponding shared library?

With the inline qualifier this error disappears.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2020-03-22  Mattias Ellert  <mattias.ellert@physics.uu.se>
+
+	* inst/include/Rcpp/exceptions_impl.h: Add include guard,
+	Make sure RCPP_DEMANGLER_ENABLED is always defined,
+	Add missing inline qualifier
+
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Thumbs up, and thanks, from me too.  Looks like we are breaking all sorts of (internal) record with that recent exceptions PR needing an unheard number of follow-ups :sob: 

Can you detail what makes ROOT's setup different so that it tripped here when this did not show anywhere else, including at CRAN?

Issue #1066, opened 2020-04-10
warning when compiling with clang-10 on macOS


kevinushey:
Saw this while compiling Rcpp with the latest version of LLVM on macOS:

```
/usr/local/opt/llvm/bin/clang++ -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -fsanitize=address,undefined -fno-omit-frame-pointer -fno-sanitize=alignment,function,float-divide-by-zero -std=gnu++11 -I"/Users/kevinushey/r/r-devel-sanitizers/lib/R/include" -DNDEBUG -I../inst/include/  -I/usr/local/include   -fPIC  -g -O1 -Wall -pedantic -DSWITCH_TO_REFCNT  -c module.cpp -o module.o
api.cpp:136:12: warning: address of stack memory associated with local variable 'f' returned [-Wreturn-stack-address]
    return f.c_str();
           ^
1 warning generated.
```

The offending routine is here:

https://github.com/RcppCore/Rcpp/blob/c9cf6aaf7b48216d141a3e549c74e50ebf55a13f/src/api.cpp#L129-L137

We are indeed returning a pointer to the data for a `std::string` on the stack, which is unsafe. Note that while this is a registered routine, it's only used internally (when Rcpp debugging is enabled via `RCPP_DEBUG`). Either way, two potential fixes here would be:

1. Just return a `std::string` rather than a `const char*`, or
2. Use a static `std::string`, and return a pointer to the data for that `std::string`.

eddelbuettel:
Ugg.  Option 1. is probably safer.

PR #1067, opened 2020-04-10
define safe variant of short_file_name

kevinushey:

Closes https://github.com/RcppCore/Rcpp/issues/1066. This PR comes with two pieces:

1. A "safe" definition of `short_file_name`, defined and used explicitly in the `Rcpp::debug` namespace;

2. A re-worked version of the "old" (+ registered) `short_file_name`, which now uses a static buffer. I believe the method should be effectively un-used but I figured since it's a registered routine it's best to keep the API the same and just silence the compiler warning.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2020-04-10  Kevin Ushey  <kevinushey@gmail.com>
+
+	* inst/include/Rcpp/macros/macros.h: Safer definition of short_file_name
+	* inst/include/Rcpp/String.h: Idem
+	* inst/include/Rcpp/macros/debug.h: Idem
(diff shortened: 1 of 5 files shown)


eddelbuettel:
I _think_ we once use proper deprecation.  

We could "tag" the older unsafer variant and say it'll go in, say, 12 month.  Thoughts?  (That is somewhat orthogonal to the PR though of course not fully...)

kevinushey:
I don't have any strong feelings re: deprecation -- I'll let you make the call. I expect removing it would go smoothly, but experience has taught me to always expect surprises.

Issue #1074, opened 2020-04-25
Memory leak in Rcpp::sample


kkrismer:
First of all, thanks for the great work, Rcpp is awesome!

I've noticed that memory usage blows up when repeatedly calling `Rcpp::sample()`.

Running this minimal reproducible example allocates multiple GB of memory:
```
Rcpp::cppFunction("void call_sample_n_times(const int n) {
        for(int i(0); i < n; ++i) {
            Rcpp::sample(15, 5);
        }
    }")

set.seed(1)
call_sample_n_times(10^8)
```
The expected memory footprint of `call_sample_n_times(10^8)` is pretty much zero. Note that the parameters to `Rcpp::sample` do not depend on `n`.

`Rcpp::RcppArmadillo::sample` does not exhibit this behavior. The equivalent `RcppArmadillo` code uses almost no memory, as expected:
```
Rcpp::IntegerVector vec(Rcpp::seq_len(15))
for(int i(0); i < n; ++i) {
    Rcpp::RcppArmadillo::sample(vec, 5, false);
}
```

Tested on Windows 10 and Ubuntu 18.04.

Ubuntu 18.04:
```
> sessionInfo()
R version 3.6.3 (2020-02-29)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.4 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_3.6.3 tools_3.6.3    Rcpp_1.0.4.6
```

Windows 10:
```
> sessionInfo()
R version 3.6.3 (2020-02-29)
[...]

PR #1075, opened 2020-04-25
Use an explicit vector rather than R_alloc in sample

eddelbuettel:
As discussed in #1074, the behavior of `R_alloc()` seems to be less than ideal, and a simple replacement of an `IntegerVector` instantiation appears to help in two places.

#### Checklist

- [x] Code compiles correctly


eddelbuettel:
@kevinushey Thanks for the hint on `no_init()` which I already added.  Anything else, or good to go?

PR #1078, opened 2020-04-30
Add new vignette about packaging C++ libraries (closes #1077)

### LINKED ISSUE #1077: Add the '13 steps for package with C++ library' paper?
How to build R package with an external C++ library is still tricky, with or without Rcpp. In workshops I often mention three cases of i) headers-only, ii) small sources (to be included), iii) large sources (external library) and mention that especially the latter one is still hard esp cross-platform. 

Now, I did sit down in November on wrote down some notes that turned into a arXiv pre-print (at [arxiv.org/abs/1911.06416](https://arxiv.org/abs/1911.06416)) that I actually never really announced,. Somehow made it into a twitter thread recently, and people seem to find it useful. It occurred to me that we could stick it into the package as a new vignette as well.

Thoughts?  Yay?  Nay?

eddelbuettel:
As discussed in #1077, we have a [new vignette candidate](https://arxiv.org/abs/1911.06416). And there seems to be some support for including it. 

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2020-04-30  Dirk Eddelbuettel  <edd@debian.org>
+
+	* vignettes/rmd/Rcpp-libraries.Rmd: Add new vignette source
+	* vignettes/rmd/Rcpp.bib: Complete references
+        * inst/bib/Rcpp.bib: Idem
+
 2020-04-26  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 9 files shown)


eddelbuettel:
Will go slightly off usual schedule and merge this -- as it leaves `inst/include`, `src/`, `R/` and `inst/tinytest` alone there should really be no surprise from this on the behavior side.

PR #1085, opened 2020-05-28
Added support for std::string_view to Rcpp::as

### LINKED ISSUE #1084: Add support for std::string_view in Rcpp::as
I would like to add support for the std::string_view type (new in C++17) to Rcpp::as.

My motivation is that I'm working on C++ protobuf, trying to update [ServiceDescriptor::FindMethodByName()](https://github.com/protocolbuffers/protobuf/blob/2d654e8198a0eb6f9e228f964a70f15abd97512f/src/google/protobuf/descriptor.h#L1270) to take std::string_view. RProtobuf calls that function [here](https://github.com/eddelbuettel/rprotobuf/blob/868ad27b2f6028827f2f36fa531ef58287fa72db/src/RcppMacros.h#L147) (see also [here](https://github.com/eddelbuettel/rprotobuf/blob/868ad27b2f6028827f2f36fa531ef58287fa72db/src/wrapper_ServiceDescriptor.cpp#L15)), and relies on the Rcpp converter to convert to a string to pass to FindMethodByName(). I found that this stopped working once I updated that function to take std::string_view instead of const std::string&.

I will send out a pull request in just a minute, but I wanted to open this issue first as per the contributing guidelines.

acozzette:
Fixes #1084.

This commit adds support for std::string_view to Rcpp::as. The
std::string_view type is only available starting with C++17, but I
guarded it with the necessary ifdef so that we do not rely on it unless
it's available.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2020-05-28  Adam Cozzette  <acozzette@google.com>
+
+        * inst/include/Rcpp/as.h: Added support for std::string_view
+        * inst/include/Rcpp/traits/r_type_traits.h: Idem
+
 2020-05-18  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll minor version
--- a/inst/include/Rcpp/as.h
+++ b/inst/include/Rcpp/as.h
@@ -71,6 +71,10 @@ namespace Rcpp {
             return as_string<T>(x, typename Rcpp::traits::is_wide_string<T>::type());
         }
 
+        template <typename T> T as(SEXP x, ::Rcpp::traits::r_type_string_view_tag) {
+            return check_single_string(x);
+        }
+
         template <typename T> T as(SEXP x, ::Rcpp::traits::r_type_RcppString_tag) {
             if (! ::Rf_isString(x)) {
                 const char* fmt = "Expecting a single string value: "
--- a/inst/include/Rcpp/traits/r_type_traits.h
+++ b/inst/include/Rcpp/traits/r_type_traits.h
@@ -23,6 +23,10 @@
 #ifndef Rcpp__traits__r_type_traits__h
 #define Rcpp__traits__r_type_traits__h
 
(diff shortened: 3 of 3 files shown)


kevinushey:
This feels pretty dangerous to me.

My understanding is that `string_view` is effectively just a raw pointer to a `const char*` with some helper methods, and we're trying to construct that string view from an Rcpp object (a `CHARSXP`). So, that string view will be valid only as long as that Rcpp object is valid.

IMHO, `Rcpp::as<>` should normally give you an independent object that you can use safely and independently of the R / Rcpp stack; this PR would violate that contract.

eddelbuettel:
Thanks Kevin.  My gut fear of 'oh dear' was not unfounded...

acozzette:
Hmm, good point about the lifetime issue. That makes me think that `Rcpp::as<>` is probably not the best thing to change. To solve the protobuf problem I ran into, I can think of two other options:
1. Change `Rcpp::internal::converter` to be capable of converting to `std::string_view`, but without relying on `Rcpp::as<>`.
2. Update RProtobuf and leave Rcpp unchanged.

I guess I would lean toward 2 as the better option, since it sounds like probably no one else would benefit from `Rpc::internal::converter` supporting `std::string_view`. Does that sound like the right way to go?

kevinushey:
I would lean towards (2) as well.

eddelbuettel:
Ditto. Has the advantage of a more limited score and maybe clearer ownership and flow.   If ever we feel it is more generic we can bring it back too.

acozzette:
OK, thank you for the feedback. I just created a [pull request](https://github.com/eddelbuettel/rprotobuf/pull/72) for RProtoBuf and I will close this PR and the related issue in Rcpp.

eddelbuettel:
Sorry for the run-around, but this ended up being a good example of why talking it over first may be beneficial.

jwood000:
@eddelbuettel, I couldn't agree more.

It's a privilege to be able to see these discussions. I learn something every time.

PR #1088, opened 2020-06-06
correct package skeleton behavior (closes #1087)

### LINKED ISSUE #1087: Possible package.skeleton issue affecting Rcpp, RcppArmadillo, RStudio (!!), ...
So I just did the [live webinar](https://www.youtube.com/watch?v=1Zap6j1uUyw) on Rcpp and got to show live how package generation now fails (from RStudio) for both Rcpp and RcppArmadillo :cry: 

It looks the `NAMESPACE` file now gets an 'echo' from a temp function we create but which is no longer there when the package is finished.  And misses an actual `exportPatter()`.  

So a simple first fix for RcppArmadillo could be (done and tested)

```diff
modified   R/RcppArmadillo.package.skeleton.R
@@ -78,6 +78,8 @@ RcppArmadillo.package.skeleton <- function(name="anRpackage", list=character(),
     ## add a useDynLib to NAMESPACE,
     NAMESPACE <- file.path( root, "NAMESPACE")
     lines <- readLines( NAMESPACE )
+    lines <- lines[!grepl("^export.*fake", lines)]
+    lines <- c(lines, "exportPattern(\"^[[:alpha:]]+\")")
     if (! grepl("useDynLib", lines)) {
         lines <- c(sprintf("useDynLib(%s, .registration=TRUE)", name),
                    "importFrom(Rcpp, evalCpp)",        ## ensures Rcpp instantiation
```

do not print the `fake` one but add the pattern.  Is my fear justified that the RStudio generator may need this too?  Paging @kevinushey ....

eddelbuettel:
This is in relation to #1087 and the fact that the 'package creation helper' (both functions such as `Rcpp.package.skeleton()` and `RcppArmadillo.package.skeleton()` _and_ the functionality accessible from RStudio via File -> New Project -> Package) suffers from an issue in NAMESPACE.

The fix is simple and gets to it based on my testing today, but someone else should have a look at it.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,8 @@
 2020-06-05  Dirk Eddelbuettel  <edd@debian.org>
 
+        * DESCRIPTION (Version, Date): Roll minor version
+        * inst/include/Rcpp/config.h: Idem
+
 	* R/Rcpp.package.skeleton.R: Remove a remaining NAMESPACE entry for
 	a temporary function symbol; ensure an exportPattern is set
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2020-06-05  Dirk Eddelbuettel  <edd@debian.org>
+
+	* R/Rcpp.package.skeleton.R: Remove a remaining NAMESPACE entry for
+	a temporary function symbol; ensure an exportPattern is set
+
 2020-05-18  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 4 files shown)


kevinushey:
LGTM!

eddelbuettel:
Thanks Kevin.  Any comment on what we need to do to fix it from RStudio?  I keep forgetting which if any package functions are called here...

Issue #1097, opened 2020-06-29
avoid using EXTPTR_PTR


kevinushey:
As it's not part of the official R API. See https://r.789695.n4.nabble.com/Possible-ABI-change-in-R-4-0-1-td4764335.html for details.

We should use `R_ExternalPtrAddtr()` instead.

```
kevinushey@cdrv:~/r/pkg/Rcpp [master]
$ rg "EXTPTR_PTR"
inst/include/Rcpp/module/class.h
170:                vec_signed_method* mets = reinterpret_cast< vec_signed_method* >( EXTPTR_PTR( method_xp ) ) ;
197:                vec_signed_method* mets = reinterpret_cast< vec_signed_method* >( EXTPTR_PTR( method_xp ) ) ;
219:                vec_signed_method* mets = reinterpret_cast< vec_signed_method* >( EXTPTR_PTR( method_xp ) ) ;
396:                prop_class* prop = reinterpret_cast< prop_class* >( EXTPTR_PTR( field_xp ) ) ;
403:                prop_class* prop = reinterpret_cast< prop_class* >( EXTPTR_PTR( field_xp ) ) ;

src/api.cpp
148:    snprintf(buffer, 20, "%p", (void*)EXTPTR_PTR(xp));

src/module.cpp
56:    return EXTPTR_PTR(xp) == 0;
```

Evidently this change was done way back in 2011 as well for some usages:

https://github.com/RcppCore/Rcpp/blob/b1254701a899cb187f9a6917cb8d18c7f93290c7/ChangeLog#L6630-L6634

eddelbuettel:
Sure, good catch.  Saw that discussion, and of course the ping-pong between @gaborcsardi and myself about the `roxygen2` issue he was pinning on us which I didn't want to have here ... til @jeroen found this underlying change.

Which, almost surely, is not related to the `roxygen2` issue as the above (which I also see here using `ag` rather than `rg`) affects only Modules which "some" package use, but not many -- and almost surely not `roxygen2`. 

Let me chew over if I want to tackle this this week or rather ship what we have now as 1.0.5 and then let a change such as this "rest and mature" a little en route to 1.0.6.  If that release happens.  Because I am still not over the nonsense I had to put up with in the discussion after 1.0.4 came out, and we had risked late changes.   Very little appetite for repeating that....

eddelbuettel:
[...]

PR #1098, opened 2020-06-30
use documented R API function for extptr addr

kevinushey:
Closes https://github.com/RcppCore/Rcpp/issues/1097.

"In theory" this PR should be safe ... but I'd definitely understand if we still don't want to take it so close to release.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,11 @@
         * inst/NEWS.Rd: Idem
         * vignettes/rmd/Rcpp.bib: Idem
         * inst/bib/Rcpp.bib: Idem
+2020-06-29  Kevin Ushey  <kevinushey@gmail.com>
+
+	* inst/include/Rcpp/module/class.h: EXTPTR_PTR -> R_ExternalPtrAddr
+	* src/api.cpp: Idem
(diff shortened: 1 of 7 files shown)


eddelbuettel:
Thanks, on balance I think this will have to wait til after 1.5.0 is out.

eddelbuettel:
I rebased this against `master` which has the 1.0.5 release now on CRAN, and will add a new minor release version 1.0.5.1 for it.

Ok with you I hope?  ;-)

kevinushey:
LGTM -- thanks!

eddelbuettel:
I was a little slow preparing a new rev.dep check for this 1.0.5.1 rc, but it is running. I may give a few hours, or maybe even wait til tomorrow, to see all is ok as expected and then merge.

eddelbuettel:
It's over 1/3 done and looking reasonable so far, so merging this now.

eddelbuettel:
Just for completeness: no issue came up, the reverse depends check concluded just fine, and I just commited the summary to the usual `rcpp-logs` repo.

PR #1099, opened 2020-07-01
Implement push_back and push_front for DataFrame.

### LINKED ISSUE #1094: DataFrame converting to list on push_back with some c++ standards/compilers
I want to construct a data frame in Rcpp, with some variable number of columns. To do this, I am using `push_back()`.

However, depending on compiler options, this results in the data frame converting into a list. Something similar was observed back in 2011 in https://stackoverflow.com/questions/8631197/constructing-a-data-frame-in-rcpp.

If I have `CXX=g++ --std=gnu++11` in my `~/.R/Makevars` (the default; alternatively, if I leave `~/.R/Makevars` empty, or `clang++` with any standard), I get the following
``` r
library(Rcpp)

A <- cppFunction(code='DataFrame make_df() {
NumericVector v(1);
DataFrame df = DataFrame::create(_["v"] = v);
return(df);}', rebuild = TRUE)

B <- cppFunction(code='DataFrame make_df() {
NumericVector v(1);
DataFrame df = DataFrame();
df.push_back(v, "v");
return(df);}', rebuild = TRUE)

A() # Constructed with create()
#>   v
#> 1 0
B() # Constructed with push_back()
#> $v
#> [1] 0
```

<sup>Created on 2020-06-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I would expect that `A()` and `B()` are the same.

If I specify any other standard to `g++` (I tried `gnu++14`, `gnu++17`, `gnu++2a`, not specifying a standard), then I get the following, which is what I would expect

``` r
library(Rcpp)

A <- cppFunction(code='DataFrame make_df() {
NumericVector v(1);
DataFrame df = DataFrame::create(_["v"] = v);
return(df);}', rebuild = TRUE)

B <- cppFunction(code='DataFrame make_df() {
NumericVector v(1);
DataFrame df = DataFrame();
df.push_back(v, "v");
return(df);}', rebuild = TRUE)

A() # Constructed with create()
#>   v
#> 1 0
B() # Constructed with push_back()
#>   v
#> 1 0
```

<sup>Created on 2020-06-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I know that I can use `as.data.frame()` on the result, or `Rcpp::as<Rcpp::DataFrame>()`, to convert back to a data frame to get what I want, but this compiler-dependent behaviour seems like a bug.

``` r
sessionInfo()
#> R version 4.0.0 (2020-04-24)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: Debian GNU/Linux bullseye/sid
#> 
#> Matrix products: default
#> BLAS:   /home/walter/R/lib/R/lib/libRblas.so
#> LAPACK: /home/walter/R/lib/R/lib/libRlapack.so
#> 
#> locale:
#>  [1] LC_CTYPE=en_NZ.UTF-8       LC_NUMERIC=C              
#>  [3] LC_TIME=en_NZ.UTF-8        LC_COLLATE=en_NZ.UTF-8    
#>  [5] LC_MONETARY=en_NZ.UTF-8    LC_MESSAGES=en_NZ.UTF-8   
#>  [7] LC_PAPER=en_NZ.UTF-8       LC_NAME=C                 
#>  [9] LC_ADDRESS=C               LC_TELEPHONE=C            
#> [11] LC_MEASUREMENT=en_NZ.UTF-8 LC_IDENTIFICATION=C       
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
#>  [1] compiler_4.0.0  magrittr_1.5    tools_4.0.0     htmltools_0.4.0
#>  [5] yaml_2.2.1      Rcpp_1.0.4.6    stringi_1.4.6   rmarkdown_2.1  
#>  [9] highr_0.8       knitr_1.28      stringr_1.4.0   xfun_0.14      
#> [13] digest_0.6.25   rlang_0.4.6     evaluate_0.14
```

<sup>Created on 2020-06-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

waltersom:
This explicitly defines push_back and push_back methods for DataFrame
which copies row.names and class attributes from the original object.

This fixes #1094 where these operations converted the object to a list.

Tests for this behavior are also implemented.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,6 +17,13 @@
         * inst/NEWS.Rd: Idem
         * vignettes/rmd/Rcpp.bib: Idem
         * inst/bib/Rcpp.bib: Idem
+2020-07-01 Walter Somerville <waltersom@gmail.com>
+	
+	* inst/include/Rcpp/DataFrame.h: Implement explict push_back and
+	push_front for DataFrame, which takes care to set the class and
+	row.names attributes.
+	* inst/tinyTest/test_dataframe.R Add in tests for push_back/push_front
+	* inst/tinyTest/cpp/DataFrame.cpp
 
 2020-06-27  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,17 @@
+2020-07-09  Walter Somerville  <waltersom@gmail.com>
+
+	* inst/include/Rcpp/DataFrame.h: Warn when data.frame has varying length
+	* inst/tinyTest/cpp/DataFrame.cpp: Test added
+	* inst/tinyTest/test_dataframe.R: Idem
+
+2020-07-07  Walter Somerville  <waltersom@gmail.com>
+
+	* inst/include/Rcpp/DataFrame.h: Implement explict push_back and
+	push_front for DataFrame, which takes care to set the class and
+	row.names attributes.
+	* inst/tinyTest/test_dataframe.R: Add in tests for push_back/push_front
+	* inst/tinyTest/cpp/DataFrame.cpp: Idem
+
 2020-07-06  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll minor version
@@ -17,13 +31,6 @@
         * inst/NEWS.Rd: Idem
         * vignettes/rmd/Rcpp.bib: Idem
         * inst/bib/Rcpp.bib: Idem
-2020-07-01 Walter Somerville <waltersom@gmail.com>
-	
-	* inst/include/Rcpp/DataFrame.h: Implement explict push_back and
-	push_front for DataFrame, which takes care to set the class and
-	row.names attributes.
-	* inst/tinyTest/test_dataframe.R: Add in tests for push_back/push_front
-	* inst/tinyTest/cpp/DataFrame.cpp: Idem
 
 2020-06-27  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,8 +22,8 @@
 	* inst/include/Rcpp/DataFrame.h: Implement explict push_back and
 	push_front for DataFrame, which takes care to set the class and
 	row.names attributes.
-	* inst/tinyTest/test_dataframe.R Add in tests for push_back/push_front
-	* inst/tinyTest/cpp/DataFrame.cpp
+	* inst/tinyTest/test_dataframe.R: Add in tests for push_back/push_front
+	* inst/tinyTest/cpp/DataFrame.cpp: Idem
 
 2020-06-27  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/inst/include/Rcpp/DataFrame.h
+++ b/inst/include/Rcpp/DataFrame.h
@@ -85,38 +85,26 @@ namespace Rcpp{
 
 	template <typename T>
         void push_back( const T& object){
-            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
             Parent::push_back(object);
-            IntegerVector RowNames = Range(1, NewSize);
-            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
-            Rf_setAttrib(Parent::get__(), R_ClassSymbol, Rf_mkString("data.frame"));
+            set__(Parent::get__());
         }
 
         template <typename T>
         void push_back( const T& object, const std::string& name ){
-            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
             Parent::push_back(object, name);
-            IntegerVector RowNames = Range(1, NewSize);
-            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
-            Rf_setAttrib(Parent::get__(), R_ClassSymbol, Rf_mkString("data.frame"));
+            set__(Parent::get__());
         }
 
         template <typename T>
         void push_front( const T& object){
-            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
             Parent::push_front(object);
-            IntegerVector RowNames = Range(1, NewSize);
-            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
-            Rf_setAttrib(Parent::get__(), R_ClassSymbol, Rf_mkString("data.frame"));
+            set__(Parent::get__());
         }
 
         template <typename T>
         void push_front( const T& object, const std::string& name){
-            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
             Parent::push_front(object, name);
-            IntegerVector RowNames = Range(1, NewSize);
-            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
-            Rf_setAttrib(Parent::get__(), R_ClassSymbol, Rf_mkString("data.frame"));
+            set__(Parent::get__());
         }
                 
         // Offer multiple variants to accomodate both old interface here and signatures in other classes
--- a/inst/include/Rcpp/DataFrame.h
+++ b/inst/include/Rcpp/DataFrame.h
@@ -142,7 +142,7 @@ namespace Rcpp{
                 }
             }
             for (it = Parent::begin(); it != Parent::end(); ++it) {
-                if (Rf_xlength(*it) > 1 && max_rows % Rf_xlength(*it) != 0) {
+                if (Rf_xlength(*it) == 0 || ( Rf_xlength(*it) > 1 && max_rows % Rf_xlength(*it) != 0 )) {
                     // We have a column that is not an integer fraction of the largest
                     invalid_column_size = true;
                 }
--- a/inst/include/Rcpp/DataFrame.h
+++ b/inst/include/Rcpp/DataFrame.h
@@ -83,6 +83,42 @@ namespace Rcpp{
             return LENGTH(rn);
         }
 
+	template <typename T>
+        void push_back( const T& object){
+            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
+            Parent::push_back(object);
+            IntegerVector RowNames = Range(1, NewSize);
+            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
+            Rf_setAttrib(Parent::get__(), R_ClassSymbol, Rf_mkString("data.frame"));
+        }
+
+        template <typename T>
+        void push_back( const T& object, const std::string& name ){
+            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
+            Parent::push_back(object, name);
+            IntegerVector RowNames = Range(1, NewSize);
+            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
+            Rf_setAttrib(Parent::get__(), R_ClassSymbol, Rf_mkString("data.frame"));
+        }
+
+        template <typename T>
+        void push_front( const T& object){
+            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
+            Parent::push_front(object);
+            IntegerVector RowNames = Range(1, NewSize);
+            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
+            Rf_setAttrib(Parent::get__(), R_ClassSymbol, Rf_mkString("data.frame"));
+        }
+
+        template <typename T>
+        void push_front( const T& object, const std::string& name){
+            R_xlen_t NewSize = std::max(object.size(), static_cast<R_xlen_t>(nrow()));
+            Parent::push_front(object, name);
+            IntegerVector RowNames = Range(1, NewSize);
+            Rf_setAttrib(Parent::get__(), R_RowNamesSymbol, RowNames);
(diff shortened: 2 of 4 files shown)


waltersom:
This does not enforce that the vectors are the same length (existing package doesn't either).

waltersom:
The new version uses `set__()` in `DataFrame.h`, which has the huge benefit of padding/replicating vectors of different lengths, making names, and so forth.

waltersom:
I still feel this is a bit inelegant as it only solves the problem for data frames, but I guess it is also class-specific how attributes should change when they are combined, so it might be better that any class that has a problem implements the appropriate fix like this. I'm not sure if any other classes suffer from this problem.

eddelbuettel:
How does anybody else feel about changing the names of these verbs?  `push_back` _et al_ just invoked appending to a single vector.  Maybe `append_column` and `prepend_column` (both clumsy) or `add_col_at_back` and `add_col_at_front`?  Maybe with `column` instead of `col` ?

waltersom:
`push_*` does sound a bit clumsy, but I am a bit concerned that it is possible to combine two data frames this way (or add a list to a data frame, I guess), and not just add a column. Would it work with just `append` and `prepend`?

kevinushey:
My thoughts:

1. We should keep `push_back()` / `pop_back()`, to conform with the existing Vector interface.
2. If we feel it's worth it, we could add `append_column()` / `prepend_column()` as aliases for those functions. I don't have strong feelings here though.

> I still feel this is a bit inelegant as it only solves the problem for data frames, but I guess it is also class-specific how attributes should change when they are combined, so it might be better that any class that has a problem implements the appropriate fix like this. I'm not sure if any other classes suffer from this problem.

I agree it'd be nice to fix this in general, but it seems hard to find the right solution that would work "in general" so it's okay to keep things scoped more narrowly here for now.

eddelbuettel:
Happy to be majority-vote overruled but I still feel quite strongly that `push_back()` is not a good idea. It (to me) has clear connotations of extending a single _vector_ ("verticallly" if you wish) whereas the PR uses to expand a list holding such vectors by enlarging a `data.frame ("horizontally").  Nothing good can come from mixing terminology that way.

kevinushey:
My main concern is that `DataFrame.push_back()` already does exist (by virtue of being available in the base class) and so I don't think we should change the pre-existing semantics there.

If this were a clean-room rebuild of the class, I would definitely be on board though.

eddelbuettel:
I agreee with that point @kevinushey.  That is indeed more important.  Having the aliases is then a good compromise.

eddelbuettel:
Thanks for your patience.  Rcpp 1.0.5 is now on CRAN, I am currently looking into #1098 and will turn to this afterwards. Likely in a day or two.

eddelbuettel:
I rebased this relative to the updated main branch and pushed the update.  I have a local (not-yet committed) change adding ChangeLog and setting this as 1.0.5.2 and am now running reverse depends.  We will know more by tomorrow this time.

eddelbuettel:
Hi @waltersom and apologies for taking a moment---but I first wanted to get 1.0.5 out of the door, and then test the PR by @kevinushey which came in just before yours.

I have now done a full run, and in this case the full test paid us back.  Package lidR and metacart fail under your PR.  Both show a message of 'Could not convert using R function: as.data.frame.'.   Could you take another look at the PR?

Many thanks!

waltersom:
No worries about the time, it's good to get it right, and there's no great rush.

I'll have a look at why they might get that error.

waltersom:
Ok, I have some minimum examples which exhibit these problems; `f` is representative of the problem in `lidR`, and `g` from `metacart`.

``` r
library(Rcpp)

f <- cppFunction('DataFrame d(){
DataFrame d = DataFrame::create();

NumericVector v1(5);
NumericVector v2(5);
DataFrame d1 = DataFrame::create(_["a"]=v1, _["b"] = v2);
d.push_back(v1);
SETLENGTH(wrap(v1), 2);
d.push_back(v2);
return d;
}')
f()
#> Error in f(): Could not convert using R function: as.data.frame.


g <- cppFunction('DataFrame d(){
DataFrame d = DataFrame::create();

NumericVector v1(5);
NumericVector v2(6);
d.push_back(v1);
d.push_back(v2);
return d;
}')
g()
#> Error in g(): Could not convert using R function: as.data.frame.
```

<sup>Created on 2020-07-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


Using the pre-PR code, I get
``` r
library(Rcpp)

f <- cppFunction('DataFrame d(){
DataFrame d = DataFrame::create();

NumericVector v1(5);
NumericVector v2(5);
DataFrame d1 = DataFrame::create(_["a"]=v1, _["b"] = v2);
d.push_back(v1);
SETLENGTH(wrap(v1), 2);
d.push_back(v2);
return d;
}')
f()
#> [[1]]
#> [1] 0 0
#> 
#> [[2]]
#> [1] 0 0 0 0 0
as.data.frame(f())
#> Error in (function (..., row.names = NULL, check.rows = FALSE, check.names = TRUE, : arguments imply differing number of rows: 2, 5

g <- cppFunction('DataFrame d(){
DataFrame d = DataFrame::create();

NumericVector v1(5);
NumericVector v2(6);
d.push_back(v1);
d.push_back(v2);
return d;
}')
g()
#> [[1]]
#> [1] 0 0 0 0 0
#> 
#> [[2]]
#> [1] 0 0 0 0 0 0
as.data.frame(g())
#> Error in (function (..., row.names = NULL, check.rows = FALSE, check.names = TRUE, : arguments imply differing number of rows: 5, 6
```

<sup>Created on 2020-07-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

My feeling is that both packages are constructing invalid dataframes, and probably getting them fixed would be good.

The fix for both is easy, and just involves changing a DataFrame to a List (which is essentially what it was before, after a push_back).

waltersom:
The relevant lines are (eg) https://github.com/Jean-Romain/lidR/blob/9f30c6f8065b6fe6571c2d53457cfbab4a1fc3ca/src/LAS.cpp#L1328-L1331
and 
https://github.com/cran/metacart/blob/master/src/testxvalid.cpp#L33-L53

eddelbuettel:
> My feeling is that both packages are constructing invalid dataframes

That is possible but we would probably have to find their code [and as I type this you post it, will study.]

waltersom:
@eddelbuettel I have a fix for `lidR`, will get one for `metacart` as well.

eddelbuettel:
Do you have a handle on why their length differ between columns?  That seems unlikely to be on purpose.

[ And I don't recall from the top of my head whether we "help" by recycling, my instinct is to say we do not. ]

waltersom:
My fixes are at https://github.com/waltersom/lidR/tree/replace-dataframe-with-list and https://github.com/waltersom/metacart/tree/replace-dataframe-list.

For `lidR`, they don't even return the DataFrame/list, it's just a convenience, see
https://github.com/Jean-Romain/lidR/blob/9f30c6f8065b6fe6571c2d53457cfbab4a1fc3ca/src/LAS.cpp#L1214-L1215
I have to admit I don't really understand it, and it works most of the time, but there are occasions (which cause the test to fail) where they shorten a vector after adding it to the DataFrame, and it's already in another DataFrame. If we stick to a list, it works.

For `metacart`, I understand it even less. But the examples from `?FEmrt` and `?REmrt` run with this PR, and my fix to `metacart`.

Our `push_*` (as from `as.data.frame`) doesn't help by recycling, unless the length of a vector is 1.

eddelbuettel:
Nice work. I can't look at it end -- end of day / evening here.  There usual process is to propose a change to the affected package to let it build with the _existing_ (i.e. 1.0.5) and _proposed_ Rcpp (where it currently only works with _existing_).  That is usually done by email to them, politely asking for a fix in two weeks or so.  Once done we can change Rcpp itself.  

You've done so much already so let me be outrageous and ask directly:  Would you be able to take care of preparing the patches and emails, i.e. drive all that from your end?

waltersom:
I was already preparing a PR for lidR (Jean-Romain/lidR#358), and am about to look at metacart next.

In both cases, the changes should work both with or without the change to Rcpp. It is only that once this (#1099) is merged, they break without the changes.

eddelbuettel:
You put that well to `lidR` in its 357 and 358.  The fact is that `Rcpp::DataFrame` is _not_ one of our strongest classes and I personally have always grown lists---precisely what you propose here---because `DataFrame` is half baked.  Your PR will make it better so this kicks a few cans down the right road!

waltersom:
I have also emailed the metacart maintainer, so hopefully at some point
soon those two packages will be updated.

waltersom:
One thing these packages failing highlights is that the error message when attempting to construct an invalid `DataFrame` is not very helpful.
It would be possible to check if vector of different (non-1) length are being added, and either print a more helpful error, or not call `set__()` (and thus revert to previous behaviour). There would be some (small) overhead from doing this, and it's not very elegant, but if others thought it would be helpful, I could add that.

eddelbuettel:
A `warning()` or `error()` is probably a good idea.  

Food for thought: maybe make it even more complex and add an option() to fall back to old behavior?  Or do we think that is a bad idea, and that we should enforce the same length now?

waltersom:
I have implemented a function `set_type_after_push()`, which checks if any columns have a length that is different from 1, the maximum number of rows, or an integer factor of the max. In such a case, it doesn't call `set__()` (so wouldn't trigger an error), and does call `warning()`. It might be preferable at some point to change to `stop()`.

But, thinking about it, let me check what happens if one of the columns has zero length...

eddelbuettel:
Actually, using `warning()` is good because there is an existing hok for 'warnings-as-errors' that already gives us the `stop()`.

eddelbuettel:
I guess you noticed

```
 Error: ----- FAILED[data]: test_dataframe.R<112--112>
   call| expect_equal(DataFrame_PushReplicateLength(), df)
   diff| Names: 2 string mismatches 
  Execution halted
```

ChangeLog
```diff
@@ -17,6 +17,13 @@
         * inst/NEWS.Rd: Idem
         * vignettes/rmd/Rcpp.bib: Idem
         * inst/bib/Rcpp.bib: Idem
+2020-07-01 Walter Somerville <waltersom@gmail.com>
```
eddelbuettel, line 20: This should be 

- in chronological order
- with proper empty lines

eddelbuettel: I just sent a trivial commit back to you for that so make sure you pull to your box.

waltersom:
Ok, I have corrected that failing test. Thanks for doing the changelog.

Jean-Romain:
lidR maintainer here. The fix has been merged in `master`. The use of `DataFrame` was just a convenience to highlight the fact that all elements are the same length but the use of a `List` is valid as well.

Please keep me updated when you plan to release this change. I won't release a new version of lidR right now because I already made several releases recently. When you will be ready for a release of `Rcpp` I will release `lidR` ahead of time so this change won't break reverse dependency.

eddelbuettel:
Salut Jean-Romain

It works the other way around.  We cannot upload until you upload as our now-stricter-in-this-PR code breaks your package, and CRAN will not admit "breaking" packages.  We are not in a great hurry as we just released, and but the point also is to not have this linger.  Could you update within, say, two weeks which is the common timeframe CRAN itself often gives?

Jean-Romain:
I know it works the other way. This is why I suggested to tell me when you plan to release Rcpp so I can release my package first. I believed that the common time-frame given by the CRAN was more like 1 month. I can release in two weeks if mandatory.

eddelbuettel:
Rcpp will likely not have a release for many months, but that is no reason to push this off "forever".  I apologize if we were unclear, we did not mean to put a gun to your head and scream "update NOW" but as we depend upon your cooperation here it would help us if you did it "sooner rather than later".  Again, it does obviously not have to be this week.  But everybody is busy and it helps to commit to a) action and b) a date.  So if you could say "sure, but this is summer so give me three weeks" it would help us too.

Jean-Romain:
No problem. It will definitively be done within a month *maximum* and I won't delay that forever.

eddelbuettel:
@waltersom the last commits of yours and mine got out of sync so you needed a merge -- I just cleaned that up with squash rebase and force-push.   To catch you would need to reset your end a few commits and then pull.

eddelbuettel:
@Jean-Romain Thanks.  If and when you do please let us know here so that we can proceed too.

waltersom:
I have just heard from the maintainer of metacart that he has submitted a
new version to CRAN that won't fail with our changes.

eddelbuettel:
Just tested metacart 2.0-3, already on CRAN, and as expect it passes.  Nice work!

eddelbuettel:
Hi @Jean-Romain 

The last `lidR` release on CRAN was on July 5, which will be a month ago in a few days.  Given that CRAN _will_ close for the summer between August 14 and 24, I would _really_ appreciate it if you could update `lidR` later this week.  Respecting CRAN's wishes for updates only about once a month is important, but it would also be nice if you could help us let Rcpp move forward with the change from this PR.

Jean-Romain:
The new version is now on CRAN.

eddelbuettel:
@Jean-Romain Thank you!  I will proceed with Rcpp tests and changes then clearing the way for this PR.

eddelbuettel:
Thanks everybody for their patience, of course @waltersom for writing the PR in the first place, and to @jclaramunt for very promptly updating `metacart`, and to @Jean-Romain for doing the same with `lidR` (letting some time pass since its last update).

Reverse depends checks revealed no issue so this now goes into the main branch.  Yay!!!  And again, thanks to all for help!

eddelbuettel:
Thanks again for everybody's patience.  This is now 'live' in Rcpp 1.0.6 which just got to CRAN under our bi-annual release cycle.

PR #1109, opened 2020-08-20
Remove obsolete and confusing comment

ellert:
The comment was added in commit cfdb196267a8b04b2bda897c1bb7f2496bf3edcd to explain why Rcpp_list6() is special.
However, in a later commit 6eac444d92e26c185a0523d0bb6417862bb4fd30 Rcpp_list{2,3,4,5}() were added with similar definitions.
This means that Rcpp_list6() is no longer special, and the comment is now confusing rather than informative.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2020-08-20  Mattias Ellert  <mattias.ellert@physics.uu.se>
+
+        * inst/include/Rcpp/lang.h: Remove obsolete and confusing comment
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Thanks again for your PR and patience.  This is now 'live' in Rcpp 1.0.6 which just got to CRAN under our bi-annual release cycle.

PR #1118, opened 2020-11-13
Rcpp modules doc fixes

tdhock:
Hi @eddelbuettel I'm submitting a PR directly because these are simple doc fixes.
My issue was that it was not clear from the docs that when using Rcpp modules, we need to run compileAttributes to generate some RcppExports.cpp code before the package can compile/install.
I looked in the Rcpp modules vignette, and on the compileAttributes man page, but did not find any mention of this, so I added some details which would have been helpful for me.
I also noticed that compileAttributes silently does NOT write any RcppExports.cpp file, when there is already a R_init_PACKAGE declaration in some non-RcppExports.cpp C++ code file. That was confusing for me because I was expecting that it should tell me if/why it does not write the RcppExports.cpp file (and I added a warning that would have been helpful).

#### Checklist

- [ ] Code compiles correctly
- [ ] `R CMD check` still passes all tests


eddelbuettel:
The intersection of Modules and Attributes is ... at best "murky".

Ex-ante I am not really in favor of cememting it more by adding documentation.  I will try to find a moment to read what you wrote but let me remind you that the preferred approach around here is _to first discuss in issue tickets and establish consensus_ approach what if any approach to take.

eddelbuettel:
> My issue was that it was not clear from the docs that when using Rcpp modules, we need to run compileAttributes to generate some RcppExports.cpp code before the package can compile/install.

This reads to this non-native speaker as a "must" which is plain false as there are Modules packages not using RcppExports or Attributes.

eddelbuettel:
> I'm submitting a PR directly because these are simple doc fixes.

Plainly an incorrect statement as you also submit a change to `attributes.cpp`.

tdhock:
ok well no problem if you don't think these docs are useful or accurate enough to include, but some explanation along these lines sure would have helped me understand how to get Rcpp modules working.

man/compileAttributes.Rd
```diff
@@ -33,6 +33,10 @@ compileAttributes(pkgdir = ".", verbose = getOption("verbose"))
 	The \code{compileAttributes} function deals only with exporting C++ functions to R. If you want the functions to additionally be publicly available from your package's namespace another step may be required. Specifically, if your package \code{NAMESPACE} file does not use a pattern to export functions then you should add an explicit entry to \code{NAMESPACE} for each R function you want publicly available.
 
     In addition to exporting R bindings for C++ functions, the \code{compileAttributes} function can also generate a direct C++ interface to the functions using the \code{\link[=interfacesAttribute]{Rcpp::interfaces}} attribute.
+    This function also scans the C++ source code for \code{RCPP_MODULE} declarations, and generates the corresponding required entries in \code{src/RcppExports.cpp}.
+
+    If there is some C++ code file with \code{R_init_} registration code
+    already, then \code{src/RcppExports.cpp} file will \emph{not} be generated.
```
eddelbuettel, line 39: This needs refinement.  

The default mode of Rcpp Attributes does not include Rcpp Modules.  The intersection is less defined, and more "experts only".   We could document that last aspect.

src/attributes.cpp
```diff
@@ -1264,6 +1264,7 @@ namespace attributes {
                 Rcpp::CharacterVector match = initMatches[i];
                 if (match.size() > 0) {
                     hasPackageInit_ = true;		// #nocov start
+                    showWarning(pkgInit + " found in C++ code so not generating RcppExports.cpp.");
```
eddelbuettel, line 1267: I am not a fan of missing changes to code into a PR which claimed at the outset to be "light and easy and docs only".

vignettes/rmd/Rcpp-modules.Rmd
```diff
@@ -1213,6 +1213,42 @@ into the package namespace:
...
+static const R_CallMethodDef CallEntries[] = {
+    {"_rcpp_module_boot_yada", (DL_FUNC) &_rcpp_module_boot_yada, 0},
+    {NULL, NULL, 0}
+};
+RcppExport void R_init_testmod(DllInfo *dll) {
+    R_registerRoutines(dll, NULL, CallEntries, NULL, NULL);
+    R_useDynamicSymbols(dll, FALSE);
+}
+```
+
```
eddelbuettel, line 1251: See above.  Mixing Attributes and Modules is 'experts only'.

eddelbuettel (commented):
I think we should close this PR and get back to a defining what it is we want to change.

I for one do have a reference package that can serve as a model for doing both Modules and Attributes.  A possibly start may be to take the include 'reference' Module package and try to extend with Attributes into a new Modules+Attributes package.

eddelbuettel:
(The test failure is new and likely no related to the PR.  All of sudden we appear to take (much) longer in tests than before.  Why is unclear.)

tdhock:
ok I will create issues instead, sorry for the trouble.

Issue #1081, opened 2020-05-17
Rcpp is not using `R_PreserveObject`/R_ReleaseObject` as intended


ltierney:
This is based on an issue found by @bhive01 and discussed with him
and @clauswilke in wilkelab/gridtext#9.

Running the code included at the end in R versions up through 4.0.0
with the C stack set to 1M will segfault with a C stack overflow.  The
same happens (for me on Ubuntu at least) with the default 8M stack if
you change the value of N to 1000.

Running this code as is (i.e. N == 500) and with an 8M stack will not
fail, but takes 14 minutes due to several multi-minute-long pauses in
the second loop.

The problem is that this code, via calls to `R_PreserveObject` in
Rcpp, puts around 300K objects into the preserved object list. This is
a simple linked list, so has to be searched linearly to remove objects
pushed on early. That is the reason for the long pauses. The segfault
happens because the code for `R_ReleaseObject` was using recursion to
work down the list. R-devel and R-patched now use iteration and no
longer segfault, but that doesn't help anyone running R 4.0.0 or
earlier.

The _R Extensions Manual_ says about the
`R_PreserveObject`/`R_ReleaseObject` mechanism:

> It is less efficient than the normal protection mechanism, and
> should be used sparingly.

The linked list data structure, and even the recursive delete, are
reasonable for this usage but are not suited for handling 300K objects
that may need to be removed in random or first in/first-out order.

Assuming you need to keep alive R objects, a better way to do this is
to create an appropriate data structure that you control, and protect
only that data structure with `R_PreserveObject` (and maybe release it
on unload with `R_ReleseObject`). That way you can use a more
appropriate data structure, such as a hash table. Using a hash table
the code that currently takes 14 minutes would only take about 1.5
minutes. The improvement is even larger for N == 1000.

Even if you don't want to go to a hash table now it would be a good
idea to switch to using your own list so users of R 4.0.0 or earlier
don't run the risk of a segfault.

For using your own list you need something like this 

```C
static SEXP Rcpp_precious = NULL;
	
void R_init_Rcpp(DllInfo* dllinfo) {
    Rcpp_precious = CONS(R_NilValue, R_NilValue);
	R_PreserveObject(Rcpp_precious);
   ...
}
```

Then replace `R_PreserveObject` and `R_ReleaseObject` with

```c
void Rcpp_PreserveObject(SEXP object)
{
    SETCDR(Rcpp_precious, CONS(object, CDR(Rcpp_precious)));
}

[...]

Issue #382, opened 2015-10-18
provide better protection mechanism than `R_PreserveObject` / `R_ReleaseObject`?


kevinushey:
The `R_PreserveObject` / `R_ReleaseObject` mechanism basically works by placing items in a protected pairlist called `R_PreciousList`. The problem here is that having a large vector of `Rcpp` objects can lead to sub-optimal behaviour on destruction of that vector, basically because the order in which a vector's elements are destructed is not specified by the C++ standards. Example:

Suppose we have a large vector of numeric vectors, e.g. `std::vector<NumericVector> x` of size 1000000. If elements are destructed from start to end, then each destruction involves a deep recursion into the `R_PreciousList` in order to unprotect that R object. This manifests itself, for example, in https://github.com/hadley/dplyr/issues/1396.

It just so happens that, with `libc++`, vector elements are destructed in reverse order (and so we end up just popping elements off the `R_PreciousList` as desired); this does not seem to be the case with `libstdc++` (and so the performance regression shows up there, with deep recursions).

I'm not sure if Rcpp can be helpful here, but if we can I think it would be prudent to do so.

eddelbuettel:
This is now taken care of in Rcpp 1.0.6.2 via https://github.com/RcppCore/Rcpp/pull/1133.  Big thanks to everybody who helped with this, it is a very nice fix.

PR #1133, opened 2021-01-17
second pass at 'precious_{preserve,remove}' (addresses #382, #1081)

Enchufa2:
This PR takes the work in the `feature/rcpp_precious` branch and modifies it to implement O(1) deletions as recommended by @ltierney in #1081. Reviews welcome.

- `Rcpp_PreserveObject` takes an object, inserts the object in the precious list, and returns a tag.
- `Rcpp_ReleaseObject` takes a tag and removes it from the list.
- `Rcpp_ReplaceObject` has been removed.

#### Checklist

- [x] Code compiles correctly
- [x] Reprex in https://github.com/RcppCore/Rcpp/issues/1081#issuecomment-759413616 does not stall anymore.
- [x] Update file headers' copyright year.
- [x] Document the changes by file in [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -17,7 +17,7 @@
 
 	[ these were originally committed 17 May 2020 and have been rebased ]
 	* src/barrier.cpp: Implement Rcpp_precious_{init,teardown,preserve,remove}
-	* src/init.c: Register cpp_precious_{init,teardown,preserve,remove}
+	* src/rcpp_init.c: Register cpp_precious_{init,teardown,preserve,remove}
 	and call from R_{init,unload}_Rcpp
 	* inst/include/Rcpp/routines.h: Declare Rcpp_precious_{init,teardown,
 	preserve,remove}
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,9 +1,36 @@
+2021-01-17  Iñaki Ucar  <i.ucar86@gmail.com>
+
+	* inst/include/Rcpp/String.h: Use Rcpp_{Preserve,Release}Object
+	throughout which connects to Rcpp_precious_* routines
+	* inst/include/Rcpp/exceptions.h: Use Rcpp_ReleaseObject
+	* inst/include/Rcpp/routines.h: Several corrections and refinements
+	* inst/include/Rcpp/storage/PreserveStorage.h: Ditto
+	* inst/include/Rcpp/traits/named_object.h: Ditto
+	* inst/include/Rcpp/unwuindProtect.h: Ditto
+	* inst/include/RcppCommon.h: Ditto
+	* src/barrier.cpp: Ditto
+
 2021-01-17  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Version, Date): Roll minor version
         * inst/include/Rcpp/config.h: Idem
 
+	[ these were originally committed 17 May 2020 and have been rebased ]
+	* src/barrier.cpp: Implement Rcpp_precious_{init,teardown,preserve,remove}
+	* src/init.c: Register cpp_precious_{init,teardown,preserve,remove}
+	and call from R_{init,unload}_Rcpp
+	* inst/include/Rcpp/routines.h: Declare Rcpp_precious_{init,teardown,
+	preserve,remove}
+	* inst/include/Rcpp/exceptions.h: Use Rcpp_precious_remove
+	* inst/include/Rcpp/traits/named_object.h: Use
+	Rcpp_precious_{preserve,remove}
+	* inst/include/Rcpp/unwuindProtect.h: Use Rcpp_precious_preserve
+	* inst/include/RcppCommon.h: Use Rcpp_precious_{preserve,remove}
+
 2021-01-16  Dirk Eddelbuettel  <edd@debian.org>
+
+        * DESCRIPTION (Version, Date): Roll minor version
+        * inst/include/Rcpp/config.h: Idem
 
 	* tests/tinytest.R: Test for CODECOV when deciding to run extensive
 	tests or just a subset, and set CI=true to enable run full set
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2021-01-17  Iñaki Ucar  <i.ucar86@gmail.com>
+2021-01-17  Iñaki Ucar  <iucar@fedoraproject.org>
 
 	* inst/include/Rcpp/String.h: Use Rcpp_{Preserve,Release}Object
 	throughout which connects to Rcpp_precious_* routines
@@ -1514,7 +1514,7 @@
 
         * inst/include/Rcpp/traits/is_na.h: Speed up Rcpp::is_na<double>()
 
-2018-01-09  Iñaki Ucar  <i.ucar86@gmail.com>
+2018-01-09  Iñaki Ucar  <iucar@fedoraproject.org>
 
         * inst/include/Rcpp/vector/proxy.h: Improve support for NVCC, the CUDA
         compiler, by making three more methods const
@@ -2116,7 +2116,7 @@
 
         * DESCRIPTION (Version, Date): New minor version
 
-2017-02-14  Iñaki Ucar  <i.ucar86@gmail.com>
+2017-02-14  Iñaki Ucar  <iucar@fedoraproject.org>
 
         * inst/include/Rcpp/iostream/Rstreambuf.h: Fixed single-character handling
         (pull request #649, fixes issue #647)
--- a/ChangeLog
+++ b/ChangeLog
@@ -1514,7 +1514,7 @@
 
         * inst/include/Rcpp/traits/is_na.h: Speed up Rcpp::is_na<double>()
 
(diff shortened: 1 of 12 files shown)


Enchufa2:
I missed @eddelbuettel's first pass commit in my initial push. I recovered it and force-pushed.

eddelbuettel:
Lovely!  Will give it a whirl, a first Rcpp 1.0.6.1 reverse depends run (out of habit and prudence) just finished.  Will give it a whirl.  My initial attempt didn't fare too well.  Hope this does better.

> I missed @eddelbuettel's first pass commit in my initial push. I recovered it and force-pushed.

No worries. As long as a PR is clean...

eddelbuettel:
(But looking at `git ls` under all appropriate alias such as `ls = log --color --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit` makes it look butt-ugly as we have wild branches overlapping.  Should we consider scrapping this branch and cleanly applying the changes afresh from master, possibly just after #1132 is merged?)

Enchufa2:
> (But looking at `git ls` under an appropriate alias such as `ls = log --color --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit` makes it look butt-ugly as we have wild branches overlapping. Should we consider scrapping this branch and cleanly applying the changes afresh from master, possibly just after #1132 is merged?)

Ok!

eddelbuettel:
We may also be able to just rebase.  To be seen once I merge the other.

eddelbuettel:
It's interesting.  The reverse depends check starts out as blood bath just like the last time.  But with the recent experience of releasing 1.0.6 and even with the minor changes there seemingly requiring a rebuild of some other (binary) packages in the graph, I gave it a try in a "greenfield" setup freshly installing dependencies after the candidate package Rcpp 1.0.6.2 has been installed.  

And that experiment is 3 for 3 so far.  All the packages that failed with a "reused" `.libPaths()` now pass.  So will require two runs, one before and after partial reinstallation.   Add another day to it :)

Sadly, I wasn't that attentive last time so it is possible that the first pass at this wasn't as doomed as my glance at its reverse depends run suggested at first.  Anyway ... thanks again for kicking that can further down the road with a refined PR :)

Enchufa2:
Unfortunately, requiring the new data member `token` in `PreserveStorage`, `named_object<SEXP>` and `String` breaks ABI compatibility. Therefore, I took the opportunity also to remove `Rcpp_ReplaceObject`, which doesn't fit in the new scheme anymore.

eddelbuettel:
I didn't mention that none of these errors are typically as not as "in your face" as a "cannot compile" but rather come up later when tests are running.   Altering the API surface clearly needs rebuilds, which is why we do it rarely.  But there is something else much subtler going here with the _ABI_ rather than _API_ but it appears to mostly affect the packages built around stan which themselves to subsequent compilation.

eddelbuettel:
So the reverse depends run turned into the anticipated "blood bath" -- see [the summary](https://github.com/RcppCore/rcpp-logs/commit/322e3466c6b767cef11d350b5e40341010dcac99) -- but as mentioned, once packages are reinstalled it seems to pass.  

After re-installing several hundred packages from source, a subsequent run is now working and looking better.  A little over one hundred packages done and no errors yet (compared to around ten errors among the packages already tested).  So this is looking good!

eddelbuettel:
And I rebased from master with the just merged #1132 which looks cleaner too :)

Enchufa2:
Thanks for updating the ChangeLog. I didn't change the GPL copyright notice in the files' header because I'm not sure whether you prefer to just change the year, or to add a new one with a comma if they are not consecutive, or to add a new line in such a case. Also note that `PreserveStorage.h` has no copyright notice.

ChangeLog
```diff
@@ -1514,7 +1514,7 @@
 
         * inst/include/Rcpp/traits/is_na.h: Speed up Rcpp::is_na<double>()
 
-2018-01-09  Iñaki Ucar  <i.ucar86@gmail.com>
+2018-01-09  Iñaki Ucar  <iucar@fedoraproject.org>
```
eddelbuettel, line 1517: Rewriting old entries smells a little like rewriting history...

Enchufa2: I did a `sed` in place, didn't look at the diff. Reverted now.

eddelbuettel:
Thanks for updating the changelog. I was lazy and dug up an older one.  (And as noted, changing old one now is ... not what I'd encourage  ... and you already correct, appreciate that)

Feel free to add you to file authorship if you made changes in the file.   I also tend to update the end year to the current year.

Rev.dep still looking great. This will work.  But I am still very confused why it needed such a comprehensive rebuild with several hundred packages needing a rebuild.  It's a price worth paying, but I am a little, shall we say, "surprised" if not "worried" that I cannot assess when this is needed.   Will also make the next CRAN upload more involved.

inst/include/Rcpp/String.h
```diff
@@ -3,6 +3,7 @@
 // String.h: Rcpp R/C++ interface class library -- single string
 //
 // Copyright (C) 2012 - 2018  Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2021  Iñaki Ucar
```
eddelbuettel, line 6: I would probably have made that 

// Copyright (C) 2012 - 2020  Dirk Eddelbuettel and Romain Francois
// Copyright (C) 2021             Dirk Eddelbuettel and Romain Francois and Iñaki Ucar

which is for example what we have done here (from a quick grep)

Environment.h:// Copyright (C) 2009 - 2013  Dirk Eddelbuettel and Romain Francois
Environment.h:// Copyright (C) 2014 - 2020  Dirk Eddelbuettel, Romain Francois and Kevin Ushey
[...]
exceptions_impl.h:// Copyright (C) 2012 - 2019  Dirk Eddelbuettel and Romain Francois
exceptions_impl.h:// Copyright (C) 2020         Dirk Eddelbuettel, Romain Francois, and Joshua N. Pritikin

Thoughts?

inst/include/Rcpp/storage/PreserveStorage.h
```diff
@@ -1,3 +1,25 @@
+// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-
```
eddelbuettel, line 1: And these I no longer add as we switched to relying on editorconf

eddelbuettel:
Thanks for the header commit.  I put some small nits up above and may just do a quick cleanup commit in a few.  Ok with you?

Enchufa2:
Sure!

eddelbuettel (approved):
LGTM -- just awaiting a second rev.deps run for completeness

eddelbuettel:
Second pass done and looking good (see  https://github.com/RcppCore/rcpp-logs/commit/eec49e3cd9c1501dd1efe202112da604c4717f8b) so merging.

kevinushey:
Awesome work -- thanks for putting this together (and especially taking the time to tie up all the loose ends in the String class).

Sorry for coming to this late, but one comment: There may be packages using `Rcpp` that directly use `Rcpp_PreserveObject()` and `Rcpp_ReleaseObject()`. Is there a way to check if such packages exist? If they do, I think such packages might be in trouble as they would now be leaking the R objects they protect (since the old usages of the API don't know about the tokens being used, and they would then be trying to protect + unprotect the object directly rather than via the token).

Out of an abundance of caution, I would've preferred if we maintained the old API for backwards compatibility just in case, but moved to using a separate internal API for protecting + releasing Rcpp objects (using our new precious list). But we might want to defer that work until we have evidence that this is a real issue; and we could also take the stance that package authors shouldn't be using internal Rcpp APIs anyhow.

tl;dr: I think we should maintain the old implementations here:

https://github.com/RcppCore/Rcpp/blob/0ba1b550d64bbe98d622d3032acd91fc6c1436c2/inst/include/RcppCommon.h#L94-L100

but use separate `Rcpp_PreciousPreserve()` and `Rcpp_PreciousRelease()` APIs for protecting objects via Rcpp's precious list, and "eventually" encourage users who really want to manage protection manually to use those APIs.

eddelbuettel:
Good points.  The 2200+ packages in the reverse depends set (essentially: all CRAN rev.deps minus the exclusion we keep) did not mind but I agree in principle.  Not a single 'did not build issue'.  But you are correct that we could maintain the older API surface to the extend possible.

[ As an aide I am still confused as hell why some many packages (largely, though I think not exclusively, in the rstan nexus) needed a rebuild and a little lost as to why we are unable to anticipate this. ]

eddelbuettel:
I just pushed a new branch [feature/preserve_interface](https://github.com/RcppCore/Rcpp/tree/feature/preserve_interface) along those lines.

kevinushey:
Took a quick look; looks good to me. Thanks for jumping on it so quickly!

PR #1135, opened 2021-01-19
maintain existing interface and offer new one

eddelbuettel:
This does a partial revert of #1133 in the sense that we should not have altered the existing top-level API (my bad for sleeping on that, and thanks to @kevinushey for the wake-up on it).  The use of the preserve/release pair is however not all that widespread so this PR switches to using the new token / precious list via two new functions.  

Eventually, as hinted by @kevinushey in this https://github.com/RcppCore/Rcpp/pull/1133#issuecomment-763056635 we may want to rework this.  

A simpler solution may just be to gradually deprecate and remove the three old functions -- as the runs this weekend showed they do not appear to be used.  So we could keep them and demote to another internal header that gets pulled if a new `#define` is set and over time we flip is default value but let other users set it if they need it.

/cc @Enchufa2 whom I can't seem to tag under reviewers

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2021-01-19  Dirk Eddelbuettel  <edd@debian.org>
+
+        * DESCRIPTION (Version, Date): Roll minor version
+        * inst/include/Rcpp/config.h: Idem
+
+	* inst/include/RcppCommon.h: Preverse existing API and offer new
+	token-based API alongside via Rcpp_Precious{Preserve,Release}
+	* inst/include/Rcpp/String.h: Use Rcpp_Precious{Preserve,Release}
+	* inst/include/Rcpp/storage/PreserveStorage.h: Ditto
+
 2021-01-17  Iñaki Ucar  <iucar@fedoraproject.org>
 
 	* inst/include/Rcpp/String.h: Use Rcpp_{Preserve,Release}Object
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 1.0.6.2
-Date: 2021-01-17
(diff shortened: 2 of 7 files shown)


kevinushey (approved):
LGTM!

Enchufa2:
Thanks @kevinushey for raising this issue. Completely agree. This LGTM too. Just a small detail: on second pass, I believe that the calls I made to `Rcpp_precious_preserve` and `Rcpp_precious_remove` here:

https://github.com/RcppCore/Rcpp/blob/2c2eb89c8252c0790dc1ea1de9b5c82e509d340d/inst/include/Rcpp/traits/named_object.h#L40-L59

should have been `Rcpp_PreserveObject` and `Rcpp_ReleaseObject` instead, and thus `Rcpp_PreciousPreserve` and `Rcpp_PreciousRelease` in this PR under the new API.

eddelbuettel:
It's just two inlined one-liners but I agree that it nicer to see the top-level API function used -- so committed (and tested locally and pushed).

eddelbuettel:
Thanks for the quick feedback on this, and for highlighting it in the first place.  Will fold it now, and 1.0.6.3 is where we're at now.

jeroen:
FYI, this may have an (unintended?) side effect that packages built against new versions of Rcpp cannot be loaded with Rcpp 1.0.6. Hence, once this is released to CRAN, and CRAN rebuilds binaries against Rcpp 1.0.7, users that still have Rcpp 1.0.6 will get loading errors.



Screen Shot 2021-06-25 at 10 25 59 PM

eddelbuettel:
That sounds unfortunate but this #1135 fixes an earlier bug that needed fixing (#1133) in the actual Rcpp 1.0.6 release.  And users should be also fine by upgrading to Rcpp 1.0.7.

Issue #928, opened 2018-12-03
Different Rcout object for each translation unit


Enchufa2:
Opening this as per request of @kevinushey. As explained in [this thread](http://lists.r-forge.r-project.org/pipermail/rcpp-devel/2018-December/010227.html), 

> `Rcout` is defined in `iostream/Rstreambuf.h` as a static object. This means that different translation units see a different `Rcout`, while they see the same `std::cout` (I have a minimal package showing this if needed). As a result, for example, one object allocated in a certain cpp cannot redirect the output of other object allocated in another cpp.

And the same for `Rcerr`. I think that this might be worth fixing on the Rcpp side, but probably it would require to declare `Rcout` and `Rcerr` as `extern` and generate some code to initialise them in `RcppExports.cpp`.

Enchufa2:
I think this is still worth fixing.

eddelbuettel:
Sure. Any chance you could take a stab at it in branch?

Enchufa2:
Yes, I would like to. I'll try to find some time for this in the coming weeks. My approach would be to follow the design of `std::cout` if there is no complication that I'm not aware of.

Enchufa2:
There's a first pass at this here: https://github.com/Enchufa2/Rcpp/commit/d96d8063943db680aaafacd7d38714feaf92445f. I would like to ensure that this is the right approach before submitting a PR. Basically I have:

- declared `Rcout` and `Rcerr` (previously static objects) as extern references;
- defined two routines, `Rcpp_cout_get` and `Rcpp_cerr_get`, that create static objects and return a reference to them.

Now the problem is that the initialization of such references must be part of the scaffolding code. For example, in `RcppExports.cpp`, in calls to `cppFunction`, `sourceCpp`... So far, I've managed to make this work:

```r
f <- "CharacterVector ptr%s() {
  std::ostringstream address;
  address << (void const *)(&%s);
  return address.str();
}"

Rcpp::cppFunction(sprintf(f, "A", "Rcout"))
Rcpp::cppFunction(sprintf(f, "B", "Rcout"))
ptrA() == ptrB() # should be TRUE
```

So:

- Is this the best approach?
- If so, could help me identify all the places in which the initialization code should be injected? (Ideally, there should be one place to put this, but I'm not sure whether this is possible without a heavy refactoring).

Enchufa2:
BTW, a motivating example that could be used for testing:

```r
[...]

PR #1139, opened 2021-01-28
Address #928: enable global Rcout/Rcerr via RCPP_USE_GLOBAL_ROSTREAM

Enchufa2:
I think I got it. I've identified two places in `attributes.cpp` for the initialization that I believe cover all the cases. The new behaviour is enabled for Rcpp, but disabled by default for Rcpp users, guarded by `RCPP_USE_GLOBAL_ROSTREAM`. All tests pass with and without setting this in `~/.R/Makevars`.

#### Checklist

- [x] Code compiles correctly


inst/include/Rcpp/iostream/Rstreambuf.h
```diff
@@ -3,6 +3,7 @@
 // Rstreambuf.h: Rcpp R/C++ interface class library -- stream buffer
 //
 // Copyright (C) 2011 - 2017  Dirk Eddelbuettel, Romain Francois and Jelmer Ypma
```
eddelbuettel, line 5: 2011 - 2017 as the next line is 2021.

src/api.cpp
```diff
@@ -2,7 +2,7 @@
 //
 // api.cpp: Rcpp R/C++ interface class library -- Rcpp api
 //
-// Copyright (C) 2012 - 2020  Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2012 - 2021  Dirk Eddelbuettel, Romain Francois and Iñaki Ucar
```
eddelbuettel, line 5: Please it 

// Copyright (C) 2012 - 2020  Dirk Eddelbuettel and Romain Francois
// Copyright (C) 2021             Dirk Eddelbuettel, Romain Francois and Iñaki Ucar

src/attributes.cpp
```diff
@@ -1,7 +1,7 @@
 //
 // attributes.cpp: Rcpp R/C++ interface class library -- Rcpp attributes
 //
-// Copyright (C) 2012 - 2021  JJ Allaire, Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2012 - 2021  JJ Allaire, Dirk Eddelbuettel, Romain Francois and Iñaki Ucar
```
eddelbuettel, line 4: As above. You cannot retroactively acquire copyright for 2012 to 2020.

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 1.0.6.4
-Date: 2021-01-21
+Version: 1.0.6.5
+Date: 2021-01-28
```
eddelbuettel, line 4: It's probably best to let me (release manager) handle that though that is what I did locally to distinguish the rev dep run.

src/rcpp_init.cpp
```diff
@@ -1,7 +1,7 @@
 
 // Rcpp_init.cpp : Rcpp R/C++ interface class library -- Initialize and register
 //
-// Copyright (C) 2010 - 2020  John Chambers, Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2010 - 2021  John Chambers, Dirk Eddelbuettel, Romain Francois and Iñaki Ucar
```
eddelbuettel, line 4: As above.

eddelbuettel (changes requested):

inst/include/Rcpp/iostream/Rstreambuf.h
Enchufa2: Sorry, what do you mean here?

eddelbuettel: Typo. Meant 2011-2020.

```
// Copyright (C) 2011 - 2020  Dirk Eddelbuettel, Romain Francois and Jelmer Ypma
// Copyright (C) 2021             (same as above plus you)
```

As with the other comments, this PR cannot add you to prior years.  (We can and will of course adjust copyright noticed whenever we forgot someone for prior commits, but that is a different story....)

Enchufa2: Aaah, okok. I get it now. I've been thinking about this in the wrong way. :-\

ChangeLog
```diff
@@ -1,3 +1,16 @@
...
+        * DESCRIPTION (Version, Date): Roll minor version
+        * inst/include/Rcpp/config.h: Idem
+
+    * inst/include/Rcpp/iostream/Rstreambuf.h: Support for global Rcout/Rcerr by
+    adding `-DRCPP_USE_GLOBAL_ROSTREAM` to `PKG_CPPFLAGS`. This behavior mimics
+    std::cout/cerr. Required initialization code is automatically generated.
+    * inst/include/Rcpp/routines.h: Ditto
+    * src/api.cpp: Ditto
+    * src/attributes.cpp: Ditto
+    * src/rcpp_init.cpp: Ditto
```
eddelbuettel, line 12: If you have another pass over the files, maybe adjust whitespace here.  In Emacs I sometimes mark a region and run `M-x untabify` over it.  I can also also do it later....    I guess in the old days this was a simple eight-space tab but these days most of us prefer spaces. The [editorconfig](https://editorconfig.org) file may interfer too, maybe there is a better setting for `ChangeLog` files to.

(And we probably don't want to spuriously modify the whole file so don't whack it with `sed` or alike.)

src/api.cpp
```diff
@@ -20,6 +21,7 @@
 // along with Rcpp.  If not, see <http://www.gnu.org/licenses/>.
 
 #define COMPILING_RCPP
+#define RCPP_USE_GLOBAL_ROSTREAM
```
eddelbuettel, line 24: Should we `#ifndef` (or more properly `#if !defined()`) wrap this?  Or just not _in this segment_ ?  If so maybe add a comment.

eddelbuettel (commented):
That looks pretty clean.  I put two minor comments / questions in.

src/api.cpp
Enchufa2: In order to offer the posibility of using a global `Rcout` to consumers, Rcpp itself must be compiled with this option. So given that this file is responsible for setting `COMPILING_RCPP`, I made it also responsible for setting `RCPP_USE_GLOBAL_ROSTREAM` (because it's also responsible for initializing the references to `Rcout` and `Rcerr` down below).

Enchufa2:
Looking into those unit tests.

inst/tinytest/test_global_rostream.R
```diff
@@ -0,0 +1,108 @@
...
+##
+##  Rcpp is distributed in the hope that it will be useful, but
+##  WITHOUT ANY WARRANTY; without even the implied warranty of
+##  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+##  GNU General Public License for more details.
+##
+##  You should have received a copy of the GNU General Public License
+##  along with Rcpp.  If not, see <http://www.gnu.org/licenses/>.
+
+.runThisTest <- Sys.getenv("RunAllRcppTests") == "yes" && Sys.getenv("RunVerboseRcppTests") == "yes"
```
eddelbuettel, line 19: :heart:

eddelbuettel (approved):
This looks good to me but I'll probably let it sit here for a moment to let @kevinushey pipe in too.

eddelbuettel:
The test is nice work, by the way.  Apart from the gawdawful two space indenting, of course :wink:

eddelbuettel:
Ok, I'll wait til 'end of day' in whatever suitable time zone to see if @kevinushey has a comment.

kevinushey (approved):
Nice work! LGTM.

Issue #1145, opened 2021-03-21
Rcpp::message() ?


krlmlr:
I wonder if `Rcpp::message()` should exist. `Rcpp::warning()` forwards to `Rf_warning()`; since there is no `Rf_message()`, the new function would take care of calling `base::message()` from R.

eddelbuettel:
I am not sure I understand what you are writing about.   

Are you suggesting we need to add `Rcpp::message()` and implement it via a call to the R function `message()` ?

If so, why?  What current problem does this solve?

eddelbuettel:
If `RSQLite` needs a `message()` you implement a local one, just how you have a local `XPtr` header as I re-noticed the other (and is that something we should clean up together) ?

krlmlr:
Of course I could roll this myself, I'd rather not. I really think emitting a message from C/C++ code should be easier than the status quo.

The `XPtr.h` workaround was for an IDE I no longe ruse.

eddelbuettel:
Ok, now at least I know we are talking about :)   Net additions are easiest; no side effects to existing code.  I may have looked into that at some point myself: the (possible) absense of a genuine C API is a spot of bother but we do go via the R API in one or two other places.

So having now established that "yes, we could" the burden is now on you to convince Team Rcpp why we should, and burden ourselves with the test and maintenance work going forward.  We're listening.

krlmlr:
I really think emitting a message from C/C++ code should be easier than the status quo. There are various ways to solve this:

1. introduce a C API in R base
1. solve this in Rcpp and cpp11 via an R call (as proposed here)
1. implement directly in user libraries

Unless this is solved here first, I'll eventually go with the third option and propose PRs.

eddelbuettel:
I still don''t understand. 
- the "1." we have no control over as we're not R Core
- the "3." is orthogonal to what we do here and everybody can do as they please in their.

So has it taken us five messages to establish that you are proposing a C++ function with static instantiation of an R function you want to call?

And you still have not addressed the question in my previous post:  _Why would we need this_ ?

krlmlr:
I hear you're hesitant to take on new code, perhaps because that's more of a liability than an asset. If I contribute, I add tests and can respond to questions regarding the new code.

Personally, I think we need this 

- for symmetry: I typed `Rcpp::message()` without thinking and was genuinely surprised by the CI failure
- because the alternative is a mess: three or more lines of code that *could* be a simple call to `<whatever_namespace>::message()`

[...]

PR #1146, opened 2021-03-21
Add Rcpp::message

eddelbuettel:
Simple pull request to add Rcpp::message() as discussed in #1145 

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/api/meat/message.h
+++ b/inst/include/Rcpp/api/meat/message.h
@@ -1,0 +1,33 @@
+
+// message.h: Rcpp R/C++ interface class library -- Wrapper for base::message
+//
+// Copyright (C) 2021  Dirk Eddelbuettel
+//
+// This file is part of Rcpp.
+//
+// Rcpp is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// Rcpp is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Rcpp.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef Rcpp_api_meat_message_h
+#define Rcpp_api_meat_message_h
+
+namespace Rcpp {
+
+    inline void message(SEXP s) {
+        Rcpp::Function msg = Rcpp::Environment::base_env()["message"];
+        msg(s);
+    }
+
+}
+
+#endif
--- a/inst/tinytest/test_misc.R
+++ b/inst/tinytest/test_misc.R
@@ -184,5 +184,6 @@ expect_true(getRcppVersion(devel=TRUE) >= getRcppVersion(devel=FALSE), info="dev
 # expect_true(FALSE, info="oh noes")
 
 ## test that a message is output as is, and a suppressedMessage is not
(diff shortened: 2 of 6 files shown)


inst/include/Rcpp/api/meat/message.h
```diff
@@ -0,0 +1,33 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Rcpp.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef Rcpp_api_meat_message_h
+#define Rcpp_api_meat_message_h
+
+namespace Rcpp {
+
+    inline void message(SEXP s) {
```
Enchufa2, line 26: If the goal is to complete the trio `message`, `warning`, `stop`, shouldn't Rcpp provide a string-based interface as well as a variadic one (format string + args), as `warning` and `stop` do?

inst/tinytest/test_misc.R
```diff
@@ -176,9 +176,14 @@ expect_equal(testNullableString("blah"), "blah")
...
-expect_true(getRcppVersion(devel=TRUE) >= getRcppVersion(devel=FALSE), info="dev greater equal release")
+expect_true(inherits(Rcpp::getRcppVersion(), "package_version"), info="package_version object")
+expect_true(Rcpp::getRcppVersion(devel=TRUE) >= Rcpp::getRcppVersion(devel=FALSE), info="dev greater equal release")
 
 ## if need be it can be useful to fail to test e.g. the Docker setup
 ## commented out now as we prefer to pass when not debugging ;-)
 # expect_true(FALSE, info="oh noes")
+
+## test that a message is output as is, and a suppressedMessage is not
+expect_equal(capture.output(messageWrapper("ABCdef"), type="message"), "ABCdef")
```
kevinushey, line 187: FWIW I think it would be a bit more idiomatic to capture test against the condition object directly with something like:

```
msg <- tryCatch(message("test"), message = identity)
expect_equal(msg$message, "ABCdef\n")
```

eddelbuettel: Interesting. I don't think I have ever seen `tryCatch(... , message = identity)` before so in that sense the suggested code it harder to read for me :)

eddelbuettel: I added this now, and made the (common) payload a variable at the same time:

```r
## test that a message is output as is, and a suppressedMessage is not
txt <- "ABCdef"
expect_equal(capture.output(messageWrapper(txt), type="message"), txt)
expect_equal(capture.output(suppressMessages(messageWrapper(txt)), type="message"), character())
expect_message(messageWrapper(txt))
## test for message component
msg <- tryCatch(message(txt), message = identity)
expect_equal(msg$message, paste(txt, "\n", sep=""))
```

inst/include/Rcpp/api/meat/message.h
Enchufa2: Just in case you didn't see this one. :point_up:

eddelbuettel: I did see it.  Minimal does it for now, for me at least.

eddelbuettel: `Rcpp::message()` is a band aid, and we agreed that a band aid is good enough. I don't feel compelled to dress it up.

As always, the sky is limit, but incremental approaches work.  If we think this one is good to merge, we merge it.  Afterwards is someone (you?) feels like expanding we can discuss that question and approach.

eddelbuettel:
Quick poke just to say that I think objections have been addressed to I suggest to proceed and merge later. Please holler if you disagree...

kevinushey (approved):

Enchufa2 (approved):

eddelbuettel:
Thanks guys!

Issue #1152, opened 2021-04-09
Signed/unsigned comparison in IndexHash


Bijaelo:
This is a reboot of an old [issue from 2016 #574](https://github.com/RcppCore/Rcpp/issues/574), that seems to be present once again while compiling under windows. The compiler throws a warning after compiling the below code. Assume the below minimum reproducible example is located in the working directory in a file called **test.cpp**.  Calling `sourceCpp("test.cpp")` results in a warning.

Snippet from the warning:

> C:/Users/olive/Documents/R/win-library/4.0/Rcpp/include/Rcpp/hash/IndexHash.h:103:35: warning: comparison of integer expressions of different signedness: 'unsigned int' and 'int' [-Wsign-compare]. 

Session info included at the very bottom.

```c++
#include <Rcpp.h>
using namespace Rcpp;

inline bool allin(const SEXP& x, const SEXP& table){
  if(TYPEOF(x) != TYPEOF(table))
    stop("x and table must be of the same type.");
  switch(TYPEOF(x)){
    case INTSXP:
      return is_true(all(Rcpp::in(Vector<INTSXP>(x), Vector<INTSXP>(table))));
    case REALSXP:
      return is_true(all(Rcpp::in(Vector<REALSXP>(x), Vector<REALSXP>(table))));
    case STRSXP:
      return is_true(all(Rcpp::in(Vector<STRSXP>(x), Vector<STRSXP>(table))));
    default:
      stop("Invalid type in allin [Type = %s, valid = (%s, %s, %s)]",
           Rf_type2char(TYPEOF(x)),
           Rf_type2char(INTSXP),
           Rf_type2char(REALSXP),
           Rf_type2char(STRSXP));
  }
}

// [[Rcpp::export]]
void testStuff(const SEXP& x, const SEXP& table){
  Rcout << allin(x, table) << '\n';
}

/***R
# Example call: Unnessary for the issue
x <- 1:6
table <- 1:20
testStuff(x, table) 
*/
```

```R
>sessionInfo()
R version 4.0.3 (2020-10-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19042)

Matrix products: default

locale:
[1] LC_COLLATE=English_Denmark.1252  LC_CTYPE=English_Denmark.1252    LC_MONETARY=English_Denmark.1252 LC_NUMERIC=C                     LC_TIME=English_Denmark.1252    

attached base packages:
[...]

PR #1153, opened 2021-04-09
Suppress one 'int / unsigned int comparisom' warning

eddelbuettel:
This addresses the (spurious but tedious) one warning in #1152, and while we are at it, updates _in that file alone_ the few uses of `unsigned int` to `uint32_t`.   The second part is more debatable but should be harmless and does of course test out fine.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 2021-04-09  Dirk Eddelbuettel  <edd@debian.org>
 
-	* inst/include/Rcpp/hash/IndexHash.h: Silence one comparison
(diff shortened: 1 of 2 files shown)


Enchufa2 (approved):
LGTM. Due to their similarities, maybe it's a good opportunity to switch to `uint32_t` in `SelfHash.h` too.

eddelbuettel:
> Due to their similarities, maybe it's a good opportunity to switch to uint32_t in SelfHash.h too.

Agreed. I already did a recursive grep and it is very few files.  But not urgent, and I wanted to keep this small.  

Thanks for the review!

Issue #1158, opened 2021-04-25
Moving towards defining STRICT_R_HEADERS in Rcpp.h


eddelbuettel:
As previously discussed in the (preliminary, unmerged) pull requests #613 and #615 as well as in issues #612 and (at much greater length) in issue #898, it is desirable for compilations of R extensions to define `STRICT_R_HEADERS` so that terms like `ERROR` or `MESSAGE` do not clash due to a redefinition, or that `PI` remains undefined, and of course so that `Calloc()`, `Realloc()` or `Free()` are not defined.

However, as the use of the definition has always been an opt-in, some packages have been making use of the definitions that would otherwise have been defined away -- so that we now cannot simply define it as they would then error.  The issue ticket #898  discusses two attempts at quantifying just how extended the damage at CRAN would be: about 58 packages in late Sep 2018, about 62 packages in Jul 2019, and and around 51 earlier this month.  Based on the earlier discussions some packages did make changes, and that is greatly appreciated.

And after some more discussion in #898 today, the time seems right to try to address this, one-by-one, in the affected packages.  There is no rush: if this takes us six or eight or ten months, so be it.  It would be nice to get it done, so let's try to make a start.

What follows is a list by package. The list may be wrong, and if a package is included by accident or error in our process we will remove it.  The list may be incomplete: if we find other packages we will add them.  Otherwise it will list the names of packages that did come up as failing to pass `R CMD check` when `STRICT_R_HEADERS` was defined by us, and the error is almost always one of an immediate failure to compile.  A simple re-definition of the affected symbols will be the solution, and we will aim to provide pull requests as we go along.  Help in that process would be appreciated.

So give this list, the suggested plan is to submit PRs.  And then to follow-up as PRs are folded in addressing the issue, and to 'tick off' each package and maybe mark the version of the package containing the fix and PR.  Let's see how it goes, and discuss below what we learn.

- [x] activegp (1.0.5) -- not on GitHub, patch emailed 2021-04-25, ack from maintainer 2021-04-25, 'real soon', on CRAN 2021-11-30
- [x] atakrig (0.9.7) -- ~not on GitHub~ at maoguihu/atakrig, patch emailed 2021-04-25, [PR #1](https://github.com/maoguihu/atakrig/pull/1) send 2021-04-26 too, ack from maintainer 2021-04-26, 'soon', 0.9.8 on CRAN 2021-04-30
[...]

PR #1162, opened 2021-05-09
add cfloat header (fixes #1161)

### LINKED ISSUE #1161: We should include the <cfloat> header file by default
One thing became apparent during the work on #1158 for `STRICT_R_HEADERS`: a majority of patches / PRs simply require an `#include` of `float.h` (or, equivalently, `cfloat`).  

Is there any reason we should not add that to `RcppCommon.h`?  

We already include `<limits>` which, as I re-learned, only covers `integer` limits to `<cfloat>` would make sense.

eddelbuettel:
As seen in #1158 and described in #1161, we are apparently missing `cloat` (or `float.h`) for numeric limits.  This small commit addresses this.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,11 @@
+2021-05-09  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/RcppCommon.h: Add cfloat header
+
 2021-04-09  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/include/Rcpp/hash/IndexHash.h: Silence one comparison, update
-	use of unsigned it to uint32_t throughout
+	use of unsigned int to uint32_t throughout
 
 2021-03-22  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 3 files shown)


eddelbuettel:
No objections by anyone so will likely merge this 'soon' ...

kevinushey:
LGTM as well.

My only question is whether this could cause problems on "certain systems" if the user had used `#include <float.h>` before `#include <Rcpp.h>`, but I suspect it will be fine.

eddelbuettel:
I think all 'library' header files have include guards.  Having worked on #1158 I have seen quite a few 'include header sections' in the wild and I don't think there is an issue.

eddelbuettel:
Oh, and I just ran a full rev.dep check with `#include <cfloat>` added. I haven't looked at in detail but it does not seem to have set the world on fire :)

eddelbuettel:
I just committed the [summary of that rev.dep run](https://github.com/RcppCore/rcpp-logs/commit/46f78648c08c073bf90808af767311ed73f17351) and nothing new came up.  

As usual a few unrelated issues at CRAN but by and large the number of packages affected by adding the desired `STRICT_R_HEADERS` (cf #1158) comes down so moving in the right direction

PR #1163, opened 2021-05-27
Allow 'five element version' numbers during development

eddelbuettel:
We have used a scheme of three-element releases (such as the current 1.0.6) along with four-element development releases (currently 1.0.6.7 as the newest in the drat repo).  While testing STRICT_R_HEADERS I wanted to differentiate with/without the define tarballs ... and could not as our testing code is a little too rigid.

This PR addresses this.  No other changes, no actual functionality changes.

#### Checklist

- [x] Code compiles correctly
- [x] Prefereably, new tests were added which fail without the change (tests were updated)


kevinushey (approved):
LGTM!

PR #1173, opened 2021-09-05
update old examples with respect to RNGScope (closes #1172)

### LINKED ISSUE #1172: Remove RNGScope from examples
Several example still use an explicit `RNGScope` object; this has not been needed for _many_ years as it is added in the wrapper code from Rcpp Attributes.  (So this was likely a translation error from pre-Attributes code.)  I will make a pass and clean that up.

eddelbuettel:
A fresh entry for the Rcpp Gallery came with explicit uses of RNGScope, something we have not needed 'since forever' so I took a quick look at the examples and a) removed one entry and b) used stronger language on the old inline examples.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/examples/RcppGibbs/timeRNGs.R
+++ b/inst/examples/RcppGibbs/timeRNGs.R
@@ -3,6 +3,105 @@ suppressMessages(library(Rcpp))
 suppressMessages(library(inline))
 suppressMessages(library(rbenchmark))
 
+## NOTE: Within this section, the new way to compile Rcpp code inline has been
+## written. Please use the code next as a template for your own project, and
+## no NOT use the old code below
+
+cppFunction('
+NumericVector rcppGamma(NumericVector x){
+    int n   = x.size();
+
+    const double y = 1.234;
+    for (int i=0; i<n; i++) {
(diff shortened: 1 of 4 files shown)


inst/examples/RcppGibbs/timeRNGs.R
```diff
@@ -3,111 +3,36 @@ suppressMessages(library(Rcpp))
...
-
-  // Return to R
-  return x;
-')
-
-
 ## NOTE: Within this section, the new way to compile Rcpp code inline has been
-## written. Please use the code next as a template for your own project.
+## written. Please use the code next as a template for your own project, and
+## no NOT use the old code below
```
Enchufa2, line 8: *do NOT

Enchufa2 (approved):
Apart from the typo, LGTM.

inst/examples/RcppGibbs/timeRNGs.R
eddelbuettel: Fixed.  That was just a block I moved around so the typo has been there for years too :)

PR #1180, opened 2021-09-29
Address #1108: enable finalizeOnExit via RCPP_USE_FINALIZE_ON_EXIT

### LINKED ISSUE #1108: Add explicit finalizer signature to XPtr
In one recent project, the (default) `standard_delete_finalizer()` for an `XPtr` object was seen to not get called regularly, at least not in all settings leading to some complains from _e.g._ `valgrind` under full instrumentation.

The alternative, in that case, was to set explicit calls to `R_RegisterCFinalizerEx()` directly which we could easily add as a new alternate signature to `XPtr` too.

Enchufa2:
Building on #656, closes #1108. As discussed, enables switching default behaviour to `finalizeOnExit=true` by defining `RCPP_USE_FINALIZE_ON_EXIT`.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2021-09-29  Iñaki Ucar  <iucar@fedoraproject.org>
+
+	* inst/include/Rcpp/XPtr.h: Check for new define
+        `RCPP_USE_FINALIZE_ON_EXIT` to flip the value of XPtr's `finalizeOnExit`
+        parameter from false (default) to true.
+	* inst/tinytest/test_xptr.R: Added test for this functionality.
+
 2021-09-27  Dirk Eddelbuettel  <edd@debian.org>
 
 	* README.md: Added total downloads badge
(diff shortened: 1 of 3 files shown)


eddelbuettel (approved):
LGTM!

Enchufa2:
Some platforms fail, I suppose, because the image has an old version of tinytest which doesn't include `expect_stdout`. Do you want me to change this in favour of `expect_equal(capture.output(...))`?

eddelbuettel:
Alternatively, just do (as I often do)     `if  (getVersion() < "3.6.0") exit_file("Skip remainder on older test platform")`       which has the advantages of keeping our main code cleaner / newer as we want it and we don't waste time testing on a less-relevant platform.    We do know the underlying test is fine; life is simply too short to spend the wrong part of the 80/20 rule on arcane CI issues.   (And I don't know what the cutoff is; you can also directly test   `packageVersion("tinytest")`   for suitable cut-off.)

PR #1181, opened 2021-10-02
Add action for containerbuilds

eddelbuettel:
This PR adds a simple-enough workflow to create containers used for CI as well two convenience add-ons.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,36 @@
+name: docker
+
+on:
+  push:
+  schedule:
+  - cron: "32 1 * * 6"
+
+jobs:
+  docker-ci:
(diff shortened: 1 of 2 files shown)


Enchufa2 (approved):
LGTM. Note that, if you wish, it could be further simplified by stacking together the build&push steps for `run` and `plus` in the same job, just one after the other:

```
    - name: Build and push
      uses: docker/build-push-action@v2
      with:
        push: true
        context: docker/run
        tags: rcpp/run

    - name: Build and push
      uses: docker/build-push-action@v2
      with:
        push: true
        context: docker/plus
        tags: rcpp/plus
```

eddelbuettel:
Fair point. I will consider that for another revision.

PR #1183, opened 2021-10-11
Adding Rcpp signature attribute

traversc:
### Pull Request Template for Rcpp

Please explain the changes you want to apply to Rcpp, preferably in an issue ticket **before** you create a pull request. See the file [Contributing](https://github.com/RcppCore/Rcpp/blob/master/Contributing.md) and the other templates for details.

#### Checklist

- [ ] Code compiles correctly
- [ ] `R CMD check` still passes all tests
- [ ] Prefereably, new tests were added which fail without the change
- [ ] Document the changes by file in [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog)

Issue #1182, opened 2021-10-10
Additional Rcpp attributes


traversc:
I'd like to see a "signature" attribute that allows you to customize the R function signature in `RcppExports.R`.

I modified `src/attributes.cpp` [here](https://github.com/RcppCore/Rcpp/compare/master...traversc:master) to be able to do this. 

The syntax is like this:

```
// [[Rcpp::export( signature = {verbose = getOption("verbose")} )]]
void test1(bool verbose) {
  Rcout << "The verbose paramter is set to: ";
  Rcout << verbose << std::endl;
}
```

And in `RcppExports.R` it looks like this:

```
test1 <- function(verbose = getOption("verbose")) {
    invisible(.Call(`_RcppTest_test1`, verbose))
}
```

I wrote a bunch of [tests](https://github.com/traversc/RcppTest/blob/master/src/rcpp_test.cpp) to make sure it doesn't break. Here's a more complicated example using `sourceCpp`:

```
library(Rcpp)
sourceCpp(code='
#include <Rcpp.h>
using namespace Rcpp;
// [[Rcpp::export( rng = false, signature = {x = list("{A}", "B"), verbose = getOption("verbose")} )]]
int test_inline(List x, bool verbose) {
  Rcout << "The verbose paramter is set to: ";
  Rcout << verbose << std::endl;
  if(x.size() > 0) {
    CharacterVector first_element = x[0];
    Rcout << "The first element of list x is: ";
    Rcout << first_element << std::endl;
  }
  return x.size();
}')

test_inline()
# The verbose paramter is set to: 0
# The first element of list x is: "{A}"
# [1] 2
options(verbose=TRUE)
test_inline(list(1))
# The verbose paramter is set to: 1
# The first element of list x is: "1"
# [1] 1
```

The benefit of the "signature" attribute is that it lets you add arbitrary default parameters (lists, functions, environments, getOption, etc.). Yes you could write a wrapper function, but it would be nice to write it all inline. 

Any interest in a pull request?

traversc:
Btw when you say that it could be fragile, do you mean it could break something internally, or that it would let users to shoot themselves in the foot? :P

jjallaire:
This seems like a very nice addition to me. In terms of breakage, I know we use attributes a lot in tests but I don't recall how thoroughly the tests cover all the tributaries of attribute parsing. Generally though it seems like a revdep check would catch anything material (as if there were incorrect parameters generated anywhere presumably the package tests would fail).

[...]

PR #1184, opened 2021-10-11
Adding Rcpp signature attribute

eddelbuettel:
This PR extends / revisits #1183 which got close for technical reasons.  See #1182 for discussion motivating it.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/src/attributes.cpp
+++ b/src/attributes.cpp
@@ -400,8 +400,11 @@ namespace attributes {
             Param sigParam = paramNamed(kExportSignature);
             std::string sig = sigParam.value();
             trimWhitespace(&sig);
+            if(sig.empty()) return sig;
             if(sig.back() == '}')
                 sig = sig.substr(0, sig.size()-1);
+            // check sig.empty again since we deleted an element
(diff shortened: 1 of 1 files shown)


eddelbuettel:
So as expected, the reverse dependency check did not reveal new issues caused by this PR. It is a new code path, it alters an existing access pattern that already is extremely lightly to not all used in standard tests.  So no issue there.

But @Enchufa2 makes good points above.  We should probably add a bit more unit testing to this before going any further.  @travers, do you think you can make a few tests operational on this?

traversc:
Sorry, I didn't think those questions were directed to me. 

The first comment on checking `sig.empty` makes sense. 

The question on `-1` vs. `std::string::npos` is stylistic. I think you'd want to make it consistent with the previous line. 

The question on error proagation: I believe an error from `Rcpp::function` propagates correctly, as expected. But how errors from `Rcpp::function` work is not 100% clear to me. 

Let me know what you guys decide and I can write another pull request.

PR #1185, opened 2021-10-13
Pr/1183

traversc:
I added some syntax checking based on @Enchufa2 's feedback.

Tests here: https://github.com/traversc/RcppTest/tree/master/inst. These tests incorporate the examples from @Enchufa2. 

Currently, you can run them with `make install && make test`, but I'm not sure the best way to add the tests as a pull request.

### DIFF
--- a/inst/tinytest/test_attribute_package.R
+++ b/inst/tinytest/test_attribute_package.R
@@ -1,5 +1,5 @@
-##  Copyright (C) 2010 - 2019  Dirk Eddelbuettel and Romain Francois and
-##                             Travers Ching
+##  Copyright (C) 2010 - 2020  Dirk Eddelbuettel and Romain Francois
+##  Copyright (C) 2021         Dirk Eddelbuettel, Romain Francois and Travers Ching
 ##
 ##  This file is part of Rcpp.
 ##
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Me too, of course. Alwats happy to help with `tinytest` which, via its "it is just a script" approach really makes it easy.

Enchufa2:
@traversc Checking your `RcppTest` repo, I see that you made a package specifically to test this. So the idea would be to integrate those tests into [`inst/tinytest`](https://github.com/RcppCore/Rcpp/tree/master/inst/tinytest) as part of this PR, removing the manual `print`s, `stopifnot`s, etc., and replacing them with the proper `tinytest` utilities that `Rcpp` already uses. You can take a look there to see how other similar tests are implemented.

Enchufa2:
Looking very good, thanks for all the changes! I left a few more comments. I also activated the CI and everything passes (don't worry about the covr report).

> @Enchufa2 Replies inline. A bunch of commits for minor things, but hopefully they can be squashed at the end.

Sure, no problem.

> Also, one comment: I noticed the ChangeLog is a mixture of spaces and tabs -- is that intentional?

Yeap, I asked the same thing some time ago, but I don't recall @eddelbuettel's answer. :)

traversc:
Updated, hopefully I addressed everything properly this time :)

eddelbuettel:
Yep that's the idea: by squashing it into our pr/1183 branch the pending PR #1184 should auto-update.

Giving that a whirl now.

Issue #1188, opened 2021-11-11
CxxFlags (and friends) should quote their library paths on Unix


kevinushey:
Originally filed from here: https://github.com/rstudio/renv/issues/862

The ultimate issue is that `Rcpp:::CxxFlags()` doesn't quote the include path used, so packages installed into a path containing spaces (I know, I know...) won't be propagated to the compiler correctly. For example:

```
# use library path with spaces
libpath <- tempfile("Has Spaces")
dir.create(libpath)
.libPaths(libpath)

# install Rcpp
install.packages("Rcpp")

# ask it for CxxFlags
Rcpp:::CxxFlags()
```

gives me, on macOS:

```
> Rcpp:::CxxFlags()
-I/private/var/folders/b4/2422hswx71z8mgwtv4rhxchr0000gn/T/RtmpyPtnPl/Has Spaces82462dc1ebe6/Rcpp/include
```

Note that the include path is not quoted, which implies that installing other packages from sources which try to use `Rcpp:::CxxFlags()` will fail. For example, with `install.packages("scrypt", type = "source")`:

```
< ... >
clang++ -mmacosx-version-min=10.13 -std=gnu++14 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I'/private/var/folders/b4/2422hswx71z8mgwtv4rhxchr0000gn/T/RtmpyPtnPl/Has Spaces82462dc1ebe6/Rcpp/include' -I/usr/local/include  `/Library/Frameworks/R.framework/Resources/bin/Rscript -e "Rcpp:::CxxFlags()"` -I./scrypt-1.1.6/ -I./scrypt-1.1.6/lib -I./scrypt-1.1.6/lib/util -DHAVE_CONFIG_H -fPIC  -g -O3 -pipe -Wall -pedantic -c RcppExports.cpp -o RcppExports.o
clang: error: no such file or directory: 'Spaces82462dc1ebe6/Rcpp/include'
make: *** [RcppExports.o] Error 1
ERROR: compilation failed for package ‘scrypt’
* removing ‘/private/var/folders/b4/2422hswx71z8mgwtv4rhxchr0000gn/T/RtmpyPtnPl/Has Spaces82462dc1ebe6/scrypt’
Warning in install.packages :
  installation of package 'scrypt' had non-zero exit status
```

The `scrypt` package has this in its Makevars:

```
[...]

PR #1189, opened 2021-11-11
quote path in CxxFlags

kevinushey:

Closes https://github.com/RcppCore/Rcpp/issues/1188.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/R/RcppLdpath.R
+++ b/R/RcppLdpath.R
@@ -52,7 +52,7 @@ RcppCxxFlags <- function(cxx0x=FALSE) {
     if (.Platform$OS.type=="windows") {
         path <- asBuildPath(path)				# #nocov
     }
-    paste("-I", path, if (cxx0x && canUseCXX0X()) " -std=c++0x" else "", sep="")
+    paste0('-I"', path, if (cxx0x && canUseCXX0X()) '" -std=c++0x' else '')
 }
 
 ## Shorter names, and call cat() directly
--- a/R/RcppLdpath.R
+++ b/R/RcppLdpath.R
@@ -47,12 +47,16 @@ canUseCXX0X <- function() { TRUE }  # .Call( "canUseCXX0X", PACKAGE = "Rcpp" )
 
 ## Provide compiler flags -- i.e. -I/path/to/Rcpp.h
 RcppCxxFlags <- function(cxx0x=FALSE) {
-    # path <- RcppLdPath()
+
     path <- Rcpp.system.file( "include" )
     if (.Platform$OS.type=="windows") {
         path <- asBuildPath(path)				# #nocov
     }
-    paste0('-I"', path, if (cxx0x && canUseCXX0X()) '" -std=c++0x' else '')
+
+    paste0(
+        '-I"', path, '"',
+        if (cxx0x && canUseCXX0X()) ' -std=c++0x' else ''
+    )
 }
 
 ## Shorter names, and call cat() directly
--- a/R/RcppLdpath.R
+++ b/R/RcppLdpath.R
@@ -47,16 +47,11 @@ canUseCXX0X <- function() { TRUE }  # .Call( "canUseCXX0X", PACKAGE = "Rcpp" )
 
 ## Provide compiler flags -- i.e. -I/path/to/Rcpp.h
 RcppCxxFlags <- function(cxx0x=FALSE) {
-
     path <- Rcpp.system.file( "include" )
     if (.Platform$OS.type=="windows") {
         path <- asBuildPath(path)				# #nocov
     }
-
-    paste0(
-        '-I"', path, '"',
-        if (cxx0x && canUseCXX0X()) ' -std=c++0x' else ''
-    )
+    paste0('-I"', path, '"')
 }
 
 ## Shorter names, and call cat() directly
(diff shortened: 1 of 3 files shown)


R/RcppLdpath.R
```diff
@@ -47,12 +47,16 @@ canUseCXX0X <- function() { TRUE }  # .Call( "canUseCXX0X", PACKAGE = "Rcpp" )
...
 RcppCxxFlags <- function(cxx0x=FALSE) {
-    # path <- RcppLdPath()
+
     path <- Rcpp.system.file( "include" )
     if (.Platform$OS.type=="windows") {
         path <- asBuildPath(path)				# #nocov
     }
-    paste("-I", path, if (cxx0x && canUseCXX0X()) " -std=c++0x" else "", sep="")
+
+    paste0(
```
eddelbuettel, line 56: Can you make that line, please, to remain consistent with the rest of the code of the package?

R/RcppLdpath.R
```diff
@@ -47,12 +47,16 @@ canUseCXX0X <- function() { TRUE }  # .Call( "canUseCXX0X", PACKAGE = "Rcpp" )
...
+        '-I"', path, '"',
+        if (cxx0x && canUseCXX0X()) ' -std=c++0x' else ''
```
eddelbuettel, line 58: Shall we prune this while we're add it?  R's minimum these days is C++11, and will soon be C++14.

eddelbuettel (changes requested):

R/RcppLdpath.R
eddelbuettel: Updated / simplified.

R/RcppLdpath.R
eddelbuettel: Ditto.

eddelbuettel (approved):

eddelbuettel:
It's a wee bit circular but I just approved the changes I made ... after requesting them first :)   Will let it sit for a bit and merge if nobody complains...

kevinushey:
Thanks! I approve of your changes as well, not that you needed it here. :-)

stefanoborini:
Is it possible that this change has compromised building on macOS monterey and R 4.0.1? I am seeing a failure on 1.0.10 that was not occurring on 1.0.7:

```
clang++ -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/usr/local/include  `/Library/Frameworks/R.framework/Resources/bin/Rscript -e "Rcpp:::CxxFlags()"` -fPIC  -Wall -g -O2  -UNDEBUG -Wall -pedantic -g -O0 -c src/myfile.cc -o myfile.o
Using environment default (R version: 4.0.4, platform: x86_64-apple-darwin17.0)
src/myfile.cc:2:10: fatal error: 'Rcpp.h' file not found
#include <Rcpp.h>
         ^~~~~~~~
```

The content of src/Makevars is as follows:

```
## Use the R_HOME indirection to support installations of multiple R version
PKG_LIBS = `$(R_HOME)/bin/Rscript -e "Rcpp:::LdFlags()"`
PKG_CXXFLAGS=`$(R_HOME)/bin/Rscript -e "Rcpp:::CxxFlags()"`
```

I am unsure if this is due to the addition of the quotes, but it seems so. In 1.0.7 I obtained:

```
$ clang++ -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG   -I/usr/local/include  `/Library/Frameworks/R.framework/Resources/bin/Rscript -e "Rcpp:::CxxFlags()"` -fPIC  -Wall -g -O2  -UNDEBUG -Wall -pedantic -g -O0 -c src/myfile.cc -o myfile.o
Using environment default (R version: 4.0.4, platform: x86_64-apple-darwin17.0)
In file included from src/myfile.cc:2:
In file included from <redacted>/.envs/default/lib/Rcpp/include/Rcpp.h:27:
In file included from <redacted>/.envs/default/lib/Rcpp/include/RcppCommon.h:167:
<redacted>/.envs/default/lib/Rcpp/include/Rcpp/internal/r_coerce.h:255:7: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]

<compilation continues and finishes>
```

eddelbuettel:
Why do you use `Rscript -e "Rcpp:::CxxFlags()"` which was deprecated around a decade ago?  If you do non-standard things, and it breaks, you get to keep the pieces.  Maybe try a quick `Rcpp.package.skeleton("newpackage")` and compare the basic setup to what you use, you are likely better off using what everybody else is using as it is more tested.

eddelbuettel:
You appear to want to have it both ways:  Use _your code_ from ten years ago unchanged, yet rely an brand-new code?  Is see a disconnect here.   _Many_ packages among the 2600+ on CRAN that use Rcpp also started a long time ago, yet managed to update their processes.

In any event, we (and CRAN) test rigourously, and I think most people find that playing with (rather than against) that setup works for them.  If you manage to isolate a breakage, please try to reproduce it in a minimal setup on current versions.

stefanoborini:
I don't deploy on CRAN. All I know is that I get dumped a 10 years old package on my lap and have to make it work, and all I know is that with 1.0.7 of Rcpp it worked and with a patchlevel version bump it doesn't.

PR #1191, opened 2021-12-08
Additional headers for easy-to-use light(er,est) entry points

eddelbuettel:
This PR adds three new headers (along with a new alternate Rcpp/Rcpp pulling in Rcpp.h) to offer (hopefully) easier access to (simple) speedups from unused headers / features. 

An older work-in-in-progress branch, renamed featires/new_headers_initial, has the original commits as well as two more files which I do not think add value, hence this new (smaller) branch.

As new entry points, no existing code is changed or affected.

I had been sitting on this for months, undecided, but when I mentioned it recently @kevinushey was in favour so there it for review. 

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2021-12-08  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/Rcpp: Added as new entry point
+	* inst/include/Rcpp/Light: Added as lighter-weight entry point
+	* inst/include/Rcpp/Lighter: Idem
+	* inst/include/Rcpp/Lightest: Idem
+
 2021-11-30  Dirk Eddelbuettel  <edd@debian.org>
 
 	* README.md: Remove Travis badge
--- a/inst/include/Rcpp/Light
+++ b/inst/include/Rcpp/Light
@@ -1,4 +1,22 @@
-// New (draft) header with optional components off
+// Rcpp/Light: R/C++ interface class library -- without Rcpp Modules
+//
+// Copyright (C) 2008 - 2021  Dirk Eddelbuettel
+// Copyright (C) 2009 - 2021  Dirk Eddelbuettel and Romain Francois
+//
+// This file is part of Rcpp.
+//
+// Rcpp is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// Rcpp is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Rcpp.  If not, see <http://www.gnu.org/licenses/>.
 
 // no Modules
 #define RCPP_NO_MODULES
--- a/inst/include/Rcpp/Light
+++ b/inst/include/Rcpp/Light
@@ -1,0 +1,7 @@
+// New (draft) header with optional components off
+
+// no Modules
+#define RCPP_NO_MODULES
+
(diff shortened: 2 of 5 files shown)


kevinushey:
IMHO rather than having 3 separate headers we should just have a single header `RcppLite.h` (or similar) that defines `RCPP_NO_MODULES` + `RCPP_NO_SUGAR`, as those are the heaviest optional pieces of Rcpp.

If we keep this scheme, I think RTTI is actually less heavyweight (for compile time, at least) than sugar.

Enchufa2 (approved):
> IMHO rather than having 3 separate headers we should just have a single header `RcppLite.h` (or similar) that defines `RCPP_NO_MODULES` + `RCPP_NO_SUGAR`, as those are the heaviest optional pieces of Rcpp.

But some Rcpp sugar is so convenient. :)

As I said previously, I'm not sure what's the best way to address this, so if you both think this is the best option, then go ahead. Anyway these are my thoughts in no particular order:

- I think the important thing here is discoverability. There's a FAQ point about this which briefly mentions how to disable modules, mentions RTTI without explanation, and doesn't mention sugar. I don't think many Rcpp users ever read that, and of those who did, I would bet they didn't know what's the meaning nor the implications.
- That said, I don't see how definining new headers inside the `Rcpp` folder, next to many others, improves discoverability, and much less understanding of the implications. I agree that includes may be nicer than defines, but the good thing about the current defines is that they say explicitly what they're doing (i.e., _no modules_, _no sugar_, and even _no RTTI_, which I don't think many Rcpp users knows what this means). On the other hand, Light, Lighter, Lightest is much more obscure.
- Another option (instead of Light, Lighter and Lightest) would be to define _Base_ as no modules + no RTTI, and _Minimal_ as Base + no sugar. How many users would want RTTI but no modules? Maybe add `RCPP_ENABLE_RTTI` for them? Anyway, I wouldn't define a header for RTTI, because, unlike modules and sugar, it really is a flag.
- Regardless what we do here, I'm skeptical about general adoption. Rcpp is very well established, and there are sooo many resources, packages and examples out there with just `#include <Rcpp.h>`. Another approach would be to target disabling those by default in a future release, and create headers to enable modules and sugar. Then contact the maintainers using those features to explicitly enable them with new defines. But probably you've got enough with the strict headers stuff. :)

eddelbuettel:
> IMHO rather than having 3 separate headers we should just have a single header

That's exactly how I started (see the [feature/new_headers_initial](https://github.com/RcppCore/Rcpp/tree/feature/new_headers_initial) branch and its initial (of four) commits) and I ended up disliking the name I picked ("RcppFast", and "RcppLite" is also not great).  I now truly prefer the finer-grained approach.

As for effect, they all matter a little as I measure, but _none_ comes close to the effect of "throwing it all out" (I have a package for that) and even that is *strongly dominated* by just using `ccache` in front.  And net-net I think more headers == more choice == more clearly stating there are different `#define` to set helps.

eddelbuettel:
> I don't see how definining new headers inside the Rcpp folder, next to many others, improves discoverability

The key is a new entry point `Rcpp/Rcpp` so that 
- (maybe "eventually") we could even remove the old `Rcpp.h` a level above
- I really want to remove the `RcppCommon.h` a level above
- with these new entry points I want to clean up `Rcpp/*` which is sprawling mess

But we need something to support it during the (slow, long) passage to a better layout.  This could be a first step.

eddelbuettel:
Assuming / hoping this is not to offensive to @kevinushey I plan to merge this today given the nod from @Enchufa2.

kevinushey:
LGTM! I still prefer a single separate "light" header but don't feel that strongly either way.

eddelbuettel:
Sounds good -- We can treat it as incremental work in progress.

eddelbuettel:
Any use in keeping the predecessor 'feature/new_headers_initial' branch around?  Probably not but I figure I might ask before blowing it out.

Enchufa2:
No need here.

eddelbuettel:
I also prefer clean and sparse so it'll likely be gone soon.

PR #1196, opened 2022-02-12
Fix pdf textbox overflow (fixes #1195)

### LINKED ISSUE #1195: Minor documentation issue with .pdf textbox overflow
While trying to understand why Rcpp was documenting a NULL object (nlmixr2/rxode2#85), I found a minor formatting issue with the .pdf documentation on page 5 of the left column in https://cran.r-project.org/web/packages/Rcpp/vignettes/Rcpp-attributes.pdf.  Specifically, if you search for "NULL", you will see a textbox where the first column overlaps the second.

I'm going to make a simple PR for it, but I can't easily test the PR because of some issue with my TeX installation relative to what Rcpp requires.

billdenney:
This fixes #1195 (I think).  I can't readily test it because of some issue with my TeX installation, but it is so simple that I think that it'll work (or if not, it'll point in the right direction for a simple fix).

### DIFF
--- a/vignettes/rmd/Rcpp-attributes.Rmd
+++ b/vignettes/rmd/Rcpp-attributes.Rmd
@@ -684,7 +684,8 @@ In this case, a call to `my_package_init()` will be added to the end of the auto
 ```{Rcpp, eval = FALSE}
 void my_package_init(DllInfo *dll);
 RcppExport void R_init_pkgname(DllInfo *dll) {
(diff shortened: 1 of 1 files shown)


eddelbuettel (approved):
Yup, that should fix it.

eddelbuettel:
Will merge, the one X is from the coverage run and due to fragility in package installation.

PR #1202, opened 2022-03-09
Remove deprecated std::(unary|binary)_function (closes #1201)

### LINKED ISSUE #1201: Possible Rcpp hotfix release to appease compiler nags w.r.t. deprecated pre-C++11 idioms
Kurt Hornik emailed me because of an (unrelated) RcppArmadillo issue (involving a possible need for `-latomic` on Debian, that is still TBD) and noted that newer compilers nag a bit with respect to `std::unary_function()` and `std::binary_function()`.   See for example in the [results for RcppRedis](https://cloud.r-project.org/web/checks/check_results_RcppRedis.html) which I looked at yesterday when it updated:

```
Result: WARN
    Found the following significant warnings:
     /home/hornik/tmp/R.check/r-devel-gcc/Work/build/Packages/Rcpp/include/Rcpp/Language.h:180:36: warning: ‘template<class _Arg, class _Result> struct std::unary_function’ is deprecated [-Wdeprecated-declarations]
     /home/hornik/tmp/R.check/r-devel-gcc/Work/build/Packages/Rcpp/include/Rcpp/Language.h:197:37: warning: ‘template<class _Arg1, class _Arg2, class _Result> struct std::binary_function’ is deprecated [-Wdeprecated-declarations]
     /home/hornik/tmp/R.check/r-devel-gcc/Work/build/Packages/Rcpp/include/Rcpp/StringTransformer.h:30:47: warning: ‘template<class _Arg, class _Result> struct std::unary_function’ is deprecated [-Wdeprecated-declarations]
```

When he emailed me earlier yesterday he only mentioned the thrid of these for `StringTransformer` and I quickly played with fix -- this is in fact simple enough thanks to `std::function` -- using `std::function<bool(int)>` instead of the deprecated `std::unary_function<int, bool>` is all it took in another example.  I think replacing `std::binary_function` will be equally simple.

The question now is how to best implement the check as we need to keep the old code for compatibility's sake.  Shall we just use

```c++
#if __cplusplus >= 201103L
```

to flip all C++11 or later compilation to use `std::function()`?  Or are there any downsides I am not seeing?  

(We will of course test fully at CRAN).

While we are at this we should probably also cherry-pick in #1193 which fixed a minor hoopla w.r.t. to C++98.   

So these two changes, and we call it Rcpp 1.0.8-1 ?  (And I will make sure to turn the unit test for three vs four version numbers off.)
Or do we have preference for four dots and call it 1.0.8.1 ?

eddelbuettel:
As discussed in #1201, CRAN would like us to update Rcpp to no longer trigger 'deprecated' warnings under (fairly brandnew) compilers.  This PR replaces the use of `std::unary_function()` in two files and `std::binary_function()` in one.  

The reverse-depends check is ongoing and as not revealed any issues yet.

- [x] Code compiles correctly


kevinushey (approved):
LGTM, but I do wonder if we can just remove the inheritance from unary_function / binary_function altogether, since all those classes are is a bunch of typedefs. E.g. on macOS with libc++:

```
template <class _Arg, class _Result>
struct _LIBCPP_TEMPLATE_VIS unary_function
{
    typedef _Arg    argument_type;
    typedef _Result result_type;
};
```

```
template <class _Arg1, class _Arg2, class _Result>
struct _LIBCPP_TEMPLATE_VIS binary_function
{
    typedef _Arg1   first_argument_type;
    typedef _Arg2   second_argument_type;
    typedef _Result result_type;
};
```

And since we don't use those typedefs (or pass these along to functions expecting these typedefs) I think it's safe to just remove them.

Enchufa2 (changes requested):
There's another `unary_function` in `inst/tinytest/cpp/sugar.cpp`, and then we should remove the unary-thing also from the vignettes (the sugar and FAQ vignettes contain examples with `unary_function`).

eddelbuettel:
Thanks for the heads-up. I had seen the one in the test file, but then forgot about it. Made the same conditional change.  And replaced unconditional in the two markdown files.  (One had not been touched in a while and Emacs decided to fiddle with a lot of whitespace so the diff looks bigger there.  It isn't.  GH has a toggle too for diff disply without whitespace.)

rev.dep is about half-way done and looks fine so far.  I had ran mostly RcppArmadillo of late so a few build-depends need refreshing and I have a few failed-only-because-xyz-missing.

Enchufa2 (approved):
> GH has a toggle too for diff disply without whitespace.

Thanks, that's useful, didn't know about it.

LGTM!

PR #1209, opened 2022-03-30
Use fast_eval when constructing Date (resolves #1208)

### LINKED ISSUE #1208: Converting strings to dates is slow due to tryCatch
The Date constructor for strings uses R functions, which causes Rcpp to insert a very costly tryCatch around any Rcpp code which uses it. Using fast_eval prevents this issue. Handling of invalid input is unchanged.

I've implemented this solution here: #1209

As a benchmark, the following script's runtime drops from 40 seconds to 16 seconds on my machine:

```R
library(Rcpp)

cppFunction('
  double dates_offset(std::string& base, Rcpp::StringVector& date_strs) {
    double offset = 0;
    Date date = Date(base, "%Y-%m-%d");
    for (auto it = date_strs.begin(); it != date_strs.end(); ++it) {
      Date d = Date(Rcpp::as<std::string>(*it), "%Y-%m-%d");
      offset += d.getDate() - date.getDate();
    }
    return offset;
  }
')

base <- "2000-01-01"
dates <- seq(as.Date('1000-01-01'), as.Date('2999-12-31'), by="day")
dates_offset(base, dates)
```

And the proof is in the profvis:
image

Compiling with g++ (GCC) 11.2.0. My sessionInfo:
```
R version 4.1.2 (2021-11-01)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Arch Linux

Matrix products: default
BLAS:   /usr/lib/libblas.so.3.10.0
LAPACK: /usr/lib/liblapack.so.3.10.0

locale:
 [1] LC_CTYPE=en_NZ.UTF-8       LC_NUMERIC=C
 [3] LC_TIME=en_NZ.UTF-8        LC_COLLATE=en_NZ.UTF-8
 [5] LC_MONETARY=en_NZ.UTF-8    LC_MESSAGES=en_NZ.UTF-8
 [7] LC_PAPER=en_NZ.UTF-8       LC_NAME=C
[11] LC_MEASUREMENT=en_NZ.UTF-8 LC_IDENTIFICATION=C

attached base packages:

[1] compiler_4.1.2
```

louisswarren:
The Date constructor for strings uses R functions, which causes Rcpp to
insert a very costly tryCatch around any Rcpp code which uses it.
Using fast_eval prevents this issue. Handling of invalid input is
unchanged.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/api/meat/Date.h
+++ b/inst/include/Rcpp/api/meat/Date.h
@@ -30,7 +30,9 @@ namespace Rcpp{
     inline Date::Date(const std::string &s, const std::string &fmt) {
         Function strptime("strptime");	// we cheat and call strptime() from R
         Function asDate("as.Date");	// and we need to convert to Date
-        m_d = Rcpp::as<int>(asDate(strptime(s, fmt, "UTC")));
+        Language strptime_call(strptime, s, fmt, "UTC");
+        Language asDate_call(asDate, strptime_call.fast_eval());
+        m_d = Rcpp::as<int>(asDate_call.fast_eval());
         update_tm();
     }
 


eddelbuettel:
When we suggest 'open an issue before sending a PR' we mean to 'open an issue _in order to establish consensus on what the PR should do_' which is AFAICT not exactly what happenened here.  See #1208, we may not be enthusiastic about this.  Truly sorry if you feel you wasted time, the hint above aims to prevent that.  If you have suggestions for rewriting it more clearly (or forcefully) let us know.  We _do_ have a history of happily taking and acknowledging PRs so we _do_ want you to work with us :)

eddelbuettel:
Per the discussion in #1208, this problem is best addressed elsewhere.  The performance of repeated `Date` ctor calls is not something I am all that proud of, but there are probably preferable ways to fix it.  Thanks for taking the time to look into matters in some detail.

Issue #1210, opened 2022-03-31
Rcpp directly includes the C library header <inttypes.h> instead of C++ <cinttypes>, resulting in macros not working when built with older c standard library headers


pettyalex:
I will work to provide a minimal reproduction example if it helps, but I am encountering a problem building https://github.com/saigegit/SAIGE that seems like a bug in Rcpp. This is manifesting as a build failure where the macros defined in `<inttypes.h>` are not defined: https://github.com/statgen/savvy/issues/23 . 

I'll make an accompanying PR to Rcpp with a fix for this.

In the C99 spec section 7.8.1, footnote 191 says that the macros defined in <inttypes.h> should only be defined when __STDC_FORMAT_MACROS is defined:

```
C++ implementations should define these macros only when _ _STDC_FORMAT_MACROS is defined
before <inttypes.h> is included.
```

`<cinttypes>`, which is the C++ version of this header has specific code to always set that macro before #including <inttypes.h>: https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.6/a00802_source.html

```
00040 // For 8.11.1/1 (see C99, Note 184)
00041 #if _GLIBCXX_HAVE_INTTYPES_H
00042 # ifndef __STDC_FORMAT_MACROS
00043 #  define _UNDEF__STDC_FORMAT_MACROS
00044 #  define __STDC_FORMAT_MACROS
00045 # endif
00046 # include <inttypes.h>
00047 # ifdef _UNDEF__STDC_FORMAT_MACROS
00048 #  undef __STDC_FORMAT_MACROS
00049 #  undef _UNDEF__STDC_FORMAT_MACROS
00050 # endif
00051 #endif
```

Because Rcpp includes `<inttypes.h>` directly without defining __STDC_FORMAT_MACROS, it will prevent any later includes that include `<cinttypes>` or `<inttypes.h>` from having the macros defined. This means that in the application, if anything that #includes Rcpp is included before another header that needs `<cinttypes>`, then the macros will not be configured, causing the build failure that I saw. I'm able to work around this by changing the order of includes, or by manually including `<cinttypes>` before including anything that depends on Rcpp, but I am fairly certain that Rcpp is a C++ -only library not designed to be consumed from C. If so, then it can use the C++ `<cinttypes>` instead of C `<inttypes.h>`. 

[...]

PR #1211, opened 2022-03-31
fix: Macro problems in consuming libraries because C <inttypes.h> was used

pettyalex:
Change to C++ standard <cinttypes> instead of C library <inttypes.h> to fix macros in consuming applications.

An issue with longer discussion is here: https://github.com/RcppCore/Rcpp/issues/1210

There is a bug in Rcpp when consuming applications are wanting to use the macros defined in <inttypes.h> or `<cinttypes>` when on a platform with C99 <inttypes.h>, which requires __STDC_FORMAT_MACROS to be defined in order to set up its formatting macros. More detail is provided in the issue. 

I will update the checklist as I do the steps

#### Checklist

- [x] Code compiles correctly
- [ ] `R CMD check` still passes all tests

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2022-03-31 Alex Petty <pettyalex@gmail.com>
+
+        * inst/include/Rcpp/hash/hash.h: Replace <inttypes.h> with <cinttypes>
+        fixing a bug when consuming applications use <cinttypes>
+
 2022-03-14  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
--- a/inst/include/Rcpp/hash/hash.h
+++ b/inst/include/Rcpp/hash/hash.h
@@ -22,7 +22,7 @@
 #ifndef RCPP__HASH__HASH_H
 #define RCPP__HASH__HASH_H
 
-#include <inttypes.h>			// needed with g++-4.7 to declare intptr_t
+#include <cinttypes>			// needed with g++-4.7 to declare intptr_t
 
 #include <Rcpp/hash/IndexHash.h>
 #include <Rcpp/hash/SelfHash.h>


eddelbuettel:
Please my comment from _yesterday_:  https://github.com/RcppCore/Rcpp/pull/1209#issuecomment-1083041038

pettyalex:
Something is clearly not set up right with my local development environment, I'm seeing these test failures in a brand new, completely fresh Ubuntu 20.04 with system R 3.6.3, both in master and in my branch. In both cases, the same tests fail with or without this change.

```
  ----- FAILED[data]: test_string.R<76--76>
   call| expect_equal(Encoding(test_String_set_encoding(a)), "UTF-8")
   diff| Expected 'UTF-8', got 'unknown'
  ----- FAILED[data]: test_string.R<77--77>
   call| expect_equal(Encoding(test_String_ctor_encoding(a)), "UTF-8")
   diff| Expected 'UTF-8', got 'unknown'
  Error: 2 out of 1586 tests failed
```

And on my Apple Silicon Mac with R 4.1.2 on clang in a far-from clean environment:
```
  ----- FAILED[data]: test_sugar.R<64--64>
   call| expect_true(is.na(fx(c(NA, 6))))
  ----- FAILED[data]: test_sugar.R<72--72>
   call| expect_true(is.na(fx(c(5, NA))))
  ----- FAILED[data]: test_sugar.R<81--81>
   call| expect_true(is.na(fx(c(NA, 1))))
  ----- FAILED[data]: test_sugar.R<1566--1566>
   call| expect_true(is.na(intmin(integer(0))), "min(integer(0))")
   info| min(integer(0))
```

eddelbuettel:
Anything unusual with locales, maybe? That might explain the string tests issue.

The Arm M1 issue is probably real, there was a good long and very detailed post on R developer blog on issues with M1 and numerics -- I think it was this one: https://developer.r-project.org/Blog/public/2020/11/02/will-r-work-on-apple-silicon/index.html

Rcpp is in good standing at CRAN, and tests fine for me here (and also in Docker to check a 2nd environment, and with `RunAllRcppTests=yes` set).

eddelbuettel:
Given the early demise of #1210 by consensus of all involved I am closing this too.  Thanks for looking into it, there is always work to do and we always value a keen look by motivated eyes.  So thanks for trying to make Rcpp better, evenif this (in hindsight) was rushed.

PR #1219, opened 2022-05-23
Avoid copies in String

### LINKED ISSUE #1218: [FR] Move semantics for Rcpp::String
`Rcpp::String`, unlike Rcpp classes that just wrap SEXPs, copies potentially large data on assignment and construction.  This can often be hidden because of the many implicit type conversions Rcpp exposes.

Some (somewhat contrived) examples:

```c++
void Test1() {
  std::string s = SomeVeryLargeString();
  // Initializes from a std::string, making a std::string copy.
  Rcpp::String rs = s;
}

void Test2(Rcpp::String& out) {
  Rcpp::String rs = SomeVeryLargeRcppString();
  if (ShouldSet()) {
    // Calls Rf_mkCharLenCE, making another copy.
    out = rs;
  }
}

void Test3(std::vector<Rcpp::String>& v) {
  Rcpp::String rs = SomeVeryLargeRcppString();
  // Depending on when SomeVeryLargeRcppString last caused String::setData
  // to get called, could either be a (shallow) SEXP copy or a Rf_mkCharLenCE copy.
  v.push_back(rs);
}

void Foo(const Rcpp::String& rs) {
  // Need to make copy of rs to mutate.
}

void Test4() {
  Foo(rs);
}
```

Better C++11 would be:

```c++
void Test1() {
  std::string s = SomeVeryLargeString();
  // Move the std::string to rs.buffer.
  Rcpp::String rs = std::move(s);
}

void Test2(Rcpp::String& out) {
  Rcpp::String rs = SomeVeryLargeRcppString();
  if (ShouldSet()) {
    // Move rs.buffer or rs.data without making a copy.
    out = std::move(rs);
  }
}

void Test3(std::vector<Rcpp::String>& v) {
  Rcpp::String rs = SomeVeryLargeRcppString();
  // Move rs.buffer or rs.data without making a copy.
  v.push_back(std::move(rs));
}

// Pass by value!
void Foo(Rcpp::String rs) {
  // rs is mutable!
}

void Test4() {
  // No deep-copy.
  Foo(std::move(rs));
}
```

For the C++11 example to actually avoid copies, it will need appropriate move/copy/assignment overloads to be added to `Rcpp::String` for both `Rcpp::String&&` and `std::string&&`.

Admittedly I don't have any real-world benchmarking to quantify the improvement.  Even without benchmarking, I think this is a small and safe change (with appropriate C++11 guards).  Happy to send a PR if it sounds acceptable.

p00ya:
Add move constructor/assignment for String&& and std::string&&, and
attempt to preserve the buffer/SEXP representation when copying in the
existing copy ctor and assignment operators.

Closes #1218.

#### Checklist

- [X] Code compiles correctly
- [X] Prefereably, new tests were added which fail without the change

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,11 @@
+2022-05-25  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/tinytest/test_string.R: Add C++11 conditioning for new unit tests
+
 2022-05-23  Dean Scarff  <dos@scarff.id.au>
 
-	* inst/include/Rcpp/String.h: Make less copies of strings by adding
-	move semantics and preserving the buffer/SEXP representation when
-	copying
+	* inst/include/Rcpp/String.h: Make less copies of strings via move
+	semantics and preserving the buffer/SEXP representation when copying
 	* inst/include/RcppCommon.h: include <utility> for std::move
 	* inst/tinytest/cpp/String.cpp: Add unit tests
 	* inst/tinytest/test_string.R: Add unit tests
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2018-10-25  Dean Scarff  <dos@scarff.id.au>
+2022-05-23  Dean Scarff  <dos@scarff.id.au>
 
 	* inst/include/Rcpp/String.h: Make less copies of strings by adding
 	move semantics and preserving the buffer/SEXP representation when
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2018-10-25  Dean Scarff  <dos@scarff.id.au>
+
+	* inst/include/Rcpp/String.h: Make less copies of strings by adding
+	move semantics and preserving the buffer/SEXP representation when
+	copying
+	* inst/include/RcppCommon.h: include <utility> for std::move
+	* inst/tinytest/cpp/String.cpp: Add unit tests
+	* inst/tinytest/test_string.R: Add unit tests
+
 2022-05-12  Dirk Eddelbuettel  <edd@debian.org>
 
 	* docker/ci/Dockerfile: Added xml2, also use more r-cran-* binaries
--- a/inst/include/Rcpp/String.h
+++ b/inst/include/Rcpp/String.h
@@ -58,8 +58,11 @@ namespace Rcpp {
         }
 
         /** copy constructor */
-        String(const String& other) : data(other.get_sexp()), token(R_NilValue), valid(true), buffer_ready(false), enc(Rf_getCharCE(other.get_sexp())) {
-            token = Rcpp_PreciousPreserve(data);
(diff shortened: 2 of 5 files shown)


ChangeLog
```diff
@@ -1,4 +1,4 @@
-2018-10-25  Dean Scarff  <dos@scarff.id.au>
+2022-05-23  Dean Scarff  <dos@scarff.id.au>
```
eddelbuettel, line 1: Hah!   :wink:

eddelbuettel:
Reverse-dependency checks are running, slowly, on an older machine.  No compilation or issues so far (besides "mechanical" stuff a few 'new-since-last-check' packages need additional build depends involved, and one graphics-device package needs a rebuild for R 4.2.0) so the PR is looking good from that angle.

Enchufa2:
Looks good on first (very quick) pass. I can go over it more carefully in a few days if needed.

p00ya:
One thing for reviewers to pay attention to: the new unit tests need C++11 and I didn't add guards in there.  I'm a bit confused about what the expected C++98 support is in Rcpp now: eddelbuettel@ said Rcpp has "C++11 as a baseline" in #1218, but Contributing.md says C++11 code needs guards.  The `inst/include` stuff should be fine with C++98 (though admittedly I haven't tested this unless it's covered by `R CMD check`), but the tests would fail.

eddelbuettel:
If no language level is specified R uses its default which for R 4.2.0 (and 4.1.0 onwards actually) is C++14, R 4.0.0 brought us eleven:

> 1.2.4 Using C++ code
> --------------------
>
> R can be built without a C++ compiler although one is available (but not
> necessarily installed) on all known R platforms.  As from R 4.0.0 a C++
> compiler will be selected only if it conforms to the 2011 standard
> (‘C++11’).  A minor update(1) (‘C++14’) was published in December 2014
> and will be used by default as from R 4.1.0 if supported.  Further
> revisions ‘C++17’ (in December 2017), and ‘C++20’ (with many new
> features in December 2020) have been published since.

So we should be good but your point is well taken: these tests could fail for someone running them under R 3.6.3, say.

I am not sure if we need (or want) to worry about that.

eddelbuettel:
That said, we have battles C++(11) availability for so long that we do of course have tools:

```r
> Rcpp:::capabilities()
    variadic templates      initializer lists     exception handling     tr1 unordered maps 
                 FALSE                   TRUE                   TRUE                   TRUE 
    tr1 unordered sets           Rcpp modules             demangling            classic api 
                  TRUE                   TRUE                   TRUE                  FALSE 
             long long   C++0x unordered maps   C++0x unordered sets     Full C++11 support 
                  TRUE                   TRUE                   TRUE                   TRUE 
new date(time) vectors 
                  TRUE 
> Rcpp:::capabilities()[["Full C++11 support"]]
[1] TRUE
> 
```

So maybe it would be best to wrap that test around this, or call `exit_file("Skipping for lack of C++11")` if you see `FALSE`.

eddelbuettel:
No further issues so happily merging what looks like a pretty nice PR -- thanks again!

PR #1223, opened 2022-07-02
Rcpp 1.0.9

eddelbuettel:
Release 1.0.9 

- [x] Code compiles correctly

Submitted to and now pending at CRAN and expect email on the grand-fathered NOTE on `.Call(symbol)` as well as possible question from what is a likely false positive or two in the reverse-depends checks against 2500+ package so planning to leave this open this accepted at CRAN.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,11 @@
 2022-07-02  Dirk Eddelbuettel  <edd@debian.org>
 
+        * DESCRIPTION (Date, Version): Release 1.0.9
+
+        * inst/include/Rcpp/config.h: Idem
+        * inst/NEWS.Rd: Idem
+        * vignettes/rmd/Rcpp.bib: Idem
+        * inst/bib/Rcpp.bib: Idem
 	* vignettes/pdf/*: Rebuilt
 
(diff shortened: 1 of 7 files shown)


eddelbuettel:
Received the 'auto-archived' email caused by the one grandfathered NOTE, will reply and request manual move to reverse-depends testing which, once it runs, will trigger another email.   This being the weekend it will be some time before the package appears but we are in rush, and the branch is good.

eddelbuettel:
"No rest for the wicked" : Uwe has just triggered reverse depends so we should know more tomorrow.

eddelbuettel:
And received (and replied-to) the results email.  Usual mix of inapplicable ("GNU make is a SystemRequirement", or `libgeos rebuit`, or 'package datasets not found') with three actual vignette errors I cannot replicate on the test box.

eddelbuettel:
And after a week of mostly waiting we are now at the awaited

> Thanks, on its way to CRAN.

No sour grapes though. Rcpp affects a lot of other packages and CRAN has every right to be careful.  But so are we.

Onwards...

PR #1225, opened 2022-07-23
Enable unwind protection by default

Enchufa2:
As discussed, this PR enables unwind protection by default:

- `RCPP_USE_UNWIND_PROTECT` has no effect anymore.
- Instead, it can be disabled by defining `RCPP_NO_UNWIND_PROTECT`, similarly to other `RCPP_NO_*` defines.
- The associated plugin is deprecated. To avoid errors, it currently shows a warning and has no effect. It could be removed in a future release.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,8 +3,10 @@
 	* inst/include/Rcpp/r/headers.h: Unwind protection is enabled by default
 	unless `RCPP_NO_UNWIND_PROTECT` is defined (`RCPP_USE_UNWIND_PROTECT` is
 	not checked anymore and has no effect)
+	* inst/include/Rcpp/vector/Vector.h: Switch evaluation to `Rcpp_eval`
 	* R/Attributes.R: `[[Rcpp::plugins(unwindProtect)]]` is deprecated
 	* inst/tinytest/cpp/stack.cpp: Adapted
+	* inst/tinytest/test_exceptions.R: Adapted
 	* inst/tinytest/test_interface.R: Adapted
 	* inst/tinytest/testRcppInterfaceExporter/src/config.h: Removed
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2022-07-23  Iñaki Ucar  <iucar@fedoraproject.org>
+
+	* inst/include/Rcpp/r/headers.h: Unwind protection is enabled by default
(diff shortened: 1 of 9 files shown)


Enchufa2:
Added the following fixes:

- 666fa8c4afe68a719d084fe4794a85f1c20a716c: `ExpressionVector::eval()` doesn't work with fast evaluation, because `Rf_eval` just returns the input expression as is. Therefore, I switched it to using `Rcpp_eval()` instead, to preserve the usual behaviour.
- fdadbb464ceb0b1c3cc4579c90ce2851bfc47037: Exceptions try to parse the output from `sys.calls()` in order to [`get_last_call()`](https://github.com/RcppCore/Rcpp/blob/e1dedab3ee24028970137ee0c171631b50ff1698/inst/include/Rcpp/exceptions.h#L268-L286), taking advantage of the call stack set by `Rcpp_eval()`. With fast evaluation, this trick doesn't work anymore. So I've commented out the checks for the call name, unless we come up with another trick to get the last call under fast evaluation.

About the deprecation of the plugin, `.Deprecated()` is handy when a function is deprecated, because it automatically captures the function name and allows you to redirect the user to another one; otherwise, it's just a wrapper around `warning()`. Given that this is not a function and we would need to override the default behaviour anyway, I just use `warning()` directly.

Also, a quick search shows [only 5 results](https://github.com/search?q=org%3Acran+%5B%5BRcpp%3A%3Aplugins%28unwindProtect%29%5D%5D&type=code) for this plugin, and one is Rcpp itself. So I could send patches to those 4 packages and there should be no trouble removing the plugin in the future.

eddelbuettel:
Catching up after a day in the country-side with a nine-hour power outage :)   The plan with respect to patches to six packages sounds good -- you could follow the pattern in #1158 which set up a useful model.  Not suggesting we _must_ do that but I felt it helped and provided transparency.

eddelbuettel:
Following the usual full reverse-depends checks (discussed with @Enchufa2 in parallel over slack) and the one change needed already merged (thanks to @Enchufa2 for the PR and @nacnudus for the merge) this can be merged.  Yay!  I will follow up with a minimal cleanup commit to roll the version number.

Enchufa2:
Yay! Thanks. I'll contact the maintainers of those packages that use the plugin to switch to the define instead, so that one day it can be safely removed. I'll document the process as suggested.

PR #1228, opened 2022-09-20
Support C++20 and C++23 (in r-devel)

eddelbuettel:
r-devel now support C++23 as a compilation standard (if and when the compiler does) so we added a plugin.  We also added one for C++20 which we had apparently skipped.

Tested both under r-devel (svn r82883) and worked.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -551,7 +551,7 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
         list(env = list(PKG_CXXFLAGS ="-std=c++20"))
 }
 
-# built-in C++23 plugin for C++20
+# built-in C++23 plugin for C++23
 .plugins[["cpp23"]] <- function() {
     if (getRversion() >= "4.3")         # with recent R versions, R can decide
         list(env = list(USE_CXX23 = "yes"))
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -573,6 +573,11 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
     list(env = list(PKG_CXXFLAGS ="-std=c++2a"))
 }
 
+## built-in C++2b plugin for compilers without C++23 support
+.plugins[["cpp2b"]] <- function() {
+    list(env = list(PKG_CXXFLAGS ="-std=c++2b"))
+}
+
 ## built-in OpenMP plugin
 .plugins[["openmp"]] <- function() {
     list(env = list(PKG_CXXFLAGS="-fopenmp",
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -556,7 +556,7 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
     if (getRversion() >= "4.3")         # with recent R versions, R can decide
         list(env = list(USE_CXX23 = "yes"))
     else
-        list(env = list(PKG_CXXFLAGS ="-std=c++20"))
+        list(env = list(PKG_CXXFLAGS ="-std=c++23"))
 }
 
 
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -543,6 +543,23 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
         list(env = list(PKG_CXXFLAGS ="-std=c++17"))
 }
 
+# built-in C++20 plugin for C++20
+.plugins[["cpp20"]] <- function() {
+    if (getRversion() >= "4.2")         # with recent R versions, R can decide
+        list(env = list(USE_CXX20 = "yes"))
+    else
+        list(env = list(PKG_CXXFLAGS ="-std=c++20"))
+}
+
+# built-in C++23 plugin for C++20
+.plugins[["cpp23"]] <- function() {
+    if (getRversion() >= "4.3")         # with recent R versions, R can decide
+        list(env = list(USE_CXX23 = "yes"))
+    else
+        list(env = list(PKG_CXXFLAGS ="-std=c++20"))
+}
+
+
 ## built-in C++1z plugin for C++17 standard under development
 ## note that as of Feb 2017 this is taken to be a moving target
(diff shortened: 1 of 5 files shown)


R/Attributes.R
```diff
@@ -543,6 +543,23 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
         list(env = list(PKG_CXXFLAGS ="-std=c++17"))
 }
 
+# built-in C++20 plugin for C++20
+.plugins[["cpp20"]] <- function() {
+    if (getRversion() >= "4.2")         # with recent R versions, R can decide
+        list(env = list(USE_CXX20 = "yes"))
+    else
+        list(env = list(PKG_CXXFLAGS ="-std=c++20"))
```
Enchufa2, line 551: I think it should be `PKG_CXXFLAGS ="-std=c++2a"` for experimental support.

R/Attributes.R
```diff
@@ -543,6 +543,23 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
...
+}
+
+# built-in C++23 plugin for C++20
+.plugins[["cpp23"]] <- function() {
+    if (getRversion() >= "4.3")         # with recent R versions, R can decide
+        list(env = list(USE_CXX23 = "yes"))
+    else
+        list(env = list(PKG_CXXFLAGS ="-std=c++20"))
```
Enchufa2, line 559: I think it should be `PKG_CXXFLAGS ="-std=c++2b"` for experimental support.

R/Attributes.R
```diff
@@ -543,6 +543,23 @@ compileAttributes <- function(pkgdir = ".", verbose = getOption("verbose")) {
...
 
+# built-in C++20 plugin for C++20
+.plugins[["cpp20"]] <- function() {
+    if (getRversion() >= "4.2")         # with recent R versions, R can decide
+        list(env = list(USE_CXX20 = "yes"))
+    else
+        list(env = list(PKG_CXXFLAGS ="-std=c++20"))
+}
+
+# built-in C++23 plugin for C++20
```
Enchufa2, line 554: for C++2**3**?

R/Attributes.R
eddelbuettel: Hm.  Note that this is an `else` branch and by default we don't even get there.  The higher-level abstraction of letting R take care of how it calls it via `USE_CXX20` is good.   

And both tests (for 20 and 23) worked with the `g++-11` I have here which is not even all that new -- I didn't yet dive into a quick Debian container to try `g++-12`.

Look winded way of saying:   do we really need it?   If someone (you?) wants to set `PKG_CXXFLAGS ="-std=c++2a"` they can outside the plugin.  Unless I miss something I think the default I put here is good.

R/Attributes.R
eddelbuettel: See previous comment.

R/Attributes.R
eddelbuettel: Yes, copy and paste-oh. Will fix.

R/Attributes.R
Enchufa2: The thing is that there are a couple of machines with gcc 8 on CRAN, which doesn't support the `-std=c++20` flag, but the experimental `-std=c++2a`. So a package using this plugin will fail in those oldrel machines.

eddelbuettel: Well, to be blunt, strictly not my problem if people opt into a standard their compiler does not support. 

And I do not think because of that we should only enable 'experimental' sub-support.  People use C++17 in production, are moving to C++20 and will before long use C++23 (with g++ 11, 12, ...).  We should support that.

If you look further down in the file we also used to add plugins for these half-baked but not quite standards.  We could do that but I still think such users can still fiddle PKG_CXXFLAGS as they see fit.

R/Attributes.R
Enchufa2: Here this is not the C++23 standard, and the `-std=c++23` flag doesn't exist yet, only the `c++2b` one. But I just checked, and it was added in gcc 11, so maybe it's too soon to add a C++23 plugin? Because there are still several machines with gcc 10 on CRAN. Maybe we could add a specific `cpp2b` plugin, as we did with `cpp2a`, and wait a bit for the official C++23 standard to come out for this one.

eddelbuettel: On my machine, a few hours ago:

```
/usr/local/lib/R-devel/lib/R/bin/R CMD SHLIB --preclean -o 'sourceCpp_3.so' 'foo.cpp' 
ccache g++ -std=gnu++23 -I"/usr/local/lib/R-devel/lib/R/include" -DNDEBUG   -I"/usr/local/lib/R-devel/lib/R/library/Rcpp/include" -I"/tmp" -I/usr/local/include    -fpic  -ggdb -pipe -Wall -pedantic -flto -c foo.
cpp -o foo.o                                        
ccache g++ -std=gnu++23 -shared -ggdb -pipe -Wall -pedantic -flto -fpic -L/usr/local/lib/R-devel/lib/R/lib -L/usr/local/lib -o sourceCpp_3.so foo.o -L/usr/local/lib/R-devel/lib/R/lib -lR
> q()                            
edd@rob:~/git/rcpp(feature/c++23)$
```

I have not attempted to enable actual C++23 features -- that is for the compiler.  What I have attempted to enable in this PR is to use the C++23 compilation support _offered by R devel as of today_.  No more, no less.

R/Attributes.R
Enchufa2: I was simply talking about conforming to current CRAN support because that's what was done in the past with C++11 (as can be seen in the several branches for the `cpp11` plugin). But I'm ok if we don't want to do that anymore.

eddelbuettel: I think in the past the compiler barked and only offered the partial/experimental support via the 'a' or 'b' variant.  That's why I tried to test this, and unless I missed something the test passed.  So ... good enough?

eddelbuettel:
Yup -- added '2b' for symmetry as well.  Helps cover more compilers -- thanks for patiently arguing for that.

Issue #1230, opened 2022-09-23
Modules: ensure 'finalize' method is materialized on object instances


kevinushey:
Related to https://github.com/rspatial/terra/issues/30.

Effectively, it seems like the finalize associated with Rcpp modules can fail to run in certain contexts. I haven't been able to figure out a root cause, but I think we do have a safe bandaid for the issue.

Note that this is not clearly a bug in Rcpp; however, it seems like Rcpp is the best place to provide a workaround for this issue.

PR #1231, opened 2022-09-23
force 'finalize' method to be materialized on Modules

kevinushey:
Closes https://github.com/RcppCore/Rcpp/issues/1230.

#### Checklist

- [x] Code compiles correctly
- [ ] Preferebly, new tests were added which fail without the change

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,6 @@
 2022-09-23  Kevin Ushey  <kevinushey@gmail.com>
 
-    * R/Module.R: Force 'finalize' method to be materialized
+	* R/Module.R: Force 'finalize' method to be materialized
 
 2022-09-20  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2022-09-23  Kevin Ushey  <kevinushey@gmail.com>
+
+    * R/Module.R: Force 'finalize' method to be materialized
+
 2022-09-20  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
--- a/R/Module.R
+++ b/R/Module.R
@@ -131,6 +131,8 @@ new_dummyObject <- function(...)                                # #nocov
 
 # class method for $initialize
 cpp_object_initializer <- function(.self, .refClassDef, ..., .object_pointer){
+    # force finalize method to be materialized
+    invisible(.self$finalize)
     selfEnv <- as.environment(.self)
     ## generate the C++-side object and store its pointer, etc.
     ## access the private fields in the fieldPrototypes env.
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -7,6 +7,9 @@
   \itemize{
     \item Changes in Rcpp API:
     \itemize{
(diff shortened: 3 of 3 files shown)


ChangeLog
```diff
@@ -1,3 +1,7 @@
+2022-09-23  Kevin Ushey  <kevinushey@gmail.com>
+
+    * R/Module.R: Force 'finalize' method to be materialized
```
eddelbuettel, line 3: Old Emacs roots here: 1 tab or eight spaces, please.  

(Not sure editorconfig has a mode for ChangeLog...)

kevinushey: Sorry, will fix that up! RStudio wasn't smart enough to infer the indentation here...

R/Module.R
```diff
@@ -131,6 +131,8 @@ new_dummyObject <- function(...)                                # #nocov
 
 # class method for $initialize
 cpp_object_initializer <- function(.self, .refClassDef, ..., .object_pointer){
+    # force finalize method to be materialized
+    invisible(.self$finalize)
```
eddelbuettel, line 135: Wow. A one-liner to fix a woe.  I presume you played with some Rprintf or REprintf here to see it hits?

ChangeLog
eddelbuettel: If only we knew someone working on the IDE :joy_cat:

eddelbuettel:
I'll turn the rev.dep machine on. It'll take a few moments as usual.

kevinushey:
> Wow. A one-liner to fix a woe. I presume you played with some Rprintf or REprintf here to see it hits?

Right -- just to confirm, this is the script I was running to exercise the issue.

`````

trace(Rcpp:::cpp_object_initializer, quote({
    writeLines("Hello from C++Object initializer!")
}))

library("terra")
library("raster")
#terra version 1.2.13
fnz <- "nz_elev.rda"
if (!file.exists(fnz)) {
 download.file("https://github.com/Nowosad/spDataLarge/raw/master/data/nz_elev.rda", fnz, mode = "wb")
}
load("nz_elev.rda")
nz <- rast(nz_elev)
plot(nz)
`````

The tracer indeed fires, and the above code reliably prints some errors with the CRAN release of Rcpp, but not with the bandaid from this PR.

eddelbuettel:
Yes. I encountered similar mysterious issues with XPtr once and in one use case ensure I always explicitly set one. It's weird because it generally works.  But corner cases exist, so I think this goes the right way.

The rev.dep tests being CI we may not see real side effects. But it being September maybe we can get some user segments (geospatial, rstan, ...) to pretty-please try a rc release from the drat before 1.0.10 gets cases in January (under normal timelines).

eddelbuettel:
So big thank you to you for digging in here!

eddelbuettel:
Ran the (slow, long) battery of reverse depends checks (resuls as usual in the rcpp-logs sibbling repo) and "no change to worse" so merging this now.

eddelbuettel:
I also marked it as 1.0.9.3 and pushed a build into the [rcpp-drat](https://rcppcore.github.io/drat/) so we could reach out to the affected packages and suggest they try this version.

kevinushey:
Thanks! I'll send a ping on Twitter and see if there's anyone interested in testing.

rhijmans:
Halleluja. I can confirm that this fixes the issue in "terra" --- at least for the cases that I could reproduce this with

PR #1233, opened 2022-10-06
Support zero row data frame push back/front (closes #1232)

### LINKED ISSUE #1232: DataFrame.push_back() fails for zero-row columns
Downstream bug report: https://github.com/ropensci/pdftools/issues/108
I think the bug is that DataFrame.push_back() gives an error when trying to add a column to a data frame with 0 rows:

```cpp
DataFrame test (int n) {

  IntegerVector foo(n);
  CharacterVector bar(n);
  CharacterVector baz(n);

  Rcpp::DataFrame df = DataFrame::create(
    _["foo"] = foo,
    _["bar"] = bar
  );
  df.push_back(baz);
  return df;
}

```
```
test(0);
# Error in FUN(X[[i]], ...) : is.data.frame(df) is not TRUE
# In addition: Warning messages:
# 1: In poppler_pdf_data(loadfile(pdf), font_info, opw, upw) :
#   Column sizes are not equal in DataFrame::push_back, object degrading to List
```

eddelbuettel:
#### Rationale

This PR addresses the issue identified in #1232 where one cannot push (back or front) into zero-row data.frame objects.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2022-10-06  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/tinytest/test_dataframe.R (test.DataFrame.PushZeroLength): Add
+	new test
+	* inst/tinytest/cpp/DataFrame.cpp (DataFrame_PushOnEmpty): C++
+	support for new test
+
 2022-10-04  Dirk Eddelbuettel  <edd@debian.org>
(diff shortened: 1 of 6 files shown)


Enchufa2 (approved):
LGTM!

Issue #1234, opened 2022-10-12
sourceCpp() failed to compile if there is some space in R path


kirimaru-jp:
When there is some space in R path, sourceCpp failed to compile:
```
library(Rcpp)
code2 = "#include <Rcpp.h>\n\nusing namespace Rcpp;\n\n// [[Rcpp::export]]\nint fibonacci(const int x) {\n        if (x == 0) return(0);\n        if (x == 1) return(1);\n        return (fibonacci(x - 1)) + fibonacci(x - 2);\n    }"
sourceCpp(code = code2)
```
It showed the error:
> Error in system(cmd, intern = !showOutput) : 'd:/Program' not found

```
> sessionInfo()
R version 4.2.1 (2022-06-23 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19044)

Matrix products: default

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] Rcpp_1.0.9

loaded via a namespace (and not attached):
[1] compiler_4.2.1 tools_4.2.1   

```

A possible solution is to add the command `r <- paste0('"', r, '"')` after the [command](https://github.com/RcppCore/Rcpp/blob/8a44ae16bef552d8853c7c8fc1be2baa86b2e795/R/Attributes.R#L132) `r <- paste(R.home("bin"), "R", sep = .Platform$file.sep)`.

After done that, it compiles normally.
```
> fibonacci(10)
[1] 55
```

eddelbuettel:
Yes that line

https://github.com/RcppCore/Rcpp/blob/8a44ae16bef552d8853c7c8fc1be2baa86b2e795/R/Attributes.R#L132

is a little suspicious.  We could use `file.path()` instead of `paste` and we use `normalizePath` in a number of other places already but not here:

```sh
edd@rob:~/git/rcpp(master)$ ag normalizePath
R/tools.R
43:        path <- normalizePath(path)                  # #nocov start

R/Attributes.R
39:    cacheDir <- normalizePath(cacheDir)
50:        rWorkingDir <- normalizePath(dirname(file))
54:    file <- normalizePath(file, winslash = "/")
238:    cppSourcePath <- normalizePath(cppSourcePath)
239:    buildDirectory <- normalizePath(buildDirectory)
245:    cacheFiles <- normalizePath(cacheFiles)
[...]

PR #1235, opened 2022-10-12
Additional file path normalization to protect a windows path

eddelbuettel:
#### Motivation

Issue #1234 demonstrated that we had a left-over non-normalized path, this PR addresses this.  

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2022-10-12  Dirk Eddelbuettel  <edd@debian.org>
+
+	* R/Attributes.R (sourceCpp): Protect path to binary from spaces
+
 2022-10-06  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,8 @@
 2022-10-12  Dirk Eddelbuettel  <edd@debian.org>
 
+	* DESCRIPTION (Version, Date): Roll minor version
+	* inst/include/Rcpp/config.h (RCPP_DEV_VERSION): Idem
+
 	* R/Attributes.R (sourceCpp): Protect path to binary from spaces
 
 2022-10-06  Dirk Eddelbuettel  <edd@debian.org>
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: Rcpp
 Title: Seamless R and C++ Integration
-Version: 1.0.9.4
-Date: 2022-10-06
+Version: 1.0.9.5
+Date: 2022-10-12
 Author: Dirk Eddelbuettel, Romain Francois, JJ Allaire, Kevin Ushey, Qiang Kou,
  Nathan Russell, Inaki Ucar, Douglas Bates and John Chambers
 Maintainer: Dirk Eddelbuettel <edd@debian.org>
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -129,7 +129,8 @@ sourceCpp <- function(file = "",
         }                                                       # #nocov end
 
         # grab components we need to build command
-        r <- normalizePath(file.path(R.home("bin"), "R"), winslash = "/", mustWork = FALSE)
+        r <- file.path(R.home("bin"), "R")
+        if (.Platform$OS.type == "windows") r <- shQuote(r)
         lib  <- context$dynlibFilename
         deps <- context$cppDependencySourcePaths
         src  <- context$cppSourceFilename
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -129,7 +129,7 @@ sourceCpp <- function(file = "",
         }                                                       # #nocov end
 
         # grab components we need to build command
-        r <- paste(R.home("bin"), "R", sep = .Platform$file.sep)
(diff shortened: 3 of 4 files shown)


eddelbuettel:
While at it, do we need more `mustWork = FALSE`?  I see recurrning warnings as the one in the #1234 discussion but that may not come from us.

eddelbuettel:
Very good. Can you recurse into the discussion in #1234 and see with OP?  What I suggested works for him there ...

eddelbuettel:
"Empirically speaking" this is a rare case anyway as we had a bazillion uses of `sourceCpp()` on Windows since that line was written so it is a little puzzling anyway.

eddelbuettel:
Here is the whole block (as I currently have i.e. with the PR).  There is a `shQuote` already for the path, so what you suggest is good -- we can do it for binary too.  Just puzzled it 'worked for him' over in #1234.

```r
        # grab components we need to build command
        r <- normalizePath(file.path(R.home("bin"), "R"), winslash = "/", mustWork = FALSE)
        lib  <- context$dynlibFilename
        deps <- context$cppDependencySourcePaths
        src  <- context$cppSourceFilename

        # prepare the command (output if we are in showOutput mode)
        args <- c(
            r, "CMD", "SHLIB",
            if (windowsDebugDLL) "-d",
            if (rebuild) "--preclean",
            if (dryRun) "--dry-run",
            "-o", shQuote(lib),
            if (length(deps))
                paste(shQuote(deps), collapse = " "),
            shQuote(src)
        )

        cmd <- paste(args, collapse = " ")
        if (showOutput)
            cat(cmd, "\n")										# #nocov

        # execute the build -- suppressWarnings b/c when showOutput = FALSE
        # we are going to explicitly check for an error and print the output
        result <- suppressWarnings(system(cmd, intern = !showOutput))
```

kevinushey:
I'm surprised too, to be honest. That said, if there really was an issue affecting Windows users en-masse we surely would've heard a lot more about it by now!

eddelbuettel:
Yep!  All I can think of is that OP's use case of `sourceCpp(code="...")` is a little less standard than pointing to a file.

Given the above I have less appetite for switching wholesale to `system2()` and its different argument signature / arrangement.  Maybe as you said wrapping `shQuote()`, and maybe only on Windows is the way to go.  

Lunch time here in flyover country.  Will ponder ...

kirimaru-jp:
> Here is the whole block (as I currently have i.e. with the PR). There is a `shQuote` already for the path, so what you suggest is good -- we can do it for binary too. Just puzzled it 'worked for him' over in #1234.

Really sorry, it was my mistake. I changed some code after that to debug, and forgot to restore the original code.
I can confirm that `r <- normalizePath(file.path(R.home("bin"), "R"), mustWork = FALSE)` did not work either:
```
> sourceCpp(code = code2)
 Error in system(cmd, intern = !showOutput) : 'd:\Program' not found
```
`shQuote()` would be a good choice, as I see on my Win 10 machine:
```
> shQuote(R.home("bin"))
[1] "\"d:/Program Files/R/R-4.2.1/bin/x64\""
```

eddelbuettel:
Ah, no worries, and extra brownie points to @kevinushey for catching this. 

I will amend and (conditionally on being on Windows) wrap `shQuote()` around then.

kirimaru-jp:
> ```r
> r <- file.path(R.home("bin"), "R")
> if (.Platform$OS.type == "windows") r <- shQuote(r)
> ```

It worked perfectly!
```
> sourceCpp(code = code2)
> fibonacci(10)
[1] 55
```

kirimaru-jp:
> I don't think this is the right fix. Here's what I see on my Windows VM:
> 
> ```
> > R.home("bin")
> [1] "C:/PROGRA~2/R/R-42~1.1/bin/x64"

I did not see the short path like yours, here is mine:
```
> R.home("bin")
[1] "d:/Program Files/R/R-4.2.1/bin/x64"
```

I checked the help page of `R.home`, and it writes
> On Windows the values of R.home() and R_HOME are switched to the 8.3 short form of path elements if required and if the Windows service to do that is enabled.

I followed this [link](https://digitalsupport.ge.com/en_US/Article/Is-Windows-8-3-File-Naming-Enabled-How-Do-I-Enable-8-3-File-Naming-If-It-Is-Not), and run this command
```
C:\WINDOWS\system32>fsutil 8dot3name query
The registry state is: 2 (Per volume setting - the default).
```
It means that (at least on my Windows) the 8.3 short form is enabled only "per volume on the system", which I guess it means it is _only_ enabled for the system drive C by default, while I installed R in D.
So possibly, we do not heard much about that because almost all R users install R in C on Windows, so it works perfectly.

eddelbuettel:
It's brilliant because the (underappreciated) 'R on Windows FAQ' has recommended for 20-some years to not do that: install in a path with space.  Yet R always did, up until the 4.2.* series which (IIRC) finally converted that.  I have not worked on Windows in many years so I am a little removed from that.

And agreed on the D:/ vs C:/ difference here -- and that is likely why we generally have not seen this issue bubble up.

eddelbuettel:
We (informally) had a review here as @kevinushey spotted a thinko on my part but if anyone has a moment for a more 'formal' review and tick-off I'd appreciate it -- cleaner than me pushing through with a merge.

Enchufa2 (approved):
LGTM!

PR #1236, opened 2022-10-27
Use snprintf instead of sprintf (even in trivial case)

eddelbuettel:
This commit replace the sole use `sprintf` with `snprintf`.  A full run of reverse depends checks revealed no 'change to worse'.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/internal/r_coerce.h
+++ b/inst/include/Rcpp/internal/r_coerce.h
@@ -246,14 +246,14 @@ inline const char* coerce_to_string<REALSXP>(double x){
 template <>
 inline const char* coerce_to_string<INTSXP >(int from) {
 	static char buffer[NB] ;
-    snprintf( buffer, NB, "%*d", integer_width(from), from );
+    snprintf(buffer, NB, "%*d", integer_width(from), from);
     return buffer ;
 }
 template <>
 inline const char* coerce_to_string<RAWSXP >(Rbyte from){
 	static char buff[3];
-    ::sprintf(buff, "%02x", from);
-    return buff ;
+    snprintf(buff, 3, "%02x", from);
+    return buff;
 }
 template <>
 inline const char* coerce_to_string<LGLSXP >(int from){
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,11 @@
 	* inst/tinytest/cpp/coerce.cpp: add coerce unit tests
 	* inst/tinytest/test_coerce.R: idem
 
+2022-10-25  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/internal/r_coerce.h (coerce_to_string<RAWSXP >):
+	Replace last remaining sprintf with snprintf
+
 2022-10-12  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version


inst/include/Rcpp/internal/r_coerce.h
```diff
@@ -246,13 +246,13 @@ inline const char* coerce_to_string<REALSXP>(double x){
...
 	static char buffer[NB] ;
-    snprintf( buffer, NB, "%*d", integer_width(from), from );
+    snprintf(buffer, NB - 1, "%*d", integer_width(from), from);
     return buffer ;
 }
 template <>
 inline const char* coerce_to_string<RAWSXP >(Rbyte from){
 	static char buff[3];
-    ::sprintf(buff, "%02x", from);
+    snprintf(buff, 2, "%02x", from);
```
kevinushey, line 255: Do we have a unit test for this one? I'm worried we might be truncating the representation here.

eddelbuettel: I would welcome one.  Can you think of one?

I did run it over nearly 2600 CRAN packages though as the usual reverse depends check...

kevinushey: I'll try to put one together later this morning.

kevinushey:
Here's an example showcasing the issue:

`````
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
std::string test() {
    return Rcpp::internal::coerce_to_string<RAWSXP>(0xaa);
}

/*** R
test()
*/
`````

Running with this PR, you'll see:

```
> test()
[1] "a"
```

Given that this change doesn't cause any failures I guess this isn't a commonly-exercised code path.

**Comment by eddelbuettel** [comment]
Whoops.  Scrambled egg on my face.

**Comment by eddelbuettel** [comment]
> `snprintf()` always writes a null terminator as well, so I think you can safely pass the full buffer size when calling it (no -1).

I was a little fuzzy / unaware on this one but _do_ recall seeing _many_ use of -1 in the wild.  Including in the same source file a few lines above:

https://github.com/RcppCore/Rcpp/blob/b3b9bd97b50f26305c65e389091f6e5e6cfc6eba/inst/include/Rcpp/internal/r_coerce.h#L240-L241

but then not here (so I changed it in the PR -- the link shown is from the master branch)

https://github.com/RcppCore/Rcpp/blob/4015d891fccf1b65f684ec3da20f61041dc6e6d7/inst/include/Rcpp/internal/r_coerce.h#L248-L249

What is a good about when 'n == n' is safe and when it is not?

**Comment by eddelbuettel** [comment]
Rebased and updated.  Let me know what you think @kevinushey

**Comment by kevinushey** [comment]
> What is a good rule about when 'n == n' is safe and when it is not?

Good question... it's complicated and comes down to whether the format / print routine writes a null terminator for you or not. Since `snprintf()` always writes a null terminator, it's safe to use `n == n` where the buffer includes space for the null terminator, but for other formatting functions the documentation needs to be checked to be sure.

Another wrinkle for `snprintf()` is that, if you're using it to figure out the buffer size when printing, you need to be careful. From https://en.cppreference.com/w/cpp/io/c/fprintf:

> 4) Writes the results to a character string buffer. At most buf_size - 1 characters are written. The resulting character string will be terminated with a null character, unless buf_size is zero. If buf_size is zero, nothing is written and buffer may be a null pointer, however the return value (number of bytes that would be written not including the null terminator) is still calculated and returned.

So you might write code like:

```
// figure out size of formatted string
// NOTE: does _not_ include null terminator!
int n = snprintf(nullptr, 0, "%s\n", "code");

// create a buffer, adding 1 for the null terminator
std::vector<char> buffer(n + 1);

// format string into buffer, again adding 1 for terminator
snprintf(&buffer[0], n + 1, "%s\n", "code");
```

tl;dr: it's unfortunately complicated and depends on whether the formatter writes a null terminator for you, and in what conditions it writes the null terminator.

kevinushey:
This is also why I am so happy to have [fmt](https://github.com/fmtlib/fmt).

kevinushey (approved):
LGTM!

eddelbuettel:
Thanks for the comment above.  Reading it I realize I once knew only n-1 may be copied to leave room.  So yes, 'it depends'.

Anyway, we made a nice improvement and you rightly took me to school as there shouldn't be a PR without a test under it.  All good.

PR #1240, opened 2023-01-09
Address clang conversion warnings

eddelbuettel:
This PR addresses warnings seen when building with `clang++-14 -Wall -pedantic -Wconversion -Wno-sign-conversion` as done on the CRAN 'M1mac' box.  It is a fairly simple set of casts.  It seems to make sense to fold this into the pending 1.0.10 even though we otherwise dislike making very last minute changes...

#### Checklist

- [x] Code compiles correctly
- [x] `R CMD check` still passes all tests
- [ ] Prefereably, new tests were added which fail without the change
- [x] Document the changes by file in [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog)

PR #1241, opened 2023-01-09
Address a warning from g++-12

eddelbuettel:
This micro PR is a follow-up from #1240 and silences a warning now created by `g++-12`.  The change is included in the reverse-depends check now running.

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
Ah, right, that `size()` method returns `int`. LGTM!

PR #1243, opened 2023-01-24
Quoting only non-standard paths on linux

### LINKED ISSUE #1242: Path quoting in CxxFlags with bash
Hi. In relation to the discussion in #1188 I just wanted to mention an issue introduced there.

## Problem
Adding quotes around path in #1189 broke compilation for anybody who was using `CxxFlags` in simple compilation script like:

```bash
CXXFLAGS="$(R CMD config --cppflags)"
CXXFLAGS="$CXXFLAGS $(R --vanilla --slave -e "Rcpp:::CxxFlags()")"
g++ $CXXFLAGS -c main.cpp
```

This is because bash will use exactly what is returned by `CxxFlags` as argument for the compiler (`g++`). This means that the compiler gets `-I"/some/path"` and searches for headers in `"/some/path"` rather then `/some/path`.

## Other use cases

- You can resolve the quotes in bash by doing `eval g++ ...` in the script, but one cannot modify the all the compiler calls to use "eval" in eg. `./configure` scripts from `autoconf`.
- If one uses the `CxxFlags()` in GNU make, it works. This is because `make` resolves the quotes. For example, the compilation of RInside examples (`example/standard`) works fine.
- On the other hand, it seems to break `cmake`-based compilation. When I try to compile the same examples using `cmake 3.16.3` for configuration, it fails with the error. Apparently passing path with quotes to `include_directories()` through a variable causes some weird behaviour.

## Questions
- If anybody has a good idea how to fix this without hurting windows users and their beloved spaces, I would be happy to implement it and make a pull request. The only half-decent workaround I can think of now is quoting path *if* it has a space.
- Is there a better way to do get include paths from Rcpp, which would work with eg. autoconf?

## Notes
We run into this problem, as we use `RInside` in our `C++` code. We currently use a workaround of stripping quotes.

The `main.cpp` file can be as simple as:
```c++
#include <Rcpp.h>
int main () {
    return 0;
}
```

I'm not attaching `sessionInfo()` as this issue is not about R.

@eddelbuettel Do you think it is an issue at all and would you expect such a bash script to work? If not, I you can close the issue, and we'll stick with our workaround.

llaniewski:
This closes #1242

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-01-24  Lukasz Laniewski-Wollk  <llaniewski@meil.pw.edu.pl>
+
+	* R/RcppLdpath.R: CxxFlags() now quotes only non-standard paths on linux
+
 2023-01-08  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/include/Rcpp/String.h: Address clang++-14 conversion warning
--- a/R/RcppLdpath.R
+++ b/R/RcppLdpath.R
@@ -21,16 +21,11 @@ Rcpp.system.file <- function(...){
 }
 
 ## quote path if non-standard characters are used
-Rcpp.quoteNonStandard <- function(path, quoteAll = .Platform$OS.type!="unix") {
-    quoted <- shQuote(path)
-    if (quoteAll) {
-        quoted
-    } else {
-        # Select paths in which all characters do not need quoting
-        sel <- grepl("^[[:alnum:]/._~+@%-]*$", path)
-        # Quote remaining paths
-        ifelse(sel, path, quoted)
-    }
+Rcpp.quoteNonStandard <- function(path) {
+    ## On unix, check if path has only characters that do not need quoting
+    noquote <- .Platform$OS.type == "unix" && grepl("^[[:alnum:]/._~+@%-]*$", path)
+    ## If no quoting needed return unchanged else quote input
+    if (noquote) path else shQuote(path)
 }
 
 ## Use R's internal knowledge of path settings to find the lib/ directory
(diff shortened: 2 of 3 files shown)


eddelbuettel (approved):
Thanks for adding changelog and news, and flipping from = to <-.  I would probably still reorder the logic in function to (untested)  

```r
Rcpp.quoteNonStandard <- function(path, quoteAll = .Platform$OS.type!="unix") {
    ## Check if path has only characters that not need quoting
    sel <- grepl("^[[:alnum:]/._~+@%-]*$", path)
    ## If no quoting needed return unchanged else quote input
    if (sel) path else shQuote(path)
}
```

Simpler?

llaniewski:
I simplified the function for single path input, and removed the option, as it was not used. Hope this looks good now.

eddelbuettel (approved):
:+1:

Enchufa2:
For the regexp, I suggest following some well-established reference. For example, [this](https://cgit.freedesktop.org/pkg-config/commit/?id=69a7eaa6763bb0920e2b539fffbad51348d94deb) is what pkgconfig escapes in paths:

```
if ((s[0] < '+') ||
    (s[0] > '9' && s[0] < '@') ||
    (s[0] > 'Z' && s[0] < '^') ||
    (s[0] == '`') ||
    (s[0] > 'z')) {
	r[c] = '\\';
	c++;
}
```

So the inverse would be allowed characters.

eddelbuettel:
Good idea in theory.  In practice ... maybe I am lacking coffee (only one cup in) but what exactly are you proposing?  A per-character loop?  A re-written regexp?   Can you sketch something I actually think the regexp is reasonably careful as is.

Enchufa2:
A re-written regexp. Following the ASCII table and the reference above, it seems after all that the only difference would be `^` instead of `%`. But I don't see why `%` would require quotation. :-\

eddelbuettel:
Extra eyes on the ball are a good thing.  Thanks for checking!

PR #1245, opened 2023-01-29
Silence two tests 

eddelbuettel:
This trivial PR makes one test (that otherwise barks on `stdout`) conditional on the env var making some tests quiet.  As we use `system()` the stdout capture of tinytest does not work for us.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2023-01-29  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/tinytest/test_xptr.R: Condition one test writing to stdout
+	on test verbosity environment variable being set
+
 2023-01-24  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,6 @@
-2023-01-29  Dirk Eddelbuettel  <edd@debian.org>
+2023-01-29  Iñaki Ucar  <iucar@fedoraproject.org>
 
-	* inst/tinytest/test_xptr.R: Condition one test writing to stdout
-	on test verbosity environment variable being set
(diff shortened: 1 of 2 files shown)


Enchufa2:
Guilty! That test is not doing what I expected... I can fix it instead of silencing it. May I?

eddelbuettel:
Yes, just use the branch.  I poked around with `expect_*` variants but couldn't make it better.

Enchufa2:
Now it's me who requests your review, but GitHub doesn't allow me to add you, I suppose because you started this PR. :)

inst/tinytest/test_xptr.R
```diff
@@ -60,11 +60,9 @@ void test() {
 '
 
 writeLines(code, file_path)
-expect_silent(system(cmd), info="check that finalizer is NOT called on exit")
+expect_equal(system(cmd, intern=TRUE), character(0))
```
eddelbuettel, line 63: Doh. I need better coffee!

ChangeLog
```diff
@@ -1,7 +1,6 @@
-2023-01-29  Dirk Eddelbuettel  <edd@debian.org>
+2023-01-29  Iñaki Ucar  <iucar@fedoraproject.org>
 
-	* inst/tinytest/test_xptr.R: Condition one test writing to stdout
-	on test verbosity environment variable being set
+	* inst/tinytest/test_xptr.R: Fix a couple of tests writing to stdout
```
eddelbuettel, line 3: 'A couple' ?  'A pair'  :stuck_out_tongue_winking_eye:

eddelbuettel (commented):
LGTM!

inst/tinytest/test_xptr.R
Enchufa2: So did I when I wrote that in the first place! ;-)

ChangeLog
Enchufa2: They're married by their purpose. :)

PR #1248, opened 2023-02-02
Enable compilation under C++20 with clang++ and its standard library 

eddelbuettel:
When using C++20 with `clang++` (version 15) _and_ also adding `-stdlib=libc++` to enable use of its standard library,
the removal of `std::result_of` bites us and compilation fails.  This was deprecated in C++17, and removed in C++20.  Conditional use of the replacement `std::invoke_result` works.

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
LGTM!

PR #1252, opened 2023-02-07
Use more accurate qnorm value from R 4.3.0 (Closes #1251)

### LINKED ISSUE #1251: Update one test for increased qnorm precision in R 4.3.0
Reading the r-devel changes RSS feed has its benefits as I noticed the other day (Feb 4 to precise) that 

> CHANGES IN R-devel NEW FEATURES
> ‘qnorm(<very large negative>, log.p=TRUE)’ is now fully accurate (instead of to “only” minimally five digits).

And indeed, we now get 

```
  ----- FAILED[data]: test_stats.R<246--246>
   call| expect_equal(runit_qnorm_log(-1e+05)$lower, -447.1974945)
   diff| Expected '-447.1974945', got '-447.197893678525'
  Error: 1 out of 1590 tests failed
```

Fix is easy and I'll push it once the CITATION issue in #1250 is dealt with.

eddelbuettel:
R-devel, per the Feb 4 NEWS post, now have a more accurate `qnorm`.  We need to reflect that in a test.

There may a 'timing' issue with the docker container used for r-devel, but if this fails now I will update the container.  The new tests passes locally.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-02-07  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/tinytest/test_stats.R: Use more accurate value in R 4.3.0
+
 2023-02-06  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/CITATION: Update to new format preferred by r-devel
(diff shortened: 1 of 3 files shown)


Enchufa2 (approved):
LGTM. Alternatively, we could set e.g. a `tolerance=0.0005` and call it a day. But this is fine too.

Issue #1255, opened 2023-03-19
New R-devel nags on 'R_forceSymbols' and `compileAttributes()`


eddelbuettel:
@edzer had CCed me on a reply to CRAN after he received a gentle nudge from the CRAN team:

```
Dear maintainer,

Please see the problems shown on
<https://cran.r-project.org/web/checks/check_results_sf.html>.

The "DLL requires the use of native symbols" errors in the
r-devel-linux-x86_64-debian-gcc check results are from

r83985 | luke | 2023-03-16 02:31:57 +0100 (Thu, 16 Mar 2023) | 3 lines
Enforce R_forceSymbols for namespaces, but for now only if an
environment variable is set.

which makes R_forceSymbols work correctly, for now provided that env var
R_REALLY_FORCE_SYMBOLS is set.

You can conveniently extract the non-symbols in your package's foreign
function calls via tools:::package_native_routine_registration_db(),
e.g.,

R> tools:::package_native_routine_registration_db("/data/rsync/PKGS/tibble")
  cname                  s n
1 .Call tibble_need_coerce 1

but please note that there may be several such calls in your code.

Please correct before 2023-04-02 to safely retain your package on CRAN.
```

I took me a moment of head-scratching among several of my packages. I have some which produces a single when `tools:::package_native_routine_registration_db()` was called.  With a bit more digging, and a quick helpful email exchange with @edzer I noticed that our `compileAttributes()` write one remaining `.Call()` with straight apostrophes rather than backticks.  In the package in which I traced that it is (currently) this line:

https://github.com/eddelbuettel/rcppspdlog/blob/0c169b62ddef7b62a2217505bf06760b2b8ffea9/R/RcppExports.R#L194

I have a simple two-char patch to `src/attributes.cpp` which correct this to 

```diff
diff --git a/R/RcppExports.R b/R/RcppExports.R
index 4c7f878..5ec15ee 100644
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -191,5 +191,5 @@ format_stopwatch <- function(sw) {
 
 # Register entry points for exported C++ functions
 methods::setLoadAction(function(ns) {
-    .Call('_RcppSpdlog_RcppExport_registerCCallable', PACKAGE = 'RcppSpdlog')
+    .Call(`_RcppSpdlog_RcppExport_registerCCallable`, PACKAGE = 'RcppSpdlog')
 })
```

[...]

PR #1256, opened 2023-03-19
Update code generation with one more backtick (instead of apostrophe) pair

eddelbuettel:
As discussed in #1255, we have (at least) one remaining case of using apostrophes instead of backticks in the generated file `R/RcppExports.R` created by `compileAttributes()`.  

The narrow fix is simple, the wider issue may be to determine if there are other parts that need updating.  We can keep the PR open for a bit to discuss.

#### Checklist

- [X] Code compiles correctly

### DIFF
--- a/src/attributes.cpp
+++ b/src/attributes.cpp
@@ -2541,9 +2541,11 @@ namespace attributes {
             ostr() << "methods::setLoadAction(function(ns) {" << std::endl;
             ostr() << "    .Call("
                    << (registration_ ? "`" : "'")
-                   <<  registerCCallableExportedName()
-                   << (registration_ ? "`" : "'")
-                   << ", PACKAGE = '" << package() << "')"
+                   << registerCCallableExportedName()
+                   << (registration_ ? "`" : "'");
+            if (!registration_)
+                ostr() << ", PACKAGE = '" << package() << "'";
+            ostr() << ")"
                    << std::endl << "})" << std::endl;		// #nocov end
         }
     }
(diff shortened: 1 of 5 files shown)


src/attributes.cpp
```diff
@@ -2539,8 +2539,8 @@ namespace attributes {
             ostr() << "# Register entry points for exported C++ functions"
                    << std::endl;
             ostr() << "methods::setLoadAction(function(ns) {" << std::endl;
-            ostr() << "    .Call('" << registerCCallableExportedName()
-                   << "', PACKAGE = '" << package() << "')"
+            ostr() << "    .Call(`" << registerCCallableExportedName()
+                   << "`, PACKAGE = '" << package() << "')"
```
eddelbuettel, line 2543: This may need a refinement to condition on whether registration is use or not just as the preceeding function does.

eddelbuettel: And done in https://github.com/RcppCore/Rcpp/pull/1256/commits/f688928567568e73e1d84dc6f5c9956570d7b23b

kevinushey (approved):
LGTM!

eddelbuettel:
One more change, possibly.  I just more-or-less tried to close the circle and help @edzer with this PR https://github.com/r-spatial/sf/pull/2127 and I noticed that there (using the version from this PR) we (once I added `, registration=TRUE` which `sf` was lacking) write this as 

```r
# Register entry points for exported C++ functions
methods::setLoadAction(function(ns) {
    .Call(`_sf_RcppExport_registerCCallable`, PACKAGE = 'sf')
})
```

I probably want to take the `, PACKAGE = 'sf'` out, no?

kevinushey:
Good catch, I think that's correct.

eddelbuettel:
Yeah, I also casually tested it with my package and the PR for `sf`.  I think this now good to go.

PR #1258, opened 2023-03-24
skip tests with path issues on Windows

Enchufa2:
The fix in https://github.com/RcppCore/Rcpp/commit/54ee8fdd91d954ce9458442f3afeeaac3522b337 uncovered an issue with those two tests on Windows. Life is too short to debug path issues for tests on Windows, so this fix just skips them there.

Successful rhub tests here (there's one ERROR, but it's about vignette rebuilding, something about `texi2pdf` not being found): https://builder.r-hub.io/status/Rcpp_1.0.10.3.tar.gz-94d653c1f4064cf29639fc441187ce3d

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-03-24  Iñaki Ucar  <iucar@fedoraproject.org>
+
+	* inst/tinytest/test_xptr.R: Skip some tests on Windows
+
 2023-03-19  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
--- a/inst/tinytest/test_xptr.R
+++ b/inst/tinytest/test_xptr.R
@@ -19,6 +19,9 @@
 
 if (Sys.getenv("RunAllRcppTests") != "yes") exit_file("Set 'RunAllRcppTests' to 'yes' to run.")
 
+## used below
+.onWindows <- .Platform$OS.type == "windows"
+
 Rcpp::sourceCpp("cpp/XPtr.cpp")
 
 #    test.XPtr <- function(){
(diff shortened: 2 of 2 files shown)


kevinushey:
You could probably resolve the Windows issues by using `system2()` instead of `system()`, and using `shQuote()` on the file path passed in.

https://github.com/RcppCore/Rcpp/pull/1258/files?w=1#diff-4e4b42a31588c914e29a3de60cb92b14af71746ae9ab925744c967e8b51141c8R49

I'm also fine with just not worrying about it though, as long as we're certain it's an issue with the test itself and not with Rcpp (which appears to be the case).

eddelbuettel:
Happy to merge as is, question if we should try the suggestion by @kevinushey ?  Or not bother?

Enchufa2:
I think we shouldn't bother. Those tests already were "let's be super cautious" tests. They pass on Unixes, then everything is fine.

eddelbuettel:
We'll know how r-universe fares in a bit.  I'll report back.

eddelbuettel:
Our CI failed on the very regular / unchanged 'update container' step.  Has to do with Debian experimental changing its name to rc-buggy so for now I am just going to remove the apt list file.

eddelbuettel:
And in the meantime r-universe caught up and we are now four out of five:

image

The Apple 4.3 fail is my overeager switch to a higher precision qnorm comparison:

```
test_xptr.R...................    8 tests OK  17.5s
  ----- FAILED[data]: test_stats.R<[246](https://github.com/r-universe/rcppcore/actions/runs/4515895162/jobs/7953727465#step:7:252)--250>
   call| expect_equal(runit_qnorm_log(-1e+05)$lower, -447.197893678525)
   diff| Expected '-447.197893678525', got '-447.197494465048'
  Error: 1 out of 1590 tests failed
  ```

and I'll reverse that out and make sure we only compare to a lower tolerance of, say, 1e-6

```r
> all.equal(-447.197893678525, -447.19749446504, tol=1e-5)
[1] TRUE
> all.equal(-447.197893678525, -447.19749446504, tol=1e-6)
[1] TRUE
> all.equal(-447.197893678525, -447.19749446504, tol=1e-7)
[1] "Mean relative difference: 8.927e-07"
> 
```

PR #1259, opened 2023-03-25
switch to system2 to be able to capture stderr on error (closes #1257)

### LINKED ISSUE #1257: sourceCpp compile error stderr cannot be caught/suppressed
When writing a `try()` function to check for compile errors from `sourceCpp` the compilation error is always printed to the console, regardless of whichever combinations of `try()`, `suppressMessages()`, `suppressWarnings()` or `sink` are used.

Reprex:

``` r
dcode <- "//[[Rcpp::export]]\n int fun(int x) { const int y; return x; }"
status <- suppressWarnings(suppressMessages(
  output <- capture.output(
    try(Rcpp::sourceCpp(code = dcode, verbose=FALSE, echo=FALSE),
                 silent = TRUE))
))
```

<sup>Created on 2023-03-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<summary>
Standard output and standard error
</summary>

``` sh
fileeed07e197428.cpp:2:29: error: default initialization of an object of const type 'const int'
 int fun(int x) { const int y; return x; }
                            ^
                              = 0
1 error generated.
make: *** [fileeed07e197428.o] Error 1
```

Enchufa2:
Minimal changes, fixes #1257. According to [rcheology](https://github.com/hughjonesd/rcheology), `system2` has been around since 2.12.0, and allows us to capture both stdout and stderr when `showOutput=FALSE`. A collateral advantage is that now, when an error happens, the output is in the correct order (previously, the error appeared _first_, then the compilation command).

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-03-25  Iñaki Ucar  <iucar@fedoraproject.org>
+
+	* R/Attributes.R: Switch to system2 to be able to capture stderr on error
+
 2023-03-24  Dirk Eddelbuettel  <edd@debian.org>
 
 	* docker/ci-dev/Dockerfile: During Debian freeze experimental repo is
--- a/R/Attributes.R
+++ b/R/Attributes.R
@@ -137,7 +137,7 @@ sourceCpp <- function(file = "",
 
         # prepare the command (output if we are in showOutput mode)
         args <- c(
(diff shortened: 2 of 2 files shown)


eddelbuettel (approved):
Excellent

kevinushey (approved):

PR #1264, opened 2023-04-12
Simplified container use thanks to hint by @enchufa2

eddelbuettel:
This PR simplifies the Dockerfile for the Linux grid a little as we can indeed remove a layer of docker invocation as all jobs are already in the specified container.  Thanks to @Enchufa2 for the suggestion.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,7 +57,7 @@ jobs:
         run: ${{matrix.r }} CMD build --no-build-vignettes --no-manual .
 
       - name: Check
-        run: ${{ matrix.r }} CMD check --no-vignettes --no-manual Rcpp_*.tar.gz
+        run: CI=true ${{ matrix.r }} CMD check --no-vignettes --no-manual Rcpp_*.tar.gz
 
   #covr:
   #  runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -44,17 +50,14 @@ jobs:
...
-        run: docker run --rm -i -v ${PWD}:/mnt -w /mnt ${{ matrix.cntr }} ${{ matrix.r }} -q -e 'sessionInfo()'
+        run: ${{ matrix.r }} -q -e 'sessionInfo()'
 
       - name: Build
-        run: docker run --rm -i -v ${PWD}:/mnt -w /mnt ${{ matrix.cntr }} ${{matrix.r }} CMD build --no-build-vignettes --no-manual .
+        run: ${{matrix.r }} CMD build --no-build-vignettes --no-manual .
 
       - name: Check
-        run: docker run --rm -i -v ${PWD}:/mnt -w /mnt -e CI=true ${{ matrix.cntr }} ${{ matrix.r }} CMD check --no-vignettes --no-manual Rcpp_*.tar.gz
+        run: ${{ matrix.r }} CMD check --no-vignettes --no-manual Rcpp_*.tar.gz
```
Enchufa2, line 60: Note that `CI=true` was dropped.

eddelbuettel: Fixed -- and that's why we have more than two eyes on these.  Thanks!

Enchufa2 (approved):

kevinushey (approved):
LGTM!

PR #1267, opened 2023-06-12
Update URLs, set Encoding, roll versions

eddelbuettel:
#### Motivation

This PR regroups a set of tiny changes prompted by `R(-devel) CMD check` on the current version.  We apparently never set the now-common Encoding, so now we do. The TAS paper seems to have moved at the publisher site so we use the DOI in README.md, and while we were at this we rolled the micro release to .5.

No new code

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
LGTM!

PR #1272, opened 2023-08-03
Add missing includes

MichaelChirico:
Minor cleanup, no ticket.

Feel free to suggest another ordering of `#include`s, picked somewhat haphazardly

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,5 +7,5 @@ Please explain the changes you want to apply to Rcpp, preferably in an issue tic
 
 - [ ] Code compiles correctly
 - [ ] `R CMD check` still passes all tests
-- [ ] Prefereably, new tests were added which fail without the change
+- [ ] Preferably, new tests were added which fail without the change
 - [ ] Document the changes by file in [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog)
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2023-08-03  Michael Chirico  <chiricom@google.com>
+
+        * src/attributes.cpp: Directly `#include <ostream>` for `std::endl`
+        * src/date.cpp: Directly `#include <R_ext/Booelan.h>` for `TRUE` and `FALSE`
+
 2023-07-03  Dirk Eddelbuettel  <edd@debian.org>
 
         * DESCRIPTION (Date, Version): Release 1.0.11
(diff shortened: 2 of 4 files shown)


eddelbuettel:
I don't tjhink we say anywhere that is relevant, do we?  

This pretty much a textbook case of why I say please file an issue first.

eddelbuettel:
On balance I am in favour, header etiquette and explicit use is better in the long run but once again we do ask that folks play by the book and you would have wasted less of your and my time if you had done so.  Ah well.

MichaelChirico:
I spot-checked the files and indeed there's no direct `#include` for the respective objects (attributes.cpp -> `std::enl`, date.cpp -> `TRUE`,`FALSE`), so it seemed pretty straightforward that we should avoid transitive inclusion.

I also recall "recently" your accepting a similar PR:

https://github.com/eddelbuettel/rcppcctz/pull/42

Sorry about the ChangeLog, I'm guilty of assuming such small changes don't need a log entry. My mistake :)

eddelbuettel:
Your link points at (bleeding edge?) clang-18.  Is that what your toolchain had?  

I seem to have `clang-include-fixer-15` here (Ubuntu 23.04).  Can you (just so we have the notes and could replicate) share the exact invocation of the exact tools you used (unless it violates state secrets of two dozen country and / or your likely paranoid work rules).

MichaelChirico:
> Can you (just so we have the notes and could replicate) share the exact invocation of the exact tools you used

Good and fair question! Unfortunately I have no idea how to map between what I see (a CI warning) and an externally-visible version / command-line invocation that would reproduce it. clang-18 looks correct at a glance. Let me ask around a bit.

eddelbuettel:
Thanks -- using these tools is a net-positive, and it good that Google does (or did, at least) sponsor llvm and other tooling.  But to me really useful to selfish me I need to see how it is running.  You can't expect me to read a man page, can you?

eddelbuettel:
Also formal 'finders fee kudos' for the one-char fix of what must be a years old typo.  I'll overlook the fact that you are mixing issues :grinning: because one char does not warrant another PR.

All good, and thanks!

MichaelChirico:
> one char does not warrant another PR.

I should hope not!

MichaelChirico:
OK here we go:

The `misc-include-cleaner` check was added in June:

https://github.com/llvm/llvm-project/commits/main/clang-tools-extra/clang-tidy/misc/IncludeCleanerCheck.h

The tool I'm seeing is a gussied-up version of this:

https://clang.llvm.org/extra/clang-tidy/#id2

eddelbuettel:
Thanks. I do have `clang-tidy-15` here too but I presume I have to wait for `misc-include-cleaner` then.

Issue #1287, opened 2023-11-26
A fresh -Wformat-security issue under r-devel


eddelbuettel:
**Update 2023-11-28:**  If you came here because of a similar message in your package please read on and see particularly [this comment below](https://github.com/RcppCore/Rcpp/issues/1287#issuecomment-1829886024) for the fairly simple fix.

---

While working on an update for `RQuantLib` with a few r-devel discovered minor changes, CRAN and I both came across a new nag this time from `-Wformat-security.`.  Our glue code in `src/attributes.cpp` does

```c++
                     << "    if (rcpp_isError_gen) {" << std::endl
                     << "        SEXP rcpp_msgSEXP_gen = Rf_asChar(rcpp_result_gen);" << std::endl
                     << "        UNPROTECT(1);" << std::endl
                     << "        Rf_error(CHAR(rcpp_msgSEXP_gen));" << std::endl
                     << "    }" << std::endl
```

and the `Rf_error(someCharvariablehere)` now makes the compiler bark under `-Wformat-security`:

    RcppExports.cpp:180:18: warning: format string is not a string literal (potentially insecure) [-Wformat-security]

The fix is pretty easy:  add a `"%s"`.  I will take care of that shortly.

eddelbuettel:
Yes, I actually got _five_ such emails myself today for packages of mine using `Rcpp`.

The fix is simple thanks to PR #1288 we made two days ago.  Install Rcpp 1.0.11.5 from the Rcpp drat repo via, _e.g._,

     Rscript -e 'install.packages("Rcpp", repos=c("https://RcppCore.github.io/drat", getOption("repos")))'

and then re-run `compileAttributes()`.  That will fix the `RcppExports.cpp` file for you.

TobiasFellinger:
Thanks for the quick fix and the comment on how to fix issues with packages linking to Rcpp.

eddelbuettel:
My pleasure!  As you see in this ticket, I actually hit is myself updating a package a few days ago.  `r-devel` and a new `g++` are good at this. 

And  I am sure we will hear more about it here, at StackOverflow, or on lists such as `rcpp-devel` or `r-package-devel` so when you see it by all means feel free to share the word :grinning:

DISOhda:
> You then need to upload your updated package to CRAN. It has no change in its run-time dependency on `Rcpp` so you do not need to change anything in `DESCRIPTION`. Just re-run `compileAttributes()`.

[...]

PR #1288, opened 2023-11-26
Set a format string in Rf_error call

eddelbuettel:

As discussed in #1287 we trigger a `-Wformat-security` nag from the glue code in `src/attributes.cpp`.  The fix is pretty simple: add a missing `"%s"` format string.
#### Checklist

- [x] Code compiles correctly


kevinushey (approved):
LGTM!

eddelbuettel:
Thanks @kevinushey -- I'll wait for a complete CI run and will then send a commit rolling this to 1.0.11.5.

PR #1293, opened 2023-12-24
Support OBJSXP coming in R 4.4.0 (closes #1283)

### LINKED ISSUE #1283: Switch from S4SXP to OBJSXP
Per the R News RSS feed this morning:

> CHANGES IN R-devel C-LEVEL FACILITIES
>
> ‘SEXP’ type ‘S4SXP’ has been renamed ‘OBJSXP’ to support experimenting with alternative object systems. The ‘S4SXP’ value can still be used in ‘C’ code but is now deprecated. Based on contriubtions from the R Consortium Object-Oriented Programming Working Group.

I only see two uses in `src/api.cpp` so should be quick, and we can probably #ifdef this properly to use S4SXP for R < 4.4.0

eddelbuettel:
This addressed #1283 and the addition of `OBJSXP` as a new value / replacement for `S4SXP` to better support S7 changes. I ran a full reverse-depends check but given that the change is conditional on R 4.4.0 or later nothing much came up.

It is something we probably want to get into 1.0.12 for the next cycle. and I don't think it will do anything wrong but reviews would be much appreciated as I may still have forgotten something.

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-12-27  Dirk Eddelbuettel  <edd@debian.org>
+
+	* src/api.cpp (type2name): Refine OBJSXP return
+
 2023-12-21  Dirk Eddelbuettel  <edd@debian.org>
 
 	* src/api.cpp (type2name): Recognise OBJSXP added in R 4.4.0
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-12-21  Dirk Eddelbuettel  <edd@debian.org>
+
+	* src/api.cpp (type2name): Recognise OBJSXP added in R 4.4.0
+
 2023-11-28  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/tinytest/testRcppInterfaceExporter/R/RcppExports.R: Regenerated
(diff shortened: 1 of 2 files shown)


Enchufa2:
I think we still want to return S4SXP in the appropriate cases. So following [this](https://github.com/wch/r-source/blob/cb08cc58e6d618684b67b03ce73f9d446a60c136/src/main/inspect.c#L69-L73), in case `OBJSXP`,

- `if (IS_S4_OBJECT(x)) return "S4SXP";`
- otherwise, `return "OBJSXP";`

eddelbuettel:
I like that a lot.  So how about

```c
#if R_Version >= R_Version(4,4,0)					// replaces S4SXP in R 4.4.0
        case OBJSXP:    	return Rf_isS4(x) ? "S4SXP" : "OBJSXP"; 	// cf src/main/inspect.c
#else
        case S4SXP:     	return "S4SXP";
#endif
```

moving the switch inside the 'after R 4.4.0' conditional and returning 'S4SXP' for S4 as before and 'OBJSXP' otherwise.

Enchufa2:
Perfect!

eddelbuettel:
Even seems to work:

```r
> library(Rcpp)
> M <- Matrix::rsparsematrix(2,2,0.5)
> cppFunction("const char* t(SEXP x) { return type2name(x); }")
> t(1L)
[1] "INTSXP"
> t(M)
[1] "S4SXP"
> getRversion()
[1] ‘4.4.0’
> 
```

And now I get to learn how to create an S7 object :)

Enchufa2 (approved):
LGTM!

PR #1296, opened 2024-02-21
Condition define of no remap

eddelbuettel:
R 4.4.0 will likely define `R_NO_REMAP` so we need to protect our (currently unconditional) definition.

#### Checklist

- [x] Code compiles correctly


kevinushey (approved):
LGTM!

PR #1306, opened 2024-05-27
Fewer references to StackOverflow

eddelbuettel:
Running `R CMD check` a few times yesterday tickled warning from the URL checker for links to StackOverflow.  Given the overall disenchantment with that site, I opted to just remove actual link references as well as paragraph closer to an endoresement.

Yesterday it also whined about packages.debian.org but that was clearly a temporary hickup with that server and is fine now.

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
👍

Issue #1299, opened 2024-03-12
New warnings from -Wconversion -Wno-sign-conversion


eddelbuettel:
Prof Ripley emailed the following in a messaging concerning the earlier R_NO_MAP issue (that is apparently defered for a bit at CRAN, we did our bit here, he pointed me also to another issue in another package of mine now taken care of):

> A couple of other warnings I noted whilst compiler logs for CRAN 
> packages scrolled by:
> 
> ```
> /Users/ripley/R/Library/Rcpp/include/Rcpp/sugar/functions/sample.h:369:17: 
> warning: implicit conversion loses integer precision: 'R_xlen_t' (aka 
> 'long') to 'int' [-Wshorten-64-to-32]
> 
> /Users/ripley/R/Library/Rcpp/include/Rcpp/internal/export.h:114:9: 
> warning: implicit conversion loses integer precision: 'R_xlen_t' (aka 
> 'long') to 'uword' (aka 'unsigned int') [-Wshorten-64-to-32]
> 
> (That's from clang with -Wconversion -Wno-sign-conversion.)
> ```

eddelbuettel:
I am actually not sure I can reproduce these with `clang++-17`, at least not from building Rcpp.  `uword` hints at RcppArmadillo but it could also be any one of 1100+ packages using it.   Uggh.

Enchufa2:
I assume you asked Prof. Ripley for more details. :) In the meantime, I found [this](https://www.stats.ox.ac.uk/pub/bdr/M1mac/GPvecchia.log) for the second one via Google search, nothing for the first one.

eddelbuettel:
I looked into this some more. One can tickle a fair amount by for example testing our embedded package  `testRcppClass` via the test file `test_module_client_package.R`.  And while I can subsitute a number of `int` for `size_t` invariably I end at an impassed that I convert enough to silence all warnings, the module no longer loads.  Uggh.  One can backtrack and only convert 'some' and the module remains loadable, but so does a faily warning.  The main culprit (for this issue) appears to be [this function](https://github.com/RcppCore/Rcpp/blob/30f1a0a336a12b068ea6f994f076941c069035d0/inst/include/Rcpp/module/class.h#L320-L342).  I changed some more trivial things (adding `void` to signatures in `init.c` for the test package, a number of other `int` -> `size_t` in `class.h`) but this one stings.

So we `-Wconversion -Wno-sign-conversion` under clang are enforced we may have some trouble.

[...]

PR #1307, opened 2024-05-28
Switch to size_t to avoid conversion warnings from clang++-17

eddelbuettel:
Issue #1299 alerts to issues from the (very much opt-in, not (yet ?) compulsory) conversion warnings `clang++17` can do. As I noticed a few from one of our own tests a first attempt squashed those.

This likely does not 'close' the issue so not adding the tag, but should do no harm.

#### Checklist

- [x] Code compiles correctly


kevinushey (approved):
LGTM!

PR #1309, opened 2024-06-05
Additional switch to size_t (or R_xlen_t) to avoid conversion warnings from clang++-17 

eddelbuettel:
This is a continuation of PR #1307 with a few changes -- with many remaining to be done.  As we wrote in #1307:
> Issue https://github.com/RcppCore/Rcpp/issues/1299 alerts to issues from the (very much opt-in, not (yet ?) compulsory) conversion warnings clang++17 can do. As I noticed a few from one of our own tests a first attempt squashed those.
> 
> This likely does not 'close' the issue so not adding the tag, but should do no harm.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-06-01  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/internal/Proxy_Iterator.h (Proxy_Iterator):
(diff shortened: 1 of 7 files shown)


kevinushey (approved):
LGTM!

Enchufa2 (approved):
LGTM!

Issue #1308, opened 2024-06-03
New CRAN issues


eddelbuettel:
Per incoming email

> With current R-devel, we are seeing new segfaults and corresponding
> UBSAN problems for packages
> 
>   VAJointSurv fixest kdevine maxnodf o2geosocial openxlsx reservr vapour
> 
> with fixest and openxlsx showing
> 
>     #0 0x7fd6892ef904 in Rcpp::traits::r_vector_cache<14, Rcpp::PreserveStorage>::ref(long) /data/gannet/ripley/R/test-dev/Rcpp/include/Rcpp/vector/traits.h:46
>     #1 0x7fd6892ef904 in Rcpp::Vector<14, Rcpp::PreserveStorage>::operator[](long) /data/gannet/ripley/R/test-dev/Rcpp/include/Rcpp/vector/Vector.h:340
> 
> and the others similarly.  Can you please have a look?
> 
> [....] tells me that there are also new UBSAN issues for
> 
>   BART FLOPART FLSSS LOPAR aum lme4 magi nanotime ondisc pedmod
>   rsparse supc xrnet
> 
> most of which point to Rcpp (lme4 seems from Eigen).

with short followup

> Apparently this is from
> 
> r86629 | luke | 2024-05-27 00:29:48 +0200 (Mon, 27 May 2024) | 3 lines
> Hide some more functions in inlined.c.
> Also define STRICT_TYPECHECK when compiling inlined.c.

kevinushey:
I was able to reproduce this locally on macOS; I'll see if I can learn more.

eddelbuettel:
Which one did you try?  `openxlsx` may be the smaller of the two.

kevinushey:
This seems like potentially a bug in `openxlsx`? Here's a more minimal example:

```
library(openxlsx)
trace(openxlsx:::read_workbook, quote(print(utils::ls.str())))
xlsxFile <- system.file("extdata", "readTest.xlsx", package = "openxlsx")
read.xlsx(xlsxFile, 2, startRow = 3, colNames = FALSE, detectDates = TRUE)
```

The function `read_workbook` gets called with:

```
Browse[1]> ls.str()
clean_names : function (x, schar)  
cols_in :  int [1:88] 1 2 3 4 5 6 7 8 9 1 ...
hasColNames :  logi FALSE
hasSepNames :  chr "."
is_date :  logi [1:88] FALSE FALSE FALSE FALSE FALSE FALSE ...
[...]

PR #1310, opened 2024-06-11
add template specializations to avoid DATAPTR use

kevinushey:
Addresses https://github.com/RcppCore/Rcpp/issues/1308, but note that this PR would turn these out-of-bounds accesses into R errors, and so we might expect to see some CRAN failures. If that feels too aggressive, we could make these warnings instead, or control the behavior with an environment variable (read on package load), or something else.

This PR seeks to avoid some of the crashes seen in packages using Rcpp, where out-of-bounds accessed could trigger an ASAN misaligned read (due to special behavior in DATAPTR attempting to catch such usages).

This PR also now adds bounds checking to offset accessing via `operator[]` for the various Vector<RTYPE> classes provided by Rcpp; controlled via the same `RCPP_NO_BOUNDS_CHECK` macro used by the subsetting code.

For example:

`````
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
double get(NumericVector x, R_xlen_t index) {
    return x[index];
}

/*** R
get(42, 2)
*/
`````

Would now give:

```
> get(42, 2)
Error in eval(ei, envir) : 
  subscript out of bounds (index 2 >= vector size 1)
```

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/tests/tinytest.R
+++ b/tests/tinytest.R
@@ -43,5 +43,5 @@ if (requireNamespace("tinytest", quietly=TRUE)) {
     ## there are several more granular ways to test files in a tinytest directory,
     ## see its package vignette; tests can also run once the package is installed
     ## using the same command `test_package(pkgName)`, or by director or file
-    tinytest::test_package("Rcpp", ncpu=getOption("Ncpus", 1))
+    tinytest::test_package("Rcpp", ncpu=getOption("Ncpus"))
 }
--- a/tests/tinytest.R
+++ b/tests/tinytest.R
@@ -43,5 +43,5 @@ if (requireNamespace("tinytest", quietly=TRUE)) {
     ## there are several more granular ways to test files in a tinytest directory,
     ## see its package vignette; tests can also run once the package is installed
     ## using the same command `test_package(pkgName)`, or by director or file
-    tinytest::test_package("Rcpp", ncpu=getOption("Ncpus"))
+    tinytest::test_package("Rcpp")
 }
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-06-11  Kevin Ushey  <kevinushey@gmail.com>
+
+	* inst/include/Rcpp/internal/r_vector.h: Use template specializations to
+	avoid DATAPTR usage
+
 2024-06-02  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/include/Rcpp/internal/export.h: More R_xlen_t switching
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,8 @@
 
 	* inst/include/Rcpp/internal/r_vector.h: Use template specializations to
 	avoid DATAPTR usage
+	* inst/include/Rcpp/vector/traits.h: Implement bounds checks in
+	r_vector_cache access
 
 2024-06-02  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,9 @@
 	avoid DATAPTR usage
 	* inst/include/Rcpp/vector/traits.h: Implement bounds checks in
 	r_vector_cache access
+	* inst/tinytest/cpp/Vector.cpp: Add unit tests
+	* inst/tinytest/test_vector.R: Add unit tests
+	* tests/tinytest.R: Test in serial by default
 
 2024-06-02  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -15,7 +15,7 @@
       \item Rcpp now avoids the usage of the (non-API) DATAPTR function when
       accessing the contents of Rcpp Vector objects where possible. (Kevin in
       \ghpr{1310})
-      \item Rcpp now throws an R error on out-of-bounds Vector accesses. (Kevin
+      \item Rcpp now emits an R warning on out-of-bounds Vector accesses. (Kevin
       in \ghpr{1310})
     }
     \item Changes in Rcpp Deployment:
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -9,9 +9,11 @@
     \itemize{
       \item Set R_NO_REMAP if not already defined (Dirk in \ghpr{1296})
       \item Add variadic templates to be used instead of generated code
-      (Andrew Johnson in \ghpr{1303}
+      (Andrew Johnson in \ghpr{1303})
       \item Count variables were switches to \code{size_t} to avoid warnings
       about conversion-narrowing (Dirk in \ghpr{1307})
+      \item Rcpp now avoids the usage of the (non-API) DATAPTR function when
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: tests/tinytest.R [hunk 1]
```diff
@@ -43,5 +43,5 @@ if (requireNamespace("tinytest", quietly=TRUE)) {
     ## there are several more granular ways to test files in a tinytest directory,
     ## see its package vignette; tests can also run once the package is installed
     ## using the same command `test_package(pkgName)`, or by director or file
-    tinytest::test_package("Rcpp", ncpu=getOption("Ncpus", 1))
+    tinytest::test_package("Rcpp", ncpu=getOption("Ncpus"))
```
kevinushey, line 46: I noticed that `tinytest` will use a parallel worker whenever Ncpus is set to a numeric value; leaving it NULL / unset lets us run tests (and see test output directly) in the current R session. Does changing this default seem reasonable?

eddelbuettel: I was a fan of parallel runs (I may even have been the one who suggested it to Mark) but I no longer run them in many packages.  I guess I do / did here because I happen to have `getOption("Ncpus")` set (helps with parallel install.packages / update.packages too) .  But removing the trailing ", 1" is IMHO a bad idea as it removes the default and `getOption()`. If I try NULL explicitly I get rubbish:

```sh
edd@rob:~/git/rcpp(master)$ Rscript -e 'tinytest::test_package("Rcpp", NULL)'
Error in dir.exists(testdir) : invalid filename argument
Calls: <Anonymous> -> dir.exists
Execution halted
edd@rob:~/git/rcpp(master)$ 
```

My personal goto I use all the time is `rcc.r` from littler and it just calls `rcmdcheck::rcmdcheck()`.  So I would be fine with just simplifying this to 

```suggestion
    tinytest::test_package("Rcpp")
```

eddelbuettel:
Nice stuff.  Would you mind if I spend the usual day or two and a half on rev.deps, "just in case" ?

Enchufa2:
Nice implementation, LGTM! Given that these now are going to be ASAN errors on CRAN anyways, I'd say it's fine if they trigger errors from our side.

eddelbuettel:
> Given that these now are going to be ASAN errors on CRAN anyways, I'd say it's fine if they trigger errors from our side.

Very true but it may make updating at CRAN harder for us if we trigger N new things in M packages.

eddelbuettel:
Three hours in (had some follow-up to do on RcppArmadillo run first)
> Test of Rcpp 1.0.12.3.1.1 had 142 successes, 3 failures, and 1 skipped packages.

2 of the 3 are out of bounds (aum, BART) the third is small numerical diff on test so maybe not us (bayesianVars).

bbolker:
Will look forward to seeing whether this resolves https://github.com/lme4/lme4/issues/794 (still hoping that what we're seeing is **not** the revealing of previously cryptic out-of-bounds accesses, but rather things that are legal but trigger an ASAN warning anyway ...)

eddelbuettel:
The reverse depends run finished just as I was leaving for a two-day bike trip.  The results are not terrible but also not something which makes me think we can commit this as is:

```
Test of Rcpp 1.0.12.3.1.1 had 2787 successes, 40 failures, and 25 skipped packages.
Ran from 2024-06-12 08:35:10.73 to 2024-06-14 06:50:59.41 for 1.928 days
Average of 58.397 secs relative to 311.843 secs using 7 runners

Failed packages:  aum, BART, bayesianVARs, birdie, chouca, cna, fixest, FLOPART, FLSSS, HistDAWass, kdevine, LOPART, magi, maxnodf, mined, mixvlmc, nanotime, netdiffuseR, nlmixr2est, o2geosocial, ondisc, onemap\
, openxlsx, pedmod, PKPDsim, Rborist, reservr, revdbayes, rsparse, rstpm2, rxode2, secr, supc, thunder, TreeDist, TSrepr, VAJointSurv, vapour, xrnet, ZIPBayes
```

Almost all the 'fails' are due to run-time and not install-time so it could be the 'harsher' out of bounds access. I will try to summarize later.

eddelbuettel:
@kevinushey With the extra delay of me disappearing for 1 1/2 or 2 days :grinning: I was now able to summarize the reverse dependency run. As [logged](https://github.com/RcppCore/rcpp-logs/blob/master/results/Rcpp-20240612.txt), it is essentially 'out of bounds'. So we probably want to preemptively define the 'death by bounds check' away, or maybe tone it down to a message.

eddelbuettel:
I am _much_ less sanguine here as we are (by the standard release pattern) looking into a release in two week (early July) to four weeks (mid July).  I think we need to address this, or possibly roll it to the next release with sufficient time to communicate to affected maintainers.

Enchufa2:
Then, let's make it a warning for now and an error in the next release?

eddelbuettel:
I ran into errors when I mod'ed it to 'continue' rather than `stop()`.

@kevinushey Any insights from your end?

eddelbuettel:
Thanks for the update, @kevinushey -- I will give this a spin in another reverse-dependency run.

Another thing I noticed was that _if_ one where too define the 'off-switch' the code doesn't actually compile cleanly. I tried this diff

```diff
diff --git a/inst/include/Rcpp/vector/Subsetter.h b/inst/include/Rcpp/vector/Subsetter.h
index 0d8b32a6..0905e1a3 100644
--- a/inst/include/Rcpp/vector/Subsetter.h
+++ b/inst/include/Rcpp/vector/Subsetter.h
@@ -147,7 +147,7 @@ private:
     }
     #else
     template <typename IDX>
-    void check_indices(IDX* x, IDX n, IDX size) {}
+    void check_indices(IDX* x, R_xlen_t n, R_xlen_t size) {}
     #endif

     void get_indices( traits::identity< traits::int2type<INTSXP> > t ) {
```

Can you take a peek?

kevinushey:
Should be good now. We now give a warning rather than a hard error; I indicated in NEWS that this might change in the future.

eddelbuettel:
> We now give a warning rather than a hard error

In one spot. You left a stop() in the other. On purpose?

kevinushey:
I left https://github.com/RcppCore/Rcpp/pull/1310/files#diff-4d30eca18cf4d872269fc011605a047e2212875dcb25688bd2460986c632064bL144 alone because that was pre-existing code (we already gave a hard error in those cases) but I can switch that to a warning if we want to be uniform?

eddelbuettel:
Oh, I completely missed that we already stopped there / that that bit was not new code.  My bad!

eddelbuettel:
We are down to five now. One is possibly a test setup issue [^1].  One is a package for which CRAN requested other changes.  I will open an issue here with links and try to get in touch with the maintainers -- but first things first: I will merge this, and then make a new incremental release 1.0.12.4 with this merged and make it available via the drat (and r-universe) so that the affected maintainers can test with it.

[^1]: Sometimes tests fail on my end and not at CRAN, here we have a package not getting on with RcppArmadillo on the spot compilation.

eddelbuettel (approved):
I think this is ready now.  Big thank you for preparing it!

Issue #1313, opened 2024-07-03
`HASHTAB` is no longer allowed


mattfidler:
Raising for awareness (I know you are already working through many of these issues).  Currently `HASHTAB` is located here:

https://github.com/RcppCore/Rcpp/blob/518961121d632fe42ad93421027e98459626d724/inst/include/Rcpp/Environment.h#L88

I suppose this can use evaluation to get around this in the mean time.

kevinushey:
Are these so-called `UserDefinedDatabase` objects still in use anywhere? The only reference I see in the R sources are:

https://github.com/wch/r-source/blob/c4706b85233fbcdeefe96c05e52ad01481ef28ee/src/library/base/man/attach.Rd#L78-L80

and the RObjectTables package isn't on CRAN and hasn't been updated in over 10 years. Through that lens, I'm inclined to just remove support from Rcpp (which would allow us to drop our usage of HASHTAB here).

PR #1314, opened 2024-07-05
drop support for user-defined databases

kevinushey:

Closes https://github.com/RcppCore/Rcpp/issues/1313.

#### Checklist

- [x] Code compiles correctly


eddelbuettel:
Nice catch in #1313. I don't even recall what those user-define dbs were....

eddelbuettel:
I am kinda minutes/hours/days from a new release so how would you feel if I delayed this til after the release. Given that @mattfidler saw this I would think we need a full rev.dep check (as we change the API surface) which I would rather avoid now.  Thoughts?

kevinushey:
For sure, definitely no rush on this (or the other PR I just put up).

mattfidler:
For me I worked around it ao of CRAN does ask for an update I am ready.

I dont mind waiting either.

eddelbuettel:
Reverse dependency check is now 50% done, no new regressions yet.

eddelbuettel:
The reverse depends run had no new issues, so merging this now.  Had to do a micro-commit over conflicts in ChangeLog, will do one more clean-up (fixing tabs/spaces etc) once merged and once I increment the micro release.

Issue #1316, opened 2024-07-07
New CRAN warnings on STRING_PTR and VECTOR_PTR


eddelbuettel:
Running a recent r-devel:

<blockquote>

* checking compiled code ... NOTE
File ‘Rcpp/libs/Rcpp.so’:
  Found non-API calls to R: ‘STRING_PTR’, ‘VECTOR_PTR’

Compiled code should not call non-API entry points in R.

See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual,
and section ‘Moving into C API compliance’ for issues with the use of non-API entry points.
* checking sizes of PDF files under ‘inst/doc’ ... OK

</blockquote>

kevinushey:
I can take a gander.

eddelbuettel:
That would be sweet.  I am a little tied up right now but should have time a little later.

PR #1317, opened 2024-07-07
use read-only variants of {STRING/VECTOR}_PTR

kevinushey:

Closes https://github.com/RcppCore/Rcpp/issues/1316. I tried to do this in a way that lets us use the read-only variants with newer versions of R, and fall back to the non-API variants for older versions of R.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/r/compat.h
+++ b/inst/include/Rcpp/r/compat.h
@@ -22,13 +22,15 @@
 #ifndef RCPP_R_COMPAT_H
 #define RCPP_R_COMPAT_H
 
-#if defined(STRING_PTR_RO)
+#include <Rversion.h>
+
+#if R_VERSION >= R_Version(4, 4, 0)
 # define RCPP_STRING_PTR STRING_PTR_RO
 #else
 # define RCPP_STRING_PTR STRING_PTR
 #endif
 
-#if defined(VECTOR_PTR_RO)
+#if R_VERSION >= R_Version(4, 4, 0)
 # define RCPP_VECTOR_PTR VECTOR_PTR_RO
 #else
 # define RCPP_VECTOR_PTR VECTOR_PTR
--- a/inst/include/Rcpp/r/compat.h
+++ b/inst/include/Rcpp/r/compat.h
@@ -24,13 +24,13 @@
 
 #include <Rversion.h>
 
-#if R_VERSION >= R_Version(4, 4, 0)
+#if R_VERSION >= R_Version(4, 4, 2)
 # define RCPP_STRING_PTR STRING_PTR_RO
 #else
 # define RCPP_STRING_PTR STRING_PTR
 #endif
 
-#if R_VERSION >= R_Version(4, 4, 0)
+#if R_VERSION >= R_Version(4, 4, 2)
 # define RCPP_VECTOR_PTR VECTOR_PTR_RO
 #else
 # define RCPP_VECTOR_PTR VECTOR_PTR
--- a/inst/include/Rcpp/r/compat.h
+++ b/inst/include/Rcpp/r/compat.h
@@ -1,0 +1,37 @@
+
+//
+// compat.h: Rcpp R/C++ interface class library -- compatibility defines
+//
+// Copyright (C) 2024         Dirk Eddelbuettel, Kevin Ushey
+//
+// This file is part of Rcpp.
+//
+// Rcpp is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// Rcpp is distributed in the hope that it will be useful, but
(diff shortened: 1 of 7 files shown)


inst/include/Rcpp/r/compat.h
```diff
@@ -0,0 +1,37 @@
...
+// along with Rcpp.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef RCPP_R_COMPAT_H
+#define RCPP_R_COMPAT_H
+
+#if defined(STRING_PTR_RO)
+# define RCPP_STRING_PTR STRING_PTR_RO
+#else
+# define RCPP_STRING_PTR STRING_PTR
+#endif
```
eddelbuettel, line 29: Very nice. I wondered about the need to make this conditonal on existence of the new macro or min R version.

src/barrier.cpp
```diff
@@ -50,7 +54,8 @@ void char_set_string_elt(SEXP x, R_xlen_t i, const char* value) {
 
 // [[Rcpp::register]]
 SEXP* get_string_ptr(SEXP x) {
-    return STRING_PTR(x);
+    // TODO: should we deprecate this?
```
eddelbuettel, line 57: May need a full rev.dep run to see if it is being called?

eddelbuettel:
Reverse-depends check running. Looks good so far but many more packages to go ...

src/barrier.cpp
kevinushey: Maybe we leave this code as-is for this PR, and then later test their removal in a separate PR, to see what breaks?

eddelbuettel: Sounds like a plan!

eddelbuettel:
@kevinushey I just ran an `r-devel CMD check --as-cran` and I still got the warning on this branch.  

```sh
edd@rob:~/git/rcpp(bugfix/vector-string-ptr-read-only)$ git ls|head -4
* edcf98f8 - (HEAD -> bugfix/vector-string-ptr-read-only, origin/bugfix/vector-string-ptr-read-only) one more const (30 hours ago) <Kevin Ushey>
* 1cb485b5 - another const_cast (31 hours ago) <Kevin Ushey>
* cd0f41ac - use read-only variants of {STRING/VECTOR}_PTR (31 hours ago) <Kevin Ushey>
* 51896112 - (origin/master, origin/HEAD, master) Release 1.0.12.4 (2 weeks ago) <Dirk Eddelbuettel>
edd@rob:~/git/rcpp(bugfix/vector-string-ptr-read-only)$ 
```

I named that 1.0.12.4.1 and it is the one I am testing now but:

```sh
edd@rob:~/git/rcpp(bugfix/vector-string-ptr-read-only)$ rdcc.sh Rcpp_1.0.12.4.1.tar.gz                                                                                            
* using log directory ‘/tmp/r/Rcpp.Rcheck’                                                                                                                                        
* using R Under development (unstable) (2024-07-07 r86877)                                                                                                                        
* using platform: x86_64-pc-linux-gnu                                                                                                                                             
* R was compiled by                                                                                                                                                               
    gcc (Ubuntu 13.2.0-4ubuntu3) 13.2.0                                                                                                                                           
    GNU Fortran (Ubuntu 13.2.0-4ubuntu3) 13.2.0                               
* running under: Ubuntu 23.10                                                            
* using session charset: UTF-8                                                           
* using option ‘--as-cran’                                                               
* checking for file ‘Rcpp/DESCRIPTION’ ... OK                                               
* this is package ‘Rcpp’ version ‘1.0.12.4.1’     
[....]
* checking compiled code ... NOTE
File ‘Rcpp/libs/Rcpp.so’:
  Found non-API calls to R: ‘STRING_PTR’, ‘VECTOR_PTR’

Compiled code should not call non-API entry points in R.

See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual,
and section ‘Moving into C API compliance’ for issues with the use of non-API entry points.
* checking sizes of PDF files under ‘inst/doc’ ... OK
```

Any idea why?

src/api.cpp
```diff
@@ -112,7 +112,7 @@ namespace Rcpp {
         case BCODESXP:  	return "BCODESXP";
         case EXTPTRSXP: 	return "EXTPTRSXP";
         case WEAKREFSXP:    return "WEAKREFSXP";
-#if R_Version >= R_Version(4,4,0)					// replaces S4SXP in R 4.4.0
+#if R_VERSION >= R_Version(4,4,0)					// replaces S4SXP in R 4.4.0
```
kevinushey, line 115: This one seemed a bit strange; I'm not sure if this really worked before (or, if it did, _why_ it did)?

eddelbuettel: Yeah that looks like you spotted an 'ooops'. Thank you!

And thanks for the update.  It look good.  Will have to run another rev.dep run but we should be good by end of week. Fingers crossed!

PR #1324, opened 2024-08-20
Skip one more test on macOS

eddelbuettel:
We are seeing an additional (new) failure on macOS and arm64 at r-universe ([recent build log](https://github.com/r-universe/rcppcore/actions/runs/10464014682/job/28977023009)).  We already skip a number of `NA` related tests on that platform so I added one.  I cannot test this locally -- anybody reading this who has an arm64 mac and the `clang` that @jeroen reports, namely

```
clang --version
Apple clang version 15.0.0 (clang-1500.1.0.2.5)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -1,0 +1,40 @@
+# Run CI for R using https://eddelbuettel.github.io/r-ci/
+
+name: ci
+
+on:
+  push:
+  pull_request:
+
+env:
+  _R_CHECK_FORCE_SUGGESTS_: "false"
+
+jobs:
+  ci:
+    strategy:
+      matrix:
+        include:
+          - {os: macOS-latest}
+          #- {os: ubuntu-latest}
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup
+        uses: eddelbuettel/github-actions/r-ci-setup@master
+
+      - name: Bootstrap
+        run: ./run.sh bootstrap
+
+      - name: Dependencies
+        run: ./run.sh install_deps
+
+      - name: Test
+        run: ./run.sh run_tests
+
+      #- name: Coverage
+      #  if: ${{ matrix.os == 'ubuntu-latest' }}
+      #  run: ./run.sh coverage
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -1,6 +1,6 @@
 # Run CI for R using https://eddelbuettel.github.io/r-ci/
 
-name: ci
+name: macos
 
 on:
   push:
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2024-08-20  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/tinytest/test_sugar.R: Skip one more NA related test on arm64
+	and macOS as failure is seen at r-universe on arm64
+
 2024-08-19  Dirk Eddelbuettel  <edd@debian.org>
 
 	* Contributing.md: Refreshed content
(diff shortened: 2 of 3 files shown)


Enchufa2 (approved):
Weird. Especially because the max version, right below, seems to work fine.

eddelbuettel:
Do you happen to have suitable hardware to test this?

Enchufa2:
No, I don't. There's a Mac at home that I could borrow, but unfortunately it's Intel.

eddelbuettel:
We'll fit a creative solution.

jeroen:
@Enchufa2 note that this test is not executed for the "release" versions of Rcpp, but only for the `x.x.x.x` versions so that is why the problem did not surface before: 

https://github.com/RcppCore/Rcpp/blob/26fcdb275287c1c3243837c3c64c0e3efabdb217/tests/tinytest.R#L12-L21

Enchufa2:
@jeroen Yes, I noted that. I was referring to line 1574, which has a similar test but with `intmax` instead of `intmin`, and the build log at r-universe reports no ERROR for that one.

Enchufa2:
AFAIK, unless the session aborts or something, everything runs. E.g., put the following into a test file:

```R
library(tinytest)
expect_true(FALSE)
expect_true(TRUE)
expect_true(FALSE)
```

Then:

```bash
$ Rscript -e 'tinytest::run_test_file("test.R")'
test.R........................    3 tests 2 fails 25ms
----- FAILED[data]: test.R<2--2>
 call| expect_true(FALSE)
 ----- FAILED[data]: test.R<4--4>
 call| expect_true(FALSE)
 
Showing 2 out of 3 results: 2 fails, 1 passes (25ms)
```

eddelbuettel:
Yes you are correct. I think I confused myself mentally comparing with "another popular flavour" for unit tests which we still use in another package that keeps me busy at work.

kevinushey:
Looking at the sugar code, we try to cast `R_PosInf` to the appropriate storage type here:

https://github.com/RcppCore/Rcpp/blob/9db1de44245f863edadd6da6b6d07d06770b13c6/inst/include/Rcpp/sugar/functions/min.h#L62-L63

Distilling this into a smaller example, I used:

```
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
SEXP casting_test() {
    int value = static_cast<int>(R_PosInf);
    return Rf_ScalarInteger(value);

}

/*** R
casting_test()
*/
```

On my ARM macOS machine:

```
$ R -s -e 'Rcpp::sourceCpp("abc.cpp")'

> casting_test()
[1] 2147483647
```

And on an x86_64 Docker image:

```
root@f00a2fde26d3:~/scratch# R -s -e 'Rcpp::sourceCpp("test.cpp")'

> casting_test()
[1] NA
```

And some StackOverflow searching suggests that attempting to cast `INFINITY` to an integer is undefined behavior, e.g. https://stackoverflow.com/questions/38795544/is-casting-of-infinity-to-integer-undefined.

To wit -- should we just return `NA` here? (if the storage type is not `double`)

eddelbuettel:
Yes. That seems like a bit of logic error / thinko given 

```r
> Inf                                                                                                                                                                                                              
[1] Inf                                                                                                                                                                                                            
> as.integer(Inf)                                                                                                                                                                                                  
[1] NA                                                                                                                                                                                                             
Warning message:                                                                                                                                                                                                   
NAs introduced by coercion to integer range                                                                                                                                                                        
>        
```

Should we warn just like R does here?

kevinushey:
Perhaps... I'm not sure how closely we try to model the base R equivalents in the sugar functions.

It is a little awkward for us that `min(integer())` returns `Inf`, which is a `double` value, whereas our sugar functions try to preserve the same type. So we're forced to diverge a bit no matter what.

```
> class(min(integer()))
[1] "numeric"
Warning message:
In min(integer()) : no non-missing arguments to min; returning Inf
```

eddelbuettel:
Hoh boy.  Isn't `min(integer()) == Inf` a bug?   Maybe `-Inf` ?  Hm.

PR #1325, opened 2024-08-28
Support Authors@R in Rcpp.package.skeleton()

eddelbuettel:
This small PR changes the helper function creating an Rcpp package to use an Authors@R entry with c("aut", "cre").  Tests have bee adjusted; we just to test for Author and Maintainer being set so now we check for Authors@R and ... Date.

#### Checklist

- [x] Code compiles correctly


kevinushey (approved):
LGTM!

Issue #1329, opened 2024-09-04
"variadic templates" capability is misreported


Enchufa2:
I see:

```R
> Rcpp:::capabilities()["variadic templates"]
variadic templates 
             FALSE
```

This is checked here:

https://github.com/RcppCore/Rcpp/blob/dfa585d2b568db5c8c0ff0e88e7f857758abccd4/src/api.cpp#L174-L178

But `HAS_VARIADIC_TEMPLATES` is never set (note that the defines are commented out):

https://github.com/RcppCore/Rcpp/blob/dfa585d2b568db5c8c0ff0e88e7f857758abccd4/inst/include/Rcpp/platform/compiler.h#L66-L97

Enchufa2:
I traced this code back to 270c856f8e1fa58f3b342dec880c54340d0df52f that comes from [this discussion](https://lists.r-forge.r-project.org/pipermail/rcpp-devel/2012-December/004829.html). Those three lines ended up commented out after [this message](https://lists.r-forge.r-project.org/pipermail/rcpp-devel/2012-December/004842.html), that reports some issues with such feature enabled. However:

1. That is a long time ago, like gcc 4.7 time ago.
2. All the uses of `HAS_VARIADIC_TEMPLATES` in the current codebase are in the `#if defined(HAS_VARIADIC_TEMPLATES) || defined(RCPP_USING_CXX11)` checks added by @andrjohns as part of his PR.
3. This means that everything that defining such a macro enables is already enabled.
4. And it's already on CRAN and well tested.

So uncommenting these 3 lines has no impact beyond correctly reporting this capability. Do you want me to open a PR? Or should we kindly ask @andrjohns to enable such defines as part of #1328, and then maybe also simplify the above checks by dropping the `|| defined(RCPP_USING_CXX11)` part?

PR #1336, opened 2024-10-07
Variadic templates refinement

### LINKED ISSUE #1328: Did something change with void return of Modules functions?
Fresh CRAN email reports that my RcppCNPy package now differs between the `saveAndLoad.Rout` it creates and the `saveAndLoad.Rout.save` I have in the package, in particular calls to `npySave()` now return a `NULL` but did not before.  This is a standard Rcpp Modules binding of a `void` function (see [here](https://github.com/eddelbuettel/rcppcnpy/blob/602e8568884444f1052f17c6732b7fcef9f0a771/src/cnpyMod.cpp#L113-L187)).

Did we change anything in how this is generated within Rcpp?  I cannot think of anything. Am I missing something?

Enchufa2:
Closes #1328, #1329. Refining #1303.

- An `is_void` method was missing in @andrjohns great work. Adding it fixes the issue.
- New tests were added.
- The definition of `HAS_VARIADIC_TEMPLATES` is uncommented now, and all the ifdefs have been simplified accordingly.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/Module.h
+++ b/inst/include/Rcpp/Module.h
@@ -112,6 +112,7 @@ namespace Rcpp {
             }
 
             inline int nargs() { return sizeof...(T); }
+            inline bool is_void() { return std::is_void<RESULT_TYPE>::value; }
             inline void signature(std::string& s, const char* name) { Rcpp::signature<RESULT_TYPE, T...>(s, name); }
             inline DL_FUNC get_function_ptr() { return (DL_FUNC)ptr_fun; }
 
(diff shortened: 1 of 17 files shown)


kevinushey (approved):
LGTM!

inst/include/Rcpp/Module.h
```diff
@@ -112,6 +112,7 @@ namespace Rcpp {
             }
 
             inline int nargs() { return sizeof...(T); }
+            inline bool is_void() { return std::is_void<RESULT_TYPE>::value; }
```
eddelbuettel, line 115: Great catch!

eddelbuettel (approved):

PR #1338, opened 2024-10-25
Fix compilation on DragonFlyBSD

gaborcsardi:
Fixed compilation on DragonFlyBSD.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/exceptions_impl.h
+++ b/inst/include/Rcpp/exceptions_impl.h
@@ -23,15 +23,16 @@
 
 // disable demangler on platforms where we have no support
 #ifndef RCPP_DEMANGLER_ENABLED
-# if defined(_WIN32)     || \
-    defined(__FreeBSD__) || \
-    defined(__NetBSD__)  || \
-    defined(__OpenBSD__) || \
-    defined(__CYGWIN__)  || \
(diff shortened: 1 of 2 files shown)


inst/include/Rcpp/exceptions_impl.h
```diff
@@ -23,15 +23,16 @@
...
-    defined(__CYGWIN__)  || \
-    defined(__sun)       || \
-    defined(_AIX)        || \
-    defined(__MUSL__)    || \
-    defined(__HAIKU__)   || \
+# if defined(_WIN32)       || \
+    defined(__FreeBSD__)   || \
+    defined(__NetBSD__)    || \
+    defined(__OpenBSD__)   || \
+    defined(__DragonFly__) || \
```
eddelbuettel, line 30: Looks good!

eddelbuettel:
What's the easiest way to run that BSD variant?  Container?  VM?  RHub2?

Also, if I may: you appear to have me blocked on mastodon which is not helpful as we seem to both answer the same question more often than not -- and I cannot see what you answered which makes not duplicating redundant content harder.  If you could, maybe give this some consideration and maybe change it to a mute if I irritate you so much.

PR #1344, opened 2024-11-03
Fix typos in the vignettes

mcol:
Corrects a number of typos I've found. ~~I don't think this requires an entry in the Changelog, but let me know if you'd like one.~~ Scratch that, I've noticed that in the past one was added, so there's a second commit for that.


ChangeLog
```diff
@@ -1,3 +1,14 @@
+2024-11-03  Marco Colombo <mar.colombo13@gmail.com>
```
eddelbuettel, line 1: Two spaces between date and name (which you have) and name and email (one missing).

One empty line after the header.

eddelbuettel:
Thanks for all that attention to detail, I will take a closer look later and likely merge.  Much appreciated!

ChangeLog
mcol: Done :smile:

eddelbuettel (approved):
Really appreciate the attention to detail here catching all these little errors.

PR #1348, opened 2024-11-25
Adjust datetime format in test to not tickle change in R-devel (closes #1347)

### LINKED ISSUE #1347: R-devel changes from last week's dev day lead to datetime check bug
####  Issue

R Contributors had a Dev Day last week in conjunction with LatinR.  @hturner and I worked on an old datetime issue which we got resolved and merged to R-devel thanks to @mmaechler.  Now, of course, no small good deed goes unpunished as the buglet just found here reveals that the fix in R-devel may have been suboptimal.  I just sent a suggested fix to the Bugzilla bug tracker on that issue which is [here](https://bugs.r-project.org/show_bug.cgi?id=17350).

#### Suggested Fix

We can work around this by supplying a slightly different fixed format string:

```diff
diff --git a/inst/tinytest/test_date.R b/inst/tinytest/test_date.R
index 84891ae5..a24e2668 100644
--- a/inst/tinytest/test_date.R
+++ b/inst/tinytest/test_date.R
@@ -194,7 +194,7 @@ expect_equal(Datetime_format(d,"%Y-%m-%d %H:%M:%S"),
              format(d, "%Y-%m-%d %H:%M:%OS"),
              info="Datetime.formating.default")
 expect_equal(Datetime_format(d, "%Y/%m/%d %H:%M:%S"),
-             format(d, "%Y/%m/%d %H:%M:%OS"),
+             format(d, "%Y/%m/%d %H:%M:%OS6"),
              info="Datetime.formating.given.format")
 expect_equal(Datetime_ostream(d),
              format(d, "%Y-%m-%d %H:%M:%OS"),
```

which avoids the isssue tickled and will do no harm whether my additional change gets accepted or not.

eddelbuettel:
As described in #1347, @hturner and I landed a patch in R itself which now tickles a test difference in CI under the update R-devel.  The fix is simple and changes the format string to one that will be passed through leading again to this test of ours passing as before.  The change here should not get affected if R-devel changes again.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,9 @@
 	* DESCRIPTION (Version, Date): Roll micro version to 1.0.13.6
 	* inst/include/Rcpp/config.h (RCPP_DEV_VERSION): Idem
 
+	* inst/tinytest/test_date.R: Minor adjustment to datetime format test
+	following R Dev Day change for R bug 17350 (and issue #1347 here)
+
 2024-11-25  Simon Guest  <simon.guest@tesujimath.org>
 
 	* src/attributes.cpp: fix CPP source having to be writable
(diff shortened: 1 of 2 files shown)


Enchufa2 (approved):

eddelbuettel:
Thanks!

PR #1353, opened 2025-01-01
Protect another R_NO_REMAP with preceding #ifndef

eddelbuettel:
I noticed a second `#define R_NO_REMAP` in the (very much optional) file `Benchmark/Timer.h` so adding a now-required preceding `#ifndef / #endif` around it as R 4.5.0 now defines `R_NO_REMAP` for C++ -- just like we have done since 2009 (!!). 
 There was an earlier PR #1296 where I protected the main #define, but missed this.

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
LGTM!

PR #1354, opened 2025-01-18
Fix trivial typos in NEWS

bbolker:


eddelbuettel (approved):
Thanks much!  Care to add yourself and this change to ChangeLog?   C-x 4 a from the file makes it pretty easy in case you use _that_ editor...

bbolker:
done. I'm not very experienced at squashing commits. If that's a thing you care about I could either figure out how to do that, or delete this PR and push a new one ...

ChangeLog
```diff
@@ -1,3 +1,7 @@
+2025-01-18  Ben Bolker  <bolker@pop-os>
```
eddelbuettel, line 1: Can you make it a real (as in working) email, pretty please?  (There is an option for that to override I could look up, until then just hand-edit)

bbolker: oops.

eddelbuettel:
Squashing is the best.  Nomatter how many individual ones you pile up we end up with one.

(I became a fan of amending commits though when 'n + 1' is just a subtle refinement of 'n' and to then force-push (if in branch / fork))

PR #1355, opened 2025-01-25
Expand libraries vignette with PACKAGE_types.h

eddelbuettel:
This comes from [a discussion over at RcppEigen](https://github.com/RcppCore/RcppEigen/issues/141) which revealed that our trick of using `PACKAGE_types.h` to declare types known to `RcppExports.cpp` is not as well known.  

So a brief reference was added to Rcpp-libraries vignette which is somewhat topical.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/vignettes/rmd/Rcpp-libraries.Rmd
+++ b/vignettes/rmd/Rcpp-libraries.Rmd
@@ -165,7 +165,7 @@ package) containing the required \code{\#include} statement for the type(s) will
 compilation; see the 'Rcpp Attributes' vignette for details \citep{CRAN:Rcpp:Attributes}.
 
 It may also be a dependency on a fairly standard library available on most if
-not all systems.  The graphics formats bmp, jpeg or png may be example; text
+not all systems.  The graphics formats bmp, jpeg or png may be an example; text
 formats like JSON or XML are another.  One difficulty, though, may be that
 _run-time_ support does not always guarantee _compile-time_ support. In these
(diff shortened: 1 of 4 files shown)


vignettes/rmd/Rcpp-libraries.Rmd
```diff
@@ -154,15 +154,24 @@ it.
...
+or \pkg{AsioHeaders} \citep{CRAN:AsioHeaders} or many other examples.
+
+One aspect worth noting is that if you include a type in your function interface it will
+also be part of the generated \code{RcppExports.cpp}. In this case adding a file
+\code{PACKAGE\_types.h} (where \code{PACKAGE} is to be replaced with the name of your
+package) containing the required \code{\#include} statement for the type(s) will permit
+compilation; see the 'Rcpp Attributes' vignette for details \citep{CRAN:Rcpp:Attributes}.
+
+It may also be a dependency on a fairly standard library available on most if
+not all systems.  The graphics formats bmp, jpeg or png may be example; text
```
Enchufa2, line 168: "may be example**s**", or "may be **an** example"

Enchufa2 (approved):
Typo below; otherwise, LGTM!

vignettes/rmd/Rcpp-libraries.Rmd
eddelbuettel: Nice catch. That half-sentence is likely five years old.  Will fix.

PR #1356, opened 2025-01-30
std::string_view support in Rcpp::wrap()

evalon32:
Currently, `Rcpp::wrap()` treats a `std::string_view` argument by default as a generic container and creates a character vector from single characters. This is different from how `std::string` is handled. This change aligns the two, so that `wrap("foo"sv)` is equivalent to `wrap("foo"s)`. The added code is conditionally compiled for C++17. 

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2025-01-31  Lev Kandel  <lmakhlis@google.com>
+
+	* inst/include/Rcpp/internal/wrap.h: Add support for std::string_view
+	* inst/include/Rcpp/traits/r_type_traits.h: Idem
+	* inst/include/Rcpp/traits/wrap_type_traits.h: Idem
+	* inst/include/RcppCommon.h: Include <string_view>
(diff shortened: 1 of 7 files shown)


eddelbuettel:
Would you mind filling in the template you carried over unchanged?

evalon32:
This is a draft I created solely to reference in an issue I'm about to open to start a discussion :)

eddelbuettel:
Looking forward to it -- ex ante it's a good topic and something that would be nice to fill in.

eddelbuettel:
(Updated your first box above: if you make it `[x]` without the space you had `[ x]` then the GH flavored markdown takes it as a completed tick box.)

This looks good. I will toss this at the reverse dependency checker which, being underpowered on an ancient old box, will take a while.  We will know more in a few days.

eddelbuettel:
There were no regressions, or, in the parlance, "changes to worse" so no objections from the reverse-depends check.

eddelbuettel (commented):
LGTM so merged.

PR #1358, opened 2025-02-11
Extend Rcpp.package.skeleton to support URL and BugReports

eddelbuettel:
This is a minor PR that adds the ability to use a 'githubuser' field which, when given, leads to URL and BugReports fields being added to DESCRIPTION.  It also flips default package version to a saner '0.0.1'.

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
Tested, works nicely.

PR #1359, opened 2025-03-07
Convert remaining DATAPTR to DATAPTR_RO

eddelbuettel:
As CRAN checks nag about the one remaining `DAtAPTR` use, a quick change to `DATAPTR_RO` which _should_ pass for all reasonable use cases.  Running full reverse-depends checks now (which will take its usual time ... measured in days now) to determine if anybody out there "is naughty" instead of nice.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,9 @@
 2025-03-07  Dirk Eddelbuettel  <edd@debian.org>
 
+	* DESCRIPTION (Version, Date): Roll micro version and date
+
+	* inst/include/Rcpp/config.h: Idem
+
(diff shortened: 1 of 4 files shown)


kevinushey:
LGTM!

eddelbuettel:
:crossed_fingers: 

Rev deps looking good too so far, but it's only been a fraction...

kevinushey:
We probably need some way of eventually deprecating / removing this, but that feels like a "Later" kind of issue.

eddelbuettel:
With the `_RO` being blessed, do we really have to?

PR #1362, opened 2025-03-13
Use lsInteral3 instead of lsInternal

eddelbuettel:
This came to our attention via a CC in email from @kurthornik in the context of helping the @RfastOfficial team.  The change is benign as R has `Rf_lsInternal3` (with three arguments, we were not using the third for 'sorted or not') and asking 'blame' reveals this signature is at least 11 years old so we do not need a version check methinks.

This change is so cosmetic and we are not changing exposed APIs here so I may for once not bother to wait for the 2 1/2 days of a full reverse dependency check.  (And the first batch of CI from the commit before the PR was formed is all green too.)
 
#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,8 @@
 2025-03-13  Dirk Eddelbuettel  <edd@debian.org>
 
+	* DESCRIPTION (Version, Date): Roll micro version and date
+	* inst/include/Rcpp/config.h: Idem
+
 	* inst/include/Rcpp/Environment.h: Switch from R_lsInternal to
 	R_lsInternal3 as the former is now outlawed
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2025-03-13  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/Environment.h: Switch from R_lsInternal to
+	R_lsInternal3 as the former is now outlawed
+
 2025-03-10  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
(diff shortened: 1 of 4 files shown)


kevinushey (approved):
LGTM!

eddelbuettel:
In that context just spotted section 'Some Backports' in the R-devel version of WRE.  We may want to borrow this as needed:

```
#if R_VERSION < R_Version(4, 4, 1)
#define allocLang Rf_allocLang

SEXP Rf_allocLang(int n)
{
    if (n > 0)
          return LCONS(R_NilValue, Rf_allocList(n - 1));
    else
          return R_NilValue;
}
#endif

#if R_VERSION < R_Version(4, 5, 0)
# define isDataFrame(x) Rf_isFrame(x)
# define R_ClosureFormals(x) FORMALS(x)
# define R_ClosureEnv(x) CLOENV(x)
# define R_ParentEnv(x) ENCLOS(x)

SEXP R_mkClosure(SEXP formals, SEXP body, SEXP env)
{
    SEXP fun = Rf_allocSExp(CLOSXP);
    SET_FORMALS(fun, formals);
    SET_BODY(fun, body);
    SET_CLOENV(fun, env);
    return fun;
}

void CLEAR_ATTRIB(SEXP x)
{
    SET_ATTRIB(x, R_NilValue);
    SET_OBJECT(x, 0);
    UNSET_S4_OBJECT(x);
}
#endif
```

eddelbuettel:
I turned the rev.dep beast on ("old habits die hard") but may not wait for its conclusion.

eddelbuettel:
Rev.dep is a quarter done and no issues.  I let it run a little longer but may merge before it is done and then rebase the other branch too.

Issue #1366, opened 2025-03-19
Remove use of `HAS_VARIADIC_TEMPLATES`


eddelbuettel:
With #1364 having established C++11 as a baseline, we can now proceed and remove use of `HAS_VARIADIC_TEMPLATES` ... along with the large number of generated files it uses for the old pre-C++11 case.

eddelbuettel:
This is complete via the merge of #1367 following as-expected uneventful reverse-dependency checks.

PR #1364, opened 2025-03-14
Header cleanup with C++11 as a baseline

eddelbuettel:
Rcpp has worked very hard for very long to provide powerful idioms irrespective of the compiler encountered.  This was mostly due to 'modern C++' coming around very slowly when Rcpp got going.  It is now 2025, and C++11 itself starts to feel like 'legacy code' so we can move forward and discard accommodations for compilation standards older than C++11.  

This will allow us to remove a number of branches from numerous if/else blocks and remove included generated code.  Last years nice and very careful PR #1303 already went that way, we can now go further.  This PR starts by making the [platform/compiler.h](https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/platform/compiler.h) header simpler by removing the various tests:  if C++11 is given, a number of things can be asserted and defined.  We can build on it and examine the different defines and one-by-one remove conditional use ... and eventually the define.  (While being careful.  Invariably one or two client packages may be using the define.)

#### Checklist

- [x] Code compiles correctly
- [x] `R CMD check` still passes all tests
- [ ] Preferably, new tests were added which fail without the change
- [x] Document the changes by file in [ChangeLog](https://github.com/RcppCore/Rcpp/blob/master/ChangeLog)

PR #1367, opened 2025-03-19
Use variadic templates unconditionally

eddelbuettel:
As described in #1366 and building on PR #1364, we can simplify the code by no longer switching based on definition of `HAS_VARIADIC_TEMPLATE`:  only the positive case will be considered given the baseline of C++11.

So this PR removed this file by file, testing each commit locally in CI and arrived at removing the definition.  A reverse dependency check is now running as well.

#### Checklist

- [x] Code compiles correctly


eddelbuettel:
Don't get to do that every day :laughing: 

image

kevinushey:
> Don't get to do that every day 😆
> 
> image

It's beautiful!

eddelbuettel:
I was a little surprised it amounted to so little in the tar.gz:

```
edd@rob:~/git/rcpp(feature/variadic_templates)$ ls -tltr Rcpp_1.0.14.[78].tar.gz
-rw-r--r-- 1 edd edd 2972038 Mar 16 09:51 Rcpp_1.0.14.7.tar.gz
-rw-r--r-- 1 edd edd 2889287 Mar 18 20:37 Rcpp_1.0.14.8.tar.gz
edd@rob:~/git/rcpp(feature/variadic_templates)$ 
```

kevinushey (approved):
LGTM!

eddelbuettel:
Been burned before so I will probably await the reverse-depends.  Maybe just maybe somone relied on the define.  Otherwise this is pretty easy (if volumunious): no new code, only removal of code we no longer compiled in. Seemingly no fat-fingered typo ... :crossed_fingers:

inst/include/Rcpp/DottedPair.h
```diff
@@ -35,14 +35,11 @@ RCPP_API_CLASS(DottedPair_Impl),
...
-			Storage::set__(pairlist(args...));
-		}
-	#else
-		#include <Rcpp/generated/DottedPair__ctors.h>
-	#endif
+    template <typename... T>
+	DottedPair_Impl(const T&... args) {
+        Storage::set__(pairlist(args...));
+    }
+    // RM #include <Rcpp/generated/DottedPair__ctors.h>
```
Enchufa2, line 42: Leftover or reminder?

Enchufa2 (approved):
Nice!

inst/include/Rcpp/DottedPair.h
eddelbuettel: Leftover.  Thanks for catching this.  Gone in ddefda2dc0c8aaced0ac1a4c541e24a4f1ccc94e -- and I promptly mistyped '[ci skip]' dropping the p.  Oh well.

Issue #1365, opened 2025-03-14
C++11 as minimum supported standard


Enchufa2:
Meta-issue for discussion and to keep track of all the individual changes required to achieve this goal.

--- 

It may be helpful to define three classes of tasks and an ordering as DE sees it
- First order: establish C++11 as the baseline allowing variadic templates (and the 'can use C++11' define) to reduce to unconditional code which, as a major side effect and first order goal, leads to removal of all the pre-C++11 autogenerated code. Other cleanups below inst/include and in src/ happen basically as a side effect
- Second order, by a wide margin: cleanup below inst/ and removal of old helpers or illustrations which does not really affect anyone
- Third order, and possibly tied with or marginally ahead of the previous point: cleanup of R helper functins

---

#### Tasks

- [x] Clean up `inst/include/Rcpp/platform/compiler.h`. Ancient tests and defines that are not required anymore.
  #1363
  #1364 
- [x] Review `HAS_VARIADIC_TEMPLATES` checks and remove carefully, also cleaning up `inst/include/Rcpp/generated` stuff.  (DE: I expect this to be one single careful PR)
  #1366
  #1367
- [x] Review `RCPP_USING_CXX11` checks and remove carefully. **RInside** uses this macro (see [this search](https://github.com/search?q=org%3Acran+-repo%3Acran%2FRcpp+RCPP_USING_CXX11&type=code)). Then, the definition could be removed. (DE: Thanks for the heads-up. I can take this.)
  #1369 
- [x] Find remaining `if __cplusplus < 201103L` and assess removal
  #1370 
  #1371
- [x] Review `RCPP_USING_UNORDERED_*` checks and remove carefully. Review `RCPP_UNORDERED_*` types and just replace with the appropriate `std::unordered_*` type.
  #1372
- [x] Remove branch in `inst/include/Rcpp/sugar/functions/sapply.h:41` (see TODO there). (DE: Unsure about this one, lesser priority for me. To be determined...). (IU: Maybe the other way around: maybe we should just remove the first branch, the one that uses the std library. Then `::Rcpp::traits::result_of` could (or could not) be reimplemented based on the std library.)
  #1373 
[...]

PR #1369, opened 2025-03-22
Remove RCPP_USING_CXX11 which is now implicit

eddelbuettel:
This PR, under the umbrella of meta issue #1365, builds upon PRs #1367 and #1364.  It is a little simpler as it merely removed `RCPP_USING_CXX11` wherever it is used (given that it is now always `#define`'d) and removes the stale `#else` cases, if any.  It all fit into one commit.

A reverse-depends run has been started and is about 100 packages in, this is not expected to reveal anything but we may as well let it run its course again.

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
LGTM!

Issue #1370, opened 2025-03-24
Cleanup __cplusplus checks based on C++11 availability


Enchufa2:
- Remove paths that rely on older standards.
- Include >= C++11 paths unconditionally.
- Assess TODO in `inst/include/Rcpp/sugar/functions/sapply.h:42`
- Then, `RCPP_HAS_LONG_LONG_TYPES` is defined unconditionally, remove checks.

Enchufa2:
What to do with

https://github.com/RcppCore/Rcpp/blob/30edbf5b0b63511712eeb7d279f0638cfc101510/inst/include/Rcpp/sugar/functions/sapply.h#L33-L45

remains. We could remove the second branch as part of #1371, but I think that it would probably be better to remove **the first branch**, because I don't see how this is needed _just here_, and then a future improvement could be to reimplement `::Rcpp::traits::result_of` using `std` stuff.

Enchufa2:
I pushed a branch [`feature/result_of_function`](https://github.com/RcppCore/Rcpp/compare/master...feature/result_of_function) that rewrites `result_of` to generalize the fix proposed for `sapply`, so that also `lapply`, `mapply` and `outer` can work with lambdas. What do you think?

eddelbuettel:
That sounds like a good way forward.

Enchufa2:
Ok, then I'll rebase it when the current PR is merged, add more tests with lambdas, etc., and open a new PR for this.

eddelbuettel:
That sounds even better!

PR #1373, opened 2025-03-29
Enable lambdas in all sugar functions

Enchufa2:
Part of #1365, as discussed in #1370. Takes the idea in #296 that enabled lambdas for sapply, and generalizes it as a variadic `result_of` that now is applied to lapply, mapply and outer too, enabling lambdas for them too.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2025-03-31  Dirk Eddelbuettel  <edd@debian.org>
+
+	* DESCRIPTION (Version, Date): Roll micro version and date
+	* inst/include/Rcpp/config.h: Idem
+
 2025-03-29 Iñaki Ucar <iucar@fedoraproject.org>
 
 	* inst/include/Rcpp/traits/result_of.h: Reimplement traits::result_of using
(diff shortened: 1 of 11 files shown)


kevinushey (approved):
LGTM!

eddelbuettel:
As discussed over DM as it happened, the reverse depends run for this one tossed up three packages coughing so to be refined or to be patched there.

Enchufa2:
I think that 8ba2d7b4854a53fc18417fd6ed5ae257397d2641 fixes the issue. I dropped that before just because it was giving me problems, and collided somehow with the other specialization. But it works now, so my previous issues were probably due to some intermediate state of the PR.

Enchufa2:
Stupid mistake... I inadvertently dropped `typedef` at some point. Now `expertsurv`, `flexsurv` and `texmex` should compile just fine.

eddelbuettel:
Very nice, and confirmed for those three plus one that was flakey in the last run (and now passed too).  Waiting on one new package....

eddelbuettel (approved):
Thumbs up!

PR #1379, opened 2025-05-05
Accommodate NA behavior on arm64/aarch64 for Linux as well

### LINKED ISSUE #1378: Sugar test failure on arm64
@jeroen now builds under arm64 on Linux, and this leads to four Sugar tests failing.  From the [current log](
https://github.com/r-universe/rcppcore/actions/runs/14834592257/job/41643213186)

```
Status: 1 ERROR
See
  ‘/__w/rcppcore/rcppcore/Rcpp.Rcheck/00check.log’
for details.

Error: Check: tests, Result: ERROR
    Running ‘tinytest.R’ [223s/195s]
  Running the tests in ‘tests/tinytest.R’ failed.
  Last 13 lines of output:
    ----- FAILED[data]: test_sugar.R<69--69>
     call| expect_true(is.na(fx(c(NA, 6))))
    ----- FAILED[data]: test_sugar.R<77--77>
     call| expect_true(is.na(fx(c(5, NA))))
    ----- FAILED[data]: test_sugar.R<86--86>
     call| expect_true(is.na(fx(c(NA, 1))))
    ----- FAILED[data]: test_sugar.R<1623--1623>
     call| expect_true(is.na(intmin(integer(0))), "min(integer(0))")
     info| min(integer(0))
    Error: 9 out of 1612 tests failed
    Execution halted
```

tt so happens that r2u also has arm64 support so should be easy for me to poke at in CI so will take a look.

eddelbuettel:
Fixes #1378

The linux-on-arm platform is still reasonably rare and hence not one that CRAN tests for, by r-universe now does and this revealed that the same 'NA does not propagate well' issue we accommodate for macos-on-arm already.  This PR simply extends that treatmeant.

A new CI run has been added for linux-arm, it is by far the slowest so one open question is whether we should keep it for all runs.  Maybe a simpler approach is to take it out of the matrix it is currently in but submit a new yaml file with just the 'workflow_dispatch' option so that we dial in when we want to.  Thoughts?

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/linuxarm.yaml
+++ b/.github/workflows/linuxarm.yaml
@@ -1,10 +1,11 @@
 # Run CI for R using https://eddelbuettel.github.io/r-ci/
 
-name: macos
+name: linuxarm
 
 on:
-  push:
-  pull_request:
+  #push:
+  #pull_request:
+  workflow_dispatch:
 
 env:
   _R_CHECK_FORCE_SUGGESTS_: "false"
@@ -14,10 +15,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - {os: macos-latest}
-          - {os: macos-13}
           #- {os: ubuntu-latest}
-          #- {os: ubuntu-24.04-arm}
+          - {os: ubuntu-24.04-arm}
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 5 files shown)


Enchufa2 (approved):
LGTM!

eddelbuettel:
Any objections to folding the arm-linux test out to optional CI now that we know it works?  I have that ready here, and I think I like that better.

Enchufa2:
No objections.

PR #1385, opened 2025-05-24
Add DateVector and DatetimeVector default ctors (closes #1384)

### LINKED ISSUE #1384: Add default ctor for Date(time)Vector classes
One of the more amusing aspects of Rcpp is just how spectacularly _bad_ my design for the (original) Date and Datetime vector (and scalar !) types was.  (In my defense, I knew little R and was driven by what QuantLib did with `std::vector`.)  I did a first cleanup a decade (give or take) ago, but given the (not needed, really) scalar classes I omitted default constructors for zero length vectors.

R 4.5.0 now behaves differently for zero-length vectors. It shows less tolerance, and bonks out more clearly.  This became apparent this week in [issue #135 in `anytime`](https://github.com/eddelbuettel/anytime/issues/135) where I now (in a branch for now, release "soon") cover the case of construction from `integer()` or `numeric()`.  But this lead to some introspection here too (ahem!!) and I sketched in two simple default constructors which seem to work.  As this simply expands the API surface I do not expect trouble.

eddelbuettel:
As discussed in #1384 these were omitted in earlier implementations, mostly due to sloppy thinking and a bias towards the data(time) in `std::vector`  use seen in another library that was interfaced at the very (very !!) beginning of Rcpp.   But that is simply not the R way of things, and while we did on pass over this (when the then `newDate*Vector` classes were added we still did not add empty constructors.  Which I now use more often for base types (think `integer()` etc) so we should have something like that here too.  

The PR itself is pretty minimal and just adds these new ctors, does not alter existing interfaces, and adds a handful of test predicates.

A complete reverse-dependency check is running but it will (as usual) be a good moment before it is done.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2025-05-23  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/date_datetime/newDateVector.h: Add default
+	constructor
+	* inst/include/Rcpp/date_datetime/newDatetimeVector.h: Idem
+
 2025-05-06  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,9 @@
 	constructor
 	* inst/include/Rcpp/date_datetime/newDatetimeVector.h: Idem
 
+	* inst/tinytest/test_date.R: Add tests
+	* inst/tinytest/cpp/dates.cpp: Idem
+
 2025-05-06  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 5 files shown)


Enchufa2 (approved):
Small and focused, as you said. No further comments.

kevinushey (approved):
LGTM!

eddelbuettel:
Reverse-depends worked out as well:
- [run one](https://github.com/RcppCore/rcpp-logs/blob/master/results/Rcpp-20250524.txt) had a few new depends, plus depends on `arrow` and `RcppPlanc` neither one will install on that old machine which cannot do AVX2 but the configuration scripts are not careful enough
- [run two](https://github.com/RcppCore/rcpp-logs/blob/master/results/Rcpp-20250526.txt) is cleaner and just two known (caused by their upstream) issues

With that we're good so ... :shipit:

PR #1395, opened 2025-07-20
Remove unused and no-longer-needed function compilerCheck

eddelbuettel:
During the previous PR I noticed this unused function from 2016 hanging around.  A search under organization `cran` reveals no package using it, and we are generally in a better situation with respect to what the compilers provide / what we need.  So while I am generally very careful about removals, this one should be easy.

#### Checklist

- [x] Code compiles correctly


Enchufa2 (approved):
Wasn't exported, so shouldn't break anything.

eddelbuettel:
Correct.  

Next up are the exported and long-deprecated RcppLdPath() and friends and because they were exported I am fearing breakage.  We shall see ...

Issue #1406, opened 2025-11-11
Changes in downstream packages required by #1402


eddelbuettel:
PR #1402 addresses the long-standing issue #1247 by masking `Rf_error()` and preferring `Rcpp::stop()`.  A relatively small number of packages do need a minor adjustment each: often just a removal of an `#include <R.h>` which may mess with the `NO_REMAP` macro, or a simple re-run of `compileAttributes()`.  This issue will regroup these changes and is based on the 2025-11-04 reverse dependency run summarized [in this text file](https://github.com/RcppCore/rcpp-logs/blob/master/results/Rcpp-20251104.txt) as mention in [this comment](https://github.com/RcppCore/Rcpp/pull/1402#issuecomment-3488126548).

The checklist will be expanded "in due course" to record PRs (or emailed patches).

#### Group 1:  Including `R.h` which is best avoided

This is an issue we should be more proactive about (as for example [RcppArmadillo is w.r.t. `Rcpp.h`](https://github.com/RcppCore/RcppArmadillo/blob/master/inst/include/RcppArmadillo.h#L24-L26) as it can lead to issues with re-definitions of `error()` as seen here.  It is usually both harmless, and without side-effects -- until it isn't. 
It consists on removing `#include <R.h>` as well as any other R headers, which are already loaded by Rcpp safely. The change can be made anytime before or after we land the patch in Rcpp.

- [ ] BayesProject: patch **emailed** 2025-12-21
- [ ] GeneralizedWendland: patch **emailed** 2025-12-21
- [ ] locStra: patch **emailed** 2025-12-21
- [x] LOMAR: patch **emailed** 2025-12-21, on CRAN 2025-12-22
- [x] SAM: patch **emailed** 2025-12-21, on CRAN 2026-02-19
- [ ] TDA: patch **emailed** 2025-12-21

#### Group 2: Use `-DRCPP_NO_MASK_RF_ERROR` if needed

Some packages have a complicated-enough setup, possibly with other callbacks into R, making use of `Rf_error()` unavoidable.  In those cases a `#define` can be used to restore the status quo.  These package can apply the change anytime.

- [x] ggiraph: davidgohel/ggiraph/pull/343 filled 2025-12-22, merged 2026-01-16, version 0.9.3 on CRAN 2026-01-19
- [x] mmrm: openpharma/mmrm/pull/556 filled 2025-12-22, merged 2026-01-07, on CRAN 2026-01-08
[...]

PR #1407, opened 2025-11-12
Lay the ground for Rf_error masking (#1402)

Enchufa2:
First step to #1247

* Add a message at compilation time
* Protect the valid `Rf_error` calls generated by Rcpp

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/macros/mask.h
+++ b/inst/include/Rcpp/macros/mask.h
@@ -21,7 +21,7 @@
 #define Rcpp_macros_mask_h
 
 #ifndef RCPP_NO_MASK_RF_ERROR
-/** cannot emit a message, no warning yet
+#ifdef RCPP_MASK_RF_ERROR
 #define Rf_error(...) \
     _Pragma("GCC warning \"Use of Rf_error() instead of Rcpp::stop(). Calls \
 to Rf_error() in C++ contexts are unsafe: consider using Rcpp::stop() instead, \
@@ -29,7 +29,7 @@ or define RCPP_NO_MASK_RF_ERROR if this is a false positive. More info:\n\
  - https://github.com/RcppCore/Rcpp/issues/1247\n\
  - https://github.com/RcppCore/Rcpp/pull/1402\"") \
     Rf_error(__VA_ARGS__)
-**/
+#endif
 #endif
 
 #endif


Enchufa2:
So, as discussed, and the revdep machine made it clear, we cannot generate a note. Pragma messages are considered statements by gcc (see e.g. [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78657) and bugs linked there). Options are: be silent, or generate a warning without replacing.

In order to ask CRAN about what they think, we need to calculate how many packages would generate this warning. An upper bound was provided [here](https://github.com/RcppCore/Rcpp/issues/1247#issuecomment-3457163679) by just grepping the sources of all reverse dependencies: **263 packages**. From these, only 24 would also fail compilation or checks as documented in #1406. To further refine this list, we could collect the install logs and grep for the pragma (are they in the revdep machine from the previous run?).

inst/include/Rcpp/macros/mask.h
```diff
@@ -22,7 +22,7 @@
 
 #ifndef RCPP_NO_MASK_RF_ERROR
 #define Rf_error(...) \
-    _Pragma("message \"Use of Rf_error() instead of Rcpp::stop(). Calls \
+    _Pragma("GCC warning \"Use of Rf_error() instead of Rcpp::stop(). Calls \
```
eddelbuettel: I don't think it matters either way but in RcppArmadillo I used `#pragma message`.  Kurt was able to filter these with a heads-up.  That would be one softer than a warning.

Enchufa2:
This is a version that emits a message instead of a warning, and avoids the problem of inserting two statements. Could we run a revdep with this?

eddelbuettel (approved):
This looks good to me (modulo a micro-comment), let's see what the reverse dependencies say (in due time, now running).

(I asked copilot too which often catches little gotchas.)

eddelbuettel:
First package to break compilation is `BayesProject`:

```
/usr/local/lib/R/site-library/Rcpp/include/Rcpp/macros/mask.h:42:23: error: invalid use of ‘::’ [-Wtemplate-body]
   42 | #define Rf_error(...) Rcpp::internal::Rf_error(__VA_ARGS__)
      |                       ^~~~
/usr/share/R/include/R_ext/Error.h:84:15: note: in expansion of macro ‘Rf_error’
   84 | #define error Rf_error
      |               ^~~~~~~~
/usr/local/lib/R/site-library/RcppEigen/include/Eigen/src/IterativeLinearSolvers/IterativeSolverBase.h:305:14: note: in expansion of macro ‘error’
  305 |   RealScalar error() const
      |              ^~~~~
```

It's an interaction with ~Boost~ Eigen and R :crying_cat_face:   This one also triggers the 'R.h included' message so a cleanup may fix both.

Enchufa2:
Ah, of course, group 1 in #1406 is expected to fail due to the inclusion of `<R.h>`. But we want to fix these as soon as possible.

Enchufa2:
Apologies for the back-and-forth,  but the internal template thing didn't work. I thought that the message would be shown only if the template was instantiated, but that's not the case. I thought I tested it correctly, but obviously I did something wrong, because now I see messages all over the place when `Rf_error` is not there.

So we are back at square 1, meaning two options:

1. Prepare things, no warning yet, but at least `compileAttributes` will start protecting `Rf_error` calls. I.e. the PR as it is right know, because I've commented out the warning.
2. Uncomment the warning, so we will start generating warnings in hundreds of packages, although none of them would fail, it's just the warnings. CRAN knows and could "activate" this warning when they see fit.

Enchufa2:
Let's take (1) and merge this then?

eddelbuettel:
Yes. Let me catch my breath after all the excitement in #1430 and get back to you.

eddelbuettel:
Should we maybe not comment it out (as right now) but make it active behind an 'opt-in' environment variable?  Then we could merge now, and we (and Kurt) can run with the variable set.  Better?

Enchufa2:
Sounds good to me. Here it is with an on switch.

eddelbuettel:
Looks like a double env var :)  That'll do.

Enchufa2:
Yes, it's a great idea. CRAN may inject that flag when they're ready to process such warnings. In June the warning may be unconditional (unless disabled explicitly), then next year this may be an error. Thanks, merging here!

eddelbuettel:
So for completeness I ran a reverse-dependency check over what is now in the main branch as 1.1.0.13, and a single package came up with an issue: a `std::min( an_int, new_nrow())` balked because we now return `R_xlen_t`.  I made a quick PR for the affected package (see [here](https://github.com/r-lidar/lidR/pull/841), I also slipped in a 'vistor' to replace one `ATTRIB()`) so we should be good to go.

Issue #1410, opened 2025-11-23
Warn (or initially message) if R headers included before Rcpp headers


eddelbuettel:
Over the years, programming practice has evolved and improved when it comes to header inclusion and spillover of defintions.  For example, as of release 4.5.0 R now makes `R_NO_REMAP` a default for inclusion from C++ code ([WRE, Chapter 6](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#The-R-API)), and  Rcpp has long done so too via the file `inst/include/Rcpp/r/headers.h`.  The `#define STRICT_R_HEADERS` is another example.

To be most effective, the header includes should happen in a controlled order.  As Rcpp itself includes the R headers _after making certain definition itself_ it is preferable to _not_ include R headers before Rcpp does.  So far, we have not checked for this, and it may be time to change this.

Conceptually, things start for us from `RcppCommon.h` which first includes `Rcpp/r/headers.h`.  We could precede this with header file checking for symbols i.e. something along the lines of

```c
#if defined(R_R_H) & defined(USING_R)
#pragma message "R.h has been included before any Rcpp headers. This can lead to hard-to-debug errors, \
       and is generally not necessary. See https://github.com/RcppCore/Rcpp/issues/1410"
#endif
```

We could leave this at 'message' level for some time (a year?) and then switch to 'warning'.  For comparison, RcppArmadillo relies on a fairly intricate interplay between Rcpp headers and its own and made it an error of Rcpp was included first. This has been in place for years, and works well.  We probably also want to check for `Rinterface.h`, `Rdefines.h` and `Rinternals.h`, and possibly more.  I have a local test branch I can push is other want to play along.

PS We could and should also over an override to skip the check in case someone bulk compiles or installs and doesn't want to see it, i.e. wrap it all in `#if !defined(RCPP_SKIP_CHECK_FOR_R_H)` or alike.

Enchufa2:
> We could leave this at 'message' level for some time (a year?) and then switch to 'warning'. For comparison, RcppArmadillo relies on a fairly intricate interplay between Rcpp headers and its own and made it an error of Rcpp was included first. This has been in place for years, and works well. We probably also want to check for `Rinterface.h`, `Rdefines.h` and `Rinternals.h`, and possibly more. I have a local test branch I can push is other want to play along.

+1

Enchufa2:
[...]

PR #1411, opened 2025-11-24
Emit message if R.h included, can turn to warning

eddelbuettel:
This addresses #1410 and adds a message if `R.h` (or one of three other headers) is seen before `Rcpp.h` is included.

#### Checklist

- [x] Code compiles correctly


eddelbuettel:
Yes, as uncool it is to admit it, I now turn them on all the time.  It is especially good at the nitpicky stuff.

eddelbuettel:
Ok, having waited the polite 24 hours for veto or comment (hi @kevinushey we know you're busy so no worries) I will now fold this in.

PR #1418, opened 2025-12-04
Refactor some R_ext includes

### LINKED ISSUE #1417: R-devel removal of `R_ext/Callbacks.h`
Dear all,

when I submitted a new release to CRAN earlier today, it was [rejected](https://win-builder.r-project.org/incoming_pretest/openxlsx2_1.22_20251204_083244/Windows/00install.out) due to a recent removal of  `R_ext/Callbacks.h` in R devel: https://github.com/r-devel/r-svn/commit/671453f7c085d503dff55b533797f58d789e5510

https://github.com/RcppCore/Rcpp/blob/2b0a44bfa06fd8eacc53255b5974a382934692f6/inst/include/RcppCommon.h#L76

I haven't had time to look into this further (e.g. if the `Callback.h` line can be removed). Just wanted to let you know. I have no rush and can wait, but presumably this might block all incoming packages with `Rcpp` as dependency.

Enchufa2:
Closes #1417 

- Drop `<R_ext/Callbacks.h>`, not used anymore and dropped from the API.
- Move `<R_ext/Visibility.h>` to `Rcpp/r/headers.h`.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@
 	* inst/include/Rcpp/r/headers.h: Include <R_ext/Visibility.h>
 	* inst/include/RcppCommon.h: Drop <R_ext/Callbacks.h>, not used anymore
 	and dropped from the API; drop <R_ext/Visibility.h>, included above
+	* src/api.cpp: Drop <R_ext/PrtUtil.h>, dropped from the API
 
 2025-12-03  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/inst/include/Rcpp/r/headers.h
+++ b/inst/include/Rcpp/r/headers.h
@@ -2,6 +2,7 @@
 //
 // Copyright (C) 2008 - 2009 Dirk Eddelbuettel
 // Copyright (C) 2009 - 2024 Dirk Eddelbuettel and Romain Francois
+// Copyright (C) 2025        Dirk Eddelbuettel, Romain Francois and Iñaki Ucar
 //
 // This file is part of Rcpp.
 //
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -28,7 +28,6 @@
 using namespace Rcpp;
 
 #include "internal.h"
-#include <R_ext/PrtUtil.h>
 
 #ifdef RCPP_HAS_DEMANGLING
 #include <cxxabi.h>


eddelbuettel:
Thanks for the prompt PR!  The CI check from the 'dev' container probably may mean nothing as the update is weekly and may have missed the change.  Let me update my local build real quick and check locally too.

Enchufa2:
I missed the `PrtUtil.h`, thanks, fixed now.

eddelbuettel:
I am turning the reverse-depends machinery on, though I don't expect too much.  Packages using Rcpp as well as headers from R will have both include directories, and we are testing under R-release so this is bound to be washed over.  But one never knows and we have gotten surprises before ...

kevinushey (approved):
LGTM. I'm a bit surprised these headers are being removed outright rather than just using a compiler pragma warning to note that they'll be removed. At least that would reduce the amount of CRAN breakage in the interim, while still giving package authors a chance to update...

eddelbuettel:
I am equally shocked. It's plain cold breakage and I suspect the CRAN parts of R Core are not too happy either.

Enchufa2:
I missed this one:

https://github.com/RcppCore/Rcpp/blob/2b0a44bfa06fd8eacc53255b5974a382934692f6/inst/include/Rcpp/Interrupt.h#L25

This one is needed for `Rf_onintr`. But I think it's better if we move it to `inst/include/Rcpp/r/headers.h` too. Is this ok? This has no impact, so no need to rerun reverse dependencies.

eddelbuettel:
I would prefer if we did that next cycle.  It is entirely cosmetic and there are dozens of other cosmetic things we could be doing with the headers (the directory layout is a mess, the file naming is inconsistent etc).  We (as well as "other parties") have also learned at times that moving header files around is never quite as clean and simple as you think. Now it is just a call from us to a header of theirs so as you say "should" be low-impact.  But then again I had to put "should" in quotes because you never know.

eddelbuettel:
Reverse-depends concluded, there is not thing new (package `MIC` failed, but it seems to be fighting with `xgboost` not us).  An above-average set of fails, but these are same as in the last run (scrapnel from `rncl` needing to change, plus a handful or packages that already have deadlines at CRAN).  Committed results as https://github.com/RcppCore/rcpp-logs/commit/22496ce31f0e7de2ce3a4df5481cef925c6dc5b7

So squash-merging ahead.

jameslamb:
> _package MIC failed, but it seems to be fighting with xgboost not us_

Arrived here from r-pkg-devel. Saw this comment so thought I could save you some investigation time... I think you could expect to see `{xgboost}` issues in reverse-dependency checks for a few weeks.

That project is in the middle of trying to push out a release with API-breaking changes, and its many (though not "many" compared to Rcpp 😅) reverse dependencies have not all updated: https://github.com/dmlc/xgboost/issues/11430

PR #1419, opened 2025-12-10
Remove access to Rf_rnbeta

eddelbuettel:
It is being removed from R-devel.

#### Checklist

- [x] Code compiles correctly


kevinushey:
LGTM!

eddelbuettel:
Thank you both. Rev.deps running, not expecting anything, but should be done either very late tomorrow or the day after.

eddelbuettel:
There was a bit of scatter in the reverse dependency checks (as a few packages have issues at CRAN including deadlines) but nothing new popped up that is due to this PR so merging.

PR #1426, opened 2025-12-16
Remove checks for R versions older than 3.5.0

### LINKED ISSUE #1425: Update version checks against R in the Rcpp sources
The recent conditioning on R 4.5.0 (for some of the API change accommodations) reminded me to check other instances of _ancient_ conditioning on R_Version() (or getRversion()) in the C and R code of Rcpp.   There are a few, and I propose to establish R 3.5.0 as a (very conservative) floor or minimum that corresponds to what we check in CI as well.  Older checks will be removed and we still assume that the R version used is greater or equal than 3.5.0.

We could discuss moving the demarcation forward to even R 4.0.0 but I don't see any real difference as there are rather few comparisons.  But of course I am all ears to any and all suggestions; I suspect @kevinushey comes across this issue too.

eddelbuettel:
Closes #1425 

This updates a few of the older version comparisons used in the package to provide conditional code and establishes a new baseline of R 3.5.0 -- which is also the oldest version we test in CI.  

There is no formal test for R 3.5.0 or later now. I sort-of dislike declaring it in DESCRIPTION though that is the place for it.  I would be fine to be more cheeky and just check in `configure` (though we currently have no `configure` ...).  Or is that too clandestine?

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2025-12-17  Dirk Eddelbuettel  <edd@debian.org>
+
+	* DESCRIPTION (Depends): Add 'R (>= 3.5.0)'
+
 2025-12-16  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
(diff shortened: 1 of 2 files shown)


Enchufa2 (approved):
LGTM!

eddelbuettel:
Thanks!

I also poked around for a (new, added) `configure` snippet; it may just be easier to add `Depends: R (>= 3.5.0)`.

eddelbuettel:
I will wait a few hours in case any one of us changes their minds over 'Depends: R (>= 3.5.0)' but otherwise this is ready, will become 1.1.0.11 and is likely the basis for 1.1.1 come January.  Yay.  We got a fair number of things done this cycle.

PR #1430, opened 2025-12-22
Move away from using ATTRIB()

### LINKED ISSUE #1429: Replace ATTRIB in three places
@ltierney just now on r-devel:

> As part of clarifying and tightening the R API for packages the
non-API entry points ATTRIB and SET_ATTRIB will be removed from
installed header files and, where possible, hidden from package
access. Both expose internal structure that might need to be changed
and also make it too easy to inadvertently break consistency
assumptions about the values of some attributes.  Using these
functions is will now result in check NOTEs.  These check NOTEs will
soon be replaced by WARNINGs; this will probably happen by the end of
January.  The 'Working with attributes' section in the 'Writing R
Extensions' list some alternatives. Additional functions for working
with attributes, without exposing internal attributes structure, could
be added if they are needed.

We have three spots we can likely replace

```sh
edd@paul:~/git/rcpp(master)$ ag ATTRIB inst/include/ src/
inst/include/Rcpp/proxy/AttributeProxy.h
82:        SEXP attrs = ATTRIB( static_cast<const CLASS&>(*this).get__());
91:        SEXP attrs = ATTRIB(static_cast<const CLASS&>(*this).get__());

inst/include/Rcpp/DataFrame.h
71:            SEXP att = ATTRIB( Parent::get__() )  ;
edd@paul:~/git/rcpp(master)$ 
```

eddelbuettel:
Closes #1429 

As detailed by @ltierney in a post to the r-devel list two days ago, and as tracked in issue #1429, `ATTRIB()` may go away for the R 4.6.0 cycle.  We use it in three spots, and this PR replaces the use by employing the (brand-new, had to re-build the 'r-devel' container we use here in CI as its last build on Monday did not reflect it yet) `R_mapAttrib()` which treats attributes as an opaque object and 'walks' a user-supplied function over it. 

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/proxy/AttributeProxy.h
+++ b/inst/include/Rcpp/proxy/AttributeProxy.h
@@ -97,35 +97,8 @@ public:
         return v;
     }
 
-    struct AttributeProxyStruct {
-        const std::string* match;
-        bool found;
-    };
-
     bool hasAttribute(const std::string& attr) const {
-#if R_VERSION >= R_Version(4, 6, 0)
-        auto visitor = [](SEXP name, SEXP attr, void* data) -> SEXP {
-            AttributeProxyStruct *ptr = static_cast<struct AttributeProxyStruct*>(data);
-            if (ptr->found) return Rf_ScalarInteger(1); 	// already found
-            if (*(ptr->match) == CHAR(Rf_asChar(name))) {
-                ptr->found = true; 							// signal we have match
-                return Rf_ScalarInteger(1);
-            }
-            return NULL;           							// continue
-        };
-        AttributeProxyStruct str{&attr, false};
-        R_mapAttrib(static_cast<const CLASS&>(*this).get__(), visitor, static_cast<void*>(&str));
-        return str.found;
-#else
-        SEXP attrs = ATTRIB(static_cast<const CLASS&>(*this).get__());
-        while( attrs != R_NilValue ){
-            if( attr == CHAR(PRINTNAME(TAG(attrs))) ){
-                return true ;
-            }
-            attrs = CDR( attrs ) ;
-        }
-        return false; /* give up */
-#endif
+        return static_cast<const CLASS&>(*this).attr(attr) != R_NilValue;
     }
 
 


ltierney:
On the `hasAttribute` and `attributeNames` usage: If we added `Rf_existsAttribute` and `R_getAttribNames` one issue is how to handle the implicit `names` attribute for pairlist and language objects. Adding the implicit attribute to the names and returning `TRUE` for exists would be most consistent with the R level and with `Rf_getAttrib`, but would affect code now looking only at `ATTRIB`. Any thoughts from your end?

Enchufa2:
> On the `hasAttribute` and `attributeNames` usage: If we added `Rf_existsAttribute` and `R_getAttribNames`

This would be very convenient for us.

> one issue is how to handle the implicit `names` attribute for pairlist and language objects. Adding the implicit attribute to the names and returning `TRUE` for exists would be most consistent with the R level and with `Rf_getAttrib`, but would affect code now looking only at `ATTRIB`. Any thoughts from your end?

Since `ATTRIB` is not part of the API from now on, this is not a problem.

Enchufa2:
Oh, you meant that, from now on, pairlists and language objects would return a **new** attribute `names` compared to previous versions of Rcpp, right? It's highly unlikely this breaks any existing code, because I don't think many people use Rcpp to manipulate pairlists and language objects.

eddelbuettel:
All good. I mostly wanted to see _if_ I could get away from ATTRIB now as we have the next upload coming up.  We can likely simplify further. Too bad I didn't 'click' on the already possible simpler `nrow()` (esp as I looked at `do_shortRowNames()` just below it !!).   Also, getting `Rf_existsAttribute` and `R_getAttribNames` sounds promising and would have simplified this now.  We will have to check if something is affected by returning an attribute `names`, that is likely an empirical question.

eddelbuettel:
The thought of going the column length for a simpler `nrow` occurred to me too (though I didn't "finish" the thought with a four-line implementation as @ltierney suggested).  Somehow going with `R_RowNamesSymbol)` feels cleaner.  Using it with `Shield<>` makes it a two-liner

```c++
Shield<SEXP> rn = Rf_getAttrib(Parent::get__(), R_RowNamesSymbol);
return Rf_length(rn);
```

which stays closer to R itself at the cost of the one allocation in `Shield`. I believe we can use `Rf_xlength()` here to and return the `R_xlen_t`.  We 'upgraded' from `int` before without issues.  Of course, needs a reverse dependency check too.  

On the other hand rewriting what @ltierney suggested as follows (with a second `XLENGTH`, and a missing `)`) is pretty tempting too

```c++
SEXP x = Parent::get__();
// We can simplify: zero row DFs have no names, else length of all vector same
// by design constraint so can use vector length for desired row count
return (XLENGTH(x) == 0) ? 0 :  XLENGTH(VECTOR_ELT(x, 0))
```

All those variants pass our 1600+ tests (of which several dozen hit data.frame objects).

ltierney:
> On the other hand rewriting what @ltierney suggested as follows (with a second `XLENGTH`, and a missing `)`) is pretty tempting too
> 
> ```c++
> SEXP x = Parent::get__();
> // We can simplify: zero row DFs have no names, else length of all vector same
> // by design constraint so can use vector length for desired row count
> return (XLENGTH(x) == 0) ? 0 :  XLENGTH(VECTOR_ELT(x, 0))
> ```
 Unfortunately my comment is wrong: you can have a data frame with zero columns and non-zero rows:
```
data.frame(row.names = 1:3)
```
So for now going with `R_RowNamesSymbol` is probably best until we  figure out whether Rf_nrows can be modified to handle this or maybe add another function that does. Definitely seems like something that should be in the API.

eddelbuettel:
Thanks so much for checking.  The other two-liner relying on `Shield<SEXP>` for the one allocation here is much safer, and closer to what R does.

I am about to commit the re-done attribute proxy using two lambda visitors too.  I still think copilot is wrong and that `row.names` always the last element meaning we need all of them anyway but heck, on the chance it gains at least sometimes it is worth it.

eddelbuettel:
Those were really helpful by all -- big thanks to you all:  @ltierney @Enchufa2 @JanMarvin !

PR #1435, opened 2026-01-01
Tweak coverage

eddelbuettel:
This PR updates #nocov tags for a number of R and C++ source files.  Apart from one remaining 'function factory' function that confuses the coverage counting, this pushes the needle towards full coverage.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,5 @@
 comment: false
+
 coverage:
   status:
     project:
@@ -8,6 +9,9 @@ coverage:
     patch:
       default:
         target: 0%     # the (on purpose low) required coverage value
+
+ignore:
+  - 'R/inline.R'	# function factory setup confuses codecov
 
 #  layout: "header, diff, tree, changes"
 #  behavior: default
--- a/R/inline.R
+++ b/R/inline.R
@@ -24,7 +24,6 @@ Rcpp.plugin.maker <- function(include.before = "",
                               Makevars      = NULL ,
                               Makevars.win  = NULL,
                               package       = "Rcpp") {
-  	# #nocov start
     function(...) {
         includes <- sprintf("%s
 #include <Rcpp.h>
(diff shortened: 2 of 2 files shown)


inst/include/Rcpp/proxy/NamesProxy.h
```diff
@@ -50,14 +50,14 @@ class NamesProxyPolicy{
             Shield<SEXP> safe_x(x);
 
             /* check if we can use a fast version */
-            if( TYPEOF(x) == STRSXP && parent.size() == Rf_length(x) ){
+            if( TYPEOF(x) == STRSXP && parent.size() == Rf_length(x) ){		// #nocov start
```
Enchufa2: Maybe we should design a test to check the other side of the branch instead?

eddelbuettel: In theory, just about every instance of #nocov could deserve a test.  Some of this areas have been without one for twelve or more years though.  Adding tests is always the hope, in practice it does not happen all that much.

eddelbuettel:
@Enchufa2 Any more (blocking ?) comments?  Can this proceed or do you feel it is not an improvement?   It (in some cases) adds explicit #nocov to code lacking a test for over a decade, as do other spots:

```sh
$ ag -c '#nocov' inst/include/ src/ R/ | awk -F: '{v += $2;} END { printf("Total count is %d\n", v) }'
Total count is 451
$ 
```

Ideally we add new tests. In practice we haven't. This PR adds a few more to the existing tags. No more, no less.  Should not make anything 'worse'.

Enchufa2 (approved):
I might take a look at those things another time, but not now. The problem of increasing coverage based on nocov tags is that it obscures those potential problems, but no strong opinions about it. I'll grep for nocov and that's it.

eddelbuettel:
I think that is a totally fair view.  It's one of "those things" we could be / should be doing but there are often more pressing issues at hand.

PR #1437, opened 2026-01-04
No longer require #ifdef for UNWIND_PROTECT, had been 'on' for long

### LINKED ISSUE #1436: Use UNWIND_PROTECT unconditionally
Rcpp has had support for 'UNWIND_PROTECT' for years, but behind a `#define` with an additional toggle to not define it.  In practice, it was always on.  As R 3.5.0 or later is now the baseline along with the C++11 as the minimally supported standard, we can remove the 'conditionalness' and just turn UNWIND_PROTECT on all the time.  The change is minimal, and passes tests locally.

eddelbuettel:
Closes #1436 

This PR removes the `#define` use of  `UNWIND_PROTECT` as well as the one override allowing it to remain unset.  Effectively we have had this one for a long time, and given the established baseline of R 3.5.0 or later and C++11 or later we can clearly default to using this unconditionally.  This should not constitute a change in behavior.  

A reverse-dependency run is ongoing; no issues expected and none seen so far.

The timing of the PR is a little closer to the upcoming 1.1.1 release than I usually like but as the PR has no new code and should really only be removing some conditional testing as scaffolding we may deem this to be part of the release.  Let me know what you think.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,13 @@
 	* inst/include/Rcpp/r_cast.h: Idem
 	* src/attributes.cpp: Idem
 	* src/barrier.cpp: Idem
+2026-01-04  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/RcppCommon.h: No longer require RCPP_USING_UNWIND_PROTECT
+	* inst/include/Rcpp/r/headers.h: Idem
+	* inst/include/Rcpp/api/meat/Rcpp_eval.h: Idem
+	* inst/tinytest/cpp/stack.cpp: No longer test with RCPP_USING_UNWIND_PROTECT
+	* inst/tinytest/test_interface.R: Idem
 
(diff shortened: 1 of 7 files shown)


eddelbuettel:
Reverse-dependency checks "mostly pass": there is one new issue with package [`later`](https://github.com/r-lib/later) _when using the CRAN version_ which appears to be addressed already in the github repo. I posted a question [there](https://github.com/r-lib/later/pull/241#issuecomment-3715785832) concerning their release plans. [And in the time it took me to type this Charlie already replied in the affirmative so we're good!]

Given that it 'mostly passes' I think this is still a merge and subsequent CRAN upload.  There is the known issue with package `lidR` (and my merged PR takes of it) plus now also `later` but I supposed I could argue with CRAN that both are in fact fixed 'upstream' and awaiting uploads.  That argument has (on occassion) been good enough in the past.

Issue #1442, opened 2026-01-12
Rcpp 1.1.1 could not be installed on FreeBSD


jszhao:
I have R 4.5.2 installed on FreeBSD 15.0, and when I install Rcpp 1.1.1 using C++ compiler: ‘FreeBSD clang version 19.1.7, it gives the following error message:
```
/opt/R/library/00LOCK-Rcpp/00new/Rcpp/libs/Rcpp.so: Undefined symbol "backtrace"
```
If I install Rcpp 1.1.1 using C++ compiler: ‘g++14 (FreeBSD Ports Collection) 14.2.0’, it gives the following error message:
```
/opt/R/library/00LOCK-Rcpp/00new/Rcpp/libs/Rcpp.so: Undefined symbol "backtrace_symbols"
```

It seems that the configure of Rcpp 1.1.1 does not find `libexecinfo`, which is a system library, and provides the backtrace library on FreeBSD system. So on the FreeBSD system, I have to install Rcpp 1.1.1 in R with
```
Sys.setenv("PKG_LIBS" = "-lexecinfo")
install.packages("Rcpp")
```

I also installed libbacktrace package, however, I could not make Rcpp to link the libbacktrace.so in /usr/local/lib

```
> sessionInfo()
R version 4.5.2 (2025-10-31)
Platform: amd64-portbld-freebsd15.0
Running under: FreeBSD pc 15.0-RELEASE-p1 FreeBSD 15.0-RELEASE-p1 releng/15.0-n280999-7bceec30b351 GENERIC amd64

Matrix products: default
LAPACK: /usr/local/lib/R/lib/libRlapack.so;  LAPACK version 3.12.1

locale:
[1] C.UTF-8/C.UTF-8/C.UTF-8/C/C.UTF-8/C.UTF-8

time zone: Asia/Chongqing
tzcode source: system (FreeBSD)

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
[1] compiler_4.5.2
```

eddelbuettel:
Thanks for the report.  As I understand it, adding the linker flag `-lexecinfo` addresses your issue?  We can look into that.

I however do not understand what you want to say in 

> I also installed libbacktrace package, however, I could not make Rcpp to link the libbacktrace.so in /usr/local/lib

Are you overlooking the fact that you may also have to alter `configure.ac` to actually look for the library?

jszhao:
[...]

Issue #1400, opened 2025-10-20
switch to a whitelist for #include <execinfo.h>


bastistician:
Build-time feature detection does not seem to be an option here, but I'd like to suggest using a whitelist for `backtrace()` functionality rather than an ever-growing stoplist (e.g., #1338, #796):

https://github.com/RcppCore/Rcpp/blob/65707515bb392059bf10b773898cb67292d1ea41/inst/include/Rcpp/exceptions_impl.h#L24-L44

What brought me here is that the `__MUSL__` exception (added long ago in #449) is an unreliable kludge as musl itself doesn't define such a macro ([by design](http://openwall.com/lists/musl/2013/03/29/13)).

AFAICS, a suitable whitelist should be as simple as

```C
defined(__GLIBC__) || defined(__APPLE__)
```

where execinfo is built-in. The additional compiler check (gcc|clang) seems to be redundant (probably a relict). The suggested change is similar in spirit to the original commit in #449, which was discarded simply because it forgot to cover macOS. BSDs have long provided execinfo, too, but a web search suggests they require separate linking with `-lexecinfo`.

eddelbuettel:
Philosophically speaking, with default behaviour working as is, I am finding that the specific exclusion list is working well.  When someone on an exotic system has a need, they usually send a PR.  That worked well so far.

So with that I would prefer it you could devise a way to detect your musl environment.   We could also wrap an outer control variable around so that deficient systems that cannot have exceptions can opt to skip them using such an outer variable (off by default, of course).  How exactly do things work out here for you?  Do you fall into the wrong bucket because you still have `gcc`?  

But it is all complicated and we probably have whitelists in other places.  So open for discussion.  We have time til the next release in January anyway <wink>.

eddelbuettel:
I see this has come up recently-ish?

```
6075   │  66854     ripley @item
6076   │  79948     ripley Some headers are not portable: we have just mentioned @file{malloc.h}
6077   │  79948     ripley and often @acronym{CRAN} submissions attempt to use @file{endian.h}.
6078   │  85662     hornik The latter is a @code{glibc} extension: some OSes have
6079   │  79948     ripley @file{machine/endian.h} or @file{sys/endian.h} but some have neither.
6080   │  83523     ripley @c macOS has machine/endian.h.  Solaris had none.
6081   │  87498     hornik Header @file{execinfo.h} is only available on a few OSes: formerly nor
[...]

PR #1445, opened 2026-01-13
Refine demangler setting and execinfo use

eddelbuettel:
PR #1401 updated the detection for `execinfo.h` by relying on a header-finding macro directly which is powerful.  It also removed a set of `#if defined(...)` tests precluding several architectures from trying. This now bit us in issue #1442 as on FreeBSD, the header is found ... but also requires a library.  

Now, we could try to use autoconf to find header and library, and attempt to compile a test program (as we now do for OpenMP in RcppArmadillo and RcppEigen).  But given that the affected architectures here are on the less-widely used side with difficult access for us to test and refine, a simpler strategy would be to punt and again rely on a set of `#if define()` so this PR restores this.  That is plain KISS Principle: Keep it Simple (and) Stupid.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,8 @@
 2026-01-12  Dirk Eddelbuettel  <edd@debian.org>
 
+	* DESCRIPTION (Version, Date): Roll micro version and date
+	* inst/include/Rcpp/config.h: Idem
+
 	* inst/include/Rcpp/exceptions_impl.h: Combine previous and current
 	methods enabling when 'execinfo.h' is found on selected platforms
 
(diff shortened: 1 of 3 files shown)


bastistician:
FWIW and sorry to be a nuisance, I had suggested a KISS solution in #1400 (the "whitelist" approach, where I did mention that "BSDs have long provided execinfo, too, but a web search suggests they require separate linking with -lexecinfo"). That would be much simpler than the now-restored bunch of conditionals (including the non-standard `__MUSL__`), but I understand that this is now settled and should work everywhere, which is the main outcome of this revision. Thanks for maintaining portability!

eddelbuettel:
Thanks.  Might it be worthwhile to add the hint about linking needed to the short paragraph in R-exts.texi?

>  Header @file{execinfo.h} is only available on a few OSes: formerly nor
in MacOS nor Solaris, and currently not on Linux systems (such as Alpine
Linux) using @code{musl}.

PR #1446, opened 2026-01-14
Additional tests for attributes

eddelbuettel:
This pull request is a bit of an experiment in that I asked Claude Sonnet 4.5 (via the Google Antigravity IDE, another Code clone) to examine existing `#nocov` tags and fill in some tests.  And so it did which we can now look at in more detail. 

It is all fairly reasonable if pedestrian (i.e. checking on Windows for permitted filenames etc) but does exactly what it was asked: increase coverage of tests across the codebase by a bit.  If we like this we can repeat as this touched only one C++ and one R file so far (for attributes, in this case).

I did one minimal round of postprocessing so far. In one case we still need the `#nocov`: the windows-specific code will never run under Linux coverage analysis.  And one expected error does not materialize as the compilation fails (cannot have two functions in `cppFunction()` call).

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,26 @@
+2026-01-23  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/DataFrame.h (nrow): Use R_nrow() with R >= 4.6.0
+	* inst/include/Rcpp/proxy/AttributeProxy.h (attributeNames): Use
+	R_getAttribNames() with R >= 4.6.0;
+	* inst/include/Rcpp/proxy/AttributeProxy.h (hasAttribute): Use
+	R_hasAttrib with R >= 4.6.0
+
+2026-01-22  Dirk Eddelbuettel  <edd@debian.org>
+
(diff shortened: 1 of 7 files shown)


eddelbuettel:
So @Enchufa2 @kevinushey what do we do here?  Not a lot of passion from any one of us.  Shall we close it?  Make is smaller?  Any suggestions?

Enchufa2:
Start of semester here, I'll need a few days.

eddelbuettel:
A week later, still no love for it. Maybe best to close it?

Enchufa2:
I'll have some time this weekend for this.

Enchufa2 (approved):
Nice, seems like reasonable changes. Master is at 99.11%. Removing the nocov tags that are addressed here leaves as at 98.76%, and the tests proposed bring us back to 99.12%. In favor.

eddelbuettel:
Thanks for taking the time review.  This one wasn't easy, as we open the door to the bots, but this seems like a reasonable task for it.

PR #1448, opened 2026-01-20
Correct a printf format string reflecting 'unsigned int'

### LINKED ISSUE #1447: -Wformat warning from `include/Rcpp/date_datetime/date_datetime.h:29`
I enabled some more warnings when compiling a package that uses Rcpp and got this warning which seems to be caused by Rcpp itself:
```
In file included from /usr/local/lib/R/site-library/Rcpp/include/Rcpp/date_datetime/date_datetime.h:29:
/usr/local/lib/R/site-library/Rcpp/include/Rcpp/date_datetime/Datetime.h:83:65: warning: format specifies type 'int' but the argument has type 'unsigned int' [-Wformat]
   83 |                 res = ::snprintf(txtsec, 63, "%s.%06d", txtiso, m_us);
      |                                                  ~~~~           ^~~~
      |                                                  %06u
```
It probably doesn't make any practical difference for the result, but it's nice if Rcpp itself doesn't trigger common compiler warnings.

eddelbuettel:
Closes #1447 

Corrects a format string for the `unsigned int` variable used

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2026-01-20  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/date_datetime/Datetime.h (format): Correct a
+	format string reflecting 'unsigned int' rather than 'int'
+
 2026-01-12  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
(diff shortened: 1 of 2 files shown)


Enchufa2:
There is a more general problem in that file, because the private attribute is an unsigned int, but then it is used elsewhere as an int, which will probably bite us in the future.

eddelbuettel:
Agree that it is a smell. I was probably too influenced by the POSIX data structure where I may have seen an `unsigned int`.  It should not matter as we are 'range safe' here, the value (of fractional microseconds) cannot overflow an integer.  Which also means it never needed to be an `unsigned`. Sigh.  We could change it.

PR #1457, opened 2026-02-24
Add protection against integer overflows in sugar functions

### LINKED ISSUE #1454: Integer overflows in sugar functions
From the full test suite under the UBSAN container, as reported by @eddelbuettel:

```
test_sugar.R..................  334 tests OK /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/rowSums.h:56:10: runtime error: signed integer overflow: -2147483648 + -2147483648 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/rowSums.h:56:10
```

Basically,

``` r
Rcpp::cppFunction('
IntegerVector intRowSums(IntegerMatrix x) {
  return rowSums(x);
}')

intRowSums(matrix(rep(.Machine$integer.max, 2), nrow=1))
#> [1] -2
```

We found this one because it happens in the tests, but of course this opens up a whole can of worms if you look at any other sugar function doing sums or products. For example,

``` r
Rcpp::cppFunction('
int intSum(IntegerVector x) {
  return sum(x);
}')

intSum(rep(.Machine$integer.max, 2))
#> [1] -2
```

So I can carefully check that folder, creating failing tests along the way. As for the solution, my proposal would be to use `__builtin_add_overflow` and [the like](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html), which, from what I read, should be available in all modern compilers.

For example,

https://github.com/RcppCore/Rcpp/blob/7ff6ba25225c883f139f6cbf2669d164e7053827/inst/include/Rcpp/sugar/functions/rowSums.h#L55-L57

would become:

```c++
inline void incr(int* lhs, int rhs) {
    if (__builtin_add_overflow(*lhs, rhs, lhs))
        stop("Integer overflow in `rowSums`!");
}
```

Enchufa2:
Closes #1454 

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/inst/include/Rcpp/sugar/operators/minus.h
+++ b/inst/include/Rcpp/sugar/operators/minus.h
@@ -92,7 +92,7 @@ namespace sugar{
 		inline STORAGE operator[]( R_xlen_t i ) const {
 			STORAGE y = rhs[i] ;
 			if( Rcpp::traits::is_na<RTYPE>( y ) ) return y ;
-			return RCPP_SAFE_SUB(lhs[i], y);
+			return lhs[i] - y;
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
@@ -141,7 +141,7 @@ namespace sugar{
 		inline STORAGE operator[]( R_xlen_t i ) const {
 			STORAGE x = lhs[i] ;
 			if( Rcpp::traits::is_na<RTYPE>( x ) ) return x ;
-			return RCPP_SAFE_SUB(x, rhs[i]);
+			return x - rhs[i];
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
@@ -189,7 +189,7 @@ namespace sugar{
 			lhs(lhs_.get_ref()), rhs(rhs_.get_ref()) {}
 
 		inline STORAGE operator[]( R_xlen_t i ) const {
-			return RCPP_SAFE_SUB(lhs[i], rhs[i]);
+			return lhs[i] - rhs[i];
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
@@ -285,7 +285,7 @@ namespace sugar{
 		inline STORAGE operator[]( R_xlen_t i ) const {
 			if( rhs_na ) return rhs ;
 			STORAGE x = lhs[i] ;
-			return Rcpp::traits::is_na<RTYPE>(x) ? x : RCPP_SAFE_SUB(x, rhs);
+			return Rcpp::traits::is_na<RTYPE>(x) ? x : x - rhs;
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
--- a/inst/include/Rcpp/sugar/operators/plus.h
+++ b/inst/include/Rcpp/sugar/operators/plus.h
@@ -100,7 +100,7 @@ namespace sugar{
 		inline STORAGE operator[]( R_xlen_t i ) const {
 			STORAGE rhs_ = rhs[i] ;
 			if( traits::is_na<RTYPE>(rhs_) ) return rhs_ ;
-			return RCPP_SAFE_ADD(lhs[i], rhs_);
+			return lhs[i] + rhs_;
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
@@ -153,7 +153,7 @@ namespace sugar{
 		inline STORAGE operator[]( R_xlen_t i ) const {
 			STORAGE lhs_ = lhs[i] ;
 			if( traits::is_na<RTYPE>(lhs_) ) return lhs_ ;
-			return RCPP_SAFE_ADD(lhs_, rhs[i]);
+			return lhs_ + rhs[i];
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
@@ -205,7 +205,7 @@ namespace sugar{
 			lhs(lhs_.get_ref()), rhs(rhs_.get_ref()){}
 
 		inline STORAGE operator[]( R_xlen_t i ) const {
-			return RCPP_SAFE_ADD(lhs[i], rhs[i]);
+			return lhs[i] + rhs[i];
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
@@ -360,7 +360,7 @@ namespace sugar{
 
 		inline STORAGE operator[]( R_xlen_t i ) const {
 			STORAGE x = lhs[i] ;
-			return Rcpp::traits::is_na<RTYPE>(x) ? x : RCPP_SAFE_ADD(x, rhs);
+			return Rcpp::traits::is_na<RTYPE>(x) ? x : x + rhs;
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
@@ -407,7 +407,7 @@ namespace sugar{
 			lhs(lhs_.get_ref()), rhs(rhs_) {}
 
 		inline STORAGE operator[]( R_xlen_t i ) const {
-			return RCPP_SAFE_ADD(rhs, lhs[i]);
+			return rhs + lhs[i];
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
--- a/inst/include/Rcpp/sugar/operators/times.h
+++ b/inst/include/Rcpp/sugar/operators/times.h
@@ -97,7 +97,7 @@ namespace sugar{
 		inline STORAGE operator[]( R_xlen_t i ) const {
 			STORAGE rhs_ = rhs[i] ;
 			if( traits::is_na<RTYPE>(rhs_) ) return rhs_ ;
-			return RCPP_SAFE_MUL(lhs[i], rhs_);
+			return lhs[i] * rhs_;
 		}
 
 		inline R_xlen_t size() const { return lhs.size() ; }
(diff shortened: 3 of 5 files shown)


Enchufa2:
In principle, sugar functions already handle NAs, so these will be invoked for non-NA cases. But I haven't triaged all the cases yet, so we'll see. If this implementation looks good so far, I'll continue with that.

Enchufa2:
BTW @eddelbuettel I don't see the sugar directory listed in Codecov. Isn't the covr workflow running all the tests?

eddelbuettel:
I don't think we exclude Sugar from tests or coverage.  There is a specific entry point `test_sugar.R` with matching cpp file.

Enchufa2:
Yes, and it requires RunAllRcppTests set to yes, so I was wondering if the covr task is not setting it?

eddelbuettel:
BTW I was 'this month old' when I heard about the boo-boo with `min` here.  We need to be careful.

```r
> Rcpp::evalCpp("std::numeric_limits<int>::min()") # normal
[1] NA
> Rcpp::evalCpp("std::numeric_limits<int>::max()") # normal
[1] 2147483647
> Rcpp::evalCpp("std::numeric_limits<int>::min() * 1.0") # also normal
[1] -2147483648
> Rcpp::evalCpp("std::numeric_limits<int>::min()") # R special
[1] NA
> 
> 
> Rcpp::evalCpp("std::numeric_limits<double>::min()") # WTF: positive number
[1] 2.22507e-308
> Rcpp::evalCpp("std::numeric_limits<double>::lowest()") # normal
[1] -1.79769e+308
> 
> Rcpp::evalCpp("std::numeric_limits<int>::lowest()") 
[1] NA
> Rcpp::evalCpp("std::numeric_limits<int>::lowest() * 1.0") 
[1] -2147483648
> 
```

Mind you we are not running overflow for `double`, are we?  If it is only for `int` aka `int32_t` it may be less complex than I feared.

eddelbuettel:
`covr` runs `R CMD check` which goes via `tests/tinytest.R` as usual so it should.  I recently added another clause there because we had to break mold and upload 'four part' fix up versions like 1.1.0.8 to CRAN.  Maybe I borked the logic / started listening to the wrong environment variable. I will debug in a side thread.

Enchufa2:
> Mind you we are not running overflow for `double`, are we?

No, we are not. The overflow checks are enabled only for integral types. Because doubles do not overflow, they become `Inf`.

eddelbuettel:
> only for integral types

So `int` as we don't really do addition on `character` or `bool`.  And given that Sugar is on `SEXP` we do not worry about, say, `uint8_t` do we?

eddelbuettel:
I validated the assumptions in `tests/tinytest.R` in another debug repo I use for CI, and found no issue. So if coverage really does not run for Sugar then I do not know why.  There is one exclusion line in `codecov.yml` but I do not think it should bite. Hm.

Enchufa2:
> So `int` as we don't really do addition on `character` or `bool`. And given that Sugar is on `SEXP` we do not worry about, say, `uint8_t` do we?

Correct.

Enchufa2:
> I validated the assumptions in `tests/tinytest.R` in another debug repo I use for CI, and found no issue. So if coverage really does not run for Sugar then I do not know why. There is one exclusion line in `codecov.yml` but I do not think it should bite. Hm.

I think that the coverage we see is just what calls into `Rcpp.so`, and nothing else. Because all tests compile cpp code via `sourceCpp` into a temporary directory, and this doesn't even get the required `-O0 --coverage` flags. So the current percentage is very misleading, and fixing this is far from trivial: we would need to compile all cpp code in tests in the package tree with the coverage flags, and then instruct covr to include additional paths to look for gcov files (because I think it only looks into the `src` folder by default).

But anyway, this is another matter. We can discuss this further in another issue.

eddelbuettel:
Good thinking.  Sounds like a testing / coverage package having all the tests in `src/` may help. If we care.

I never know what to of coverage. It's a metric, and as [Goodhart told us decades ago](https://en.wikipedia.org/wiki/Goodhart%27s_law), of course it is being gamed.

kevinushey:
Here's what Claude has to say:

```

  1. Pre-existing bug preserved in diff.h (NA=false specialization)

  inst/include/Rcpp/sugar/functions/diff.h:105-112 (current master):

  STORAGE diff = RCPP_SAFE_SUB(y, previous);  // computes result
  previous = y ;                               // overwrites previous
  previous_index = i+1 ;
  return RCPP_SAFE_SUB(y, previous);           // returns y - y == 0

  The local diff is computed but never returned. After previous = y, the return re-computes y - previous which is
  always 0. This is a pre-existing bug in Diff<RTYPE,false,LHS_T> (integer, no-NA path). The fix is trivial: return
  diff; instead of the second subtraction. Since this PR is touching these exact lines, it would be good to fix it
  here.

  2. Three NA=false scalar-vector operator specializations not converted

  These generic-RTYPE (i.e. used for INTSXP) classes still use raw arithmetic:

  - plus.h ~line 311: Plus_Vector_Primitive<RTYPE,false,T>::operator[] -- rhs + lhs[i] not converted to RCPP_SAFE_ADD
  - times.h ~line 294: Times_Vector_Primitive<RTYPE,false,T>::operator[] -- rhs * lhs[i] not converted to
  RCPP_SAFE_MUL
  - minus.h ~line 380: Minus_Primitive_Vector<RTYPE,false,T>::operator[] -- lhs - rhs[i] not converted to
  RCPP_SAFE_SUB

  These have separate REALSXP specializations that handle double, so the generic template does get instantiated for
  INTSXP. Integer overflow is possible here regardless of the NA=false guarantee.

  3. Bonus bug fix in minus.h (good catch)

  The PR fixes Minus_Primitive_Vector<RTYPE,false,T> from Extractor<REALSXP,false,T> to Extractor<RTYPE,false,T>.
  This is a genuine pre-existing bug -- the wrong Extractor type was being used when RTYPE != REALSXP. Worth calling
  out in the PR description / ChangeLog.

```

Enchufa2:
> Here's what Claude has to say:
> 
>   1. Pre-existing bug preserved in diff.h (NA=false specialization)
> 
>   inst/include/Rcpp/sugar/functions/diff.h:105-112 (current master):
> 
>   STORAGE diff = RCPP_SAFE_SUB(y, previous);  // computes result
>   previous = y ;                               // overwrites previous
>   previous_index = i+1 ;
>   return RCPP_SAFE_SUB(y, previous);           // returns y - y == 0
> 
>   The local diff is computed but never returned. After previous = y, the return re-computes y - previous which is
>   always 0. This is a pre-existing bug in Diff<RTYPE,false,LHS_T> (integer, no-NA path). The fix is trivial: return
>   diff; instead of the second subtraction. Since this PR is touching these exact lines, it would be good to fix it
>   here.

Good catch, will fix.

>   2. Three NA=false scalar-vector operator specializations not converted
> 
>   These generic-RTYPE (i.e. used for INTSXP) classes still use raw arithmetic:
> 
>   - plus.h ~line 311: Plus_Vector_Primitive<RTYPE,false,T>::operator[] -- rhs + lhs[i] not converted to RCPP_SAFE_ADD
>   - times.h ~line 294: Times_Vector_Primitive<RTYPE,false,T>::operator[] -- rhs * lhs[i] not converted to
>   RCPP_SAFE_MUL
>   - minus.h ~line 380: Minus_Primitive_Vector<RTYPE,false,T>::operator[] -- lhs - rhs[i] not converted to
>   RCPP_SAFE_SUB
> 
>   These have separate REALSXP specializations that handle double, so the generic template does get instantiated for
>   INTSXP. Integer overflow is possible here regardless of the NA=false guarantee.

Not sure about that. This code is a bit of a mess, and I would lie if I say I completely understand it, but empirically all cases are covered by tests (vector-vector, primitive-vector, vector-primitive), and these are instantiated for sequences. If I add the safe functions here, the sequence tests don't compile.

>   3. Bonus bug fix in minus.h (good catch)
> 
>   The PR fixes Minus_Primitive_Vector<RTYPE,false,T> from Extractor<REALSXP,false,T> to Extractor<RTYPE,false,T>.
>   This is a genuine pre-existing bug -- the wrong Extractor type was being used when RTYPE != REALSXP. Worth calling
>   out in the PR description / ChangeLog.

Yeap, fixed that, forgot to mention it. Will do.

eddelbuettel:
Reverse dependencies, started with the PR was first changed from draft so before the last commits is fine:
- [full run](https://github.com/RcppCore/rcpp-logs/blob/master/results/Rcpp-20260226.txt)
- [incremental follow-up run](https://github.com/RcppCore/rcpp-logs/blob/master/results/Rcpp-20260301.txt)

This looks _mostly_ clean to me:
- the `gRbase` error is an actual installation failure over our new macros, @Enchufa2 can you take a look?
- the `selection.index` error appears to be them checking for overflow too but hence expecting a different (older) error; it is a brand-new package so this can probably be sorted out if we talk to them
- the `eam` failt due `arrow` is expected, the machine is old and cannot run the (optmised) Linux binary blob provided by `arrow` as too optimized so `eam` will be skipped going forward

Other than that there the 50 I skipped, some because the tests are too long and some because I cannot install. We could check them elsewhere (GitHub Actions?)

Enchufa2:
An issue with sequences again. I've fixed it and increased the test cases to cover this.

eddelbuettel:
I started another run yesterday which will need to finish before we get to test this.

eddelbuettel:
This is working out fine. A subsequent run, summarize [here](https://github.com/RcppCore/rcpp-logs/blob/master/results/Rcpp-20260301-2.txt), resulted in only two failures (not also already accompanied by CRAN deadline).  One was the examined package itself which in the interim had received a new CRAN version, which passes.  The other was `gRbase` which passes under the 'sequences again' update made.

For completeness I am running yet another full run but this is looking like we are there.

eddelbuettel:
The additional run also completed without issues, [summary here](https://github.com/RcppCore/rcpp-logs/commit/591f3fdce8135ad4095ba0b66943da3325158523).

Enchufa2:
Thanks!

PR #1458, opened 2026-02-27
Fail instead of trying to cast Inf to int

### LINKED ISSUE #1455: Inf outside of representable range of type 'int' in sugar functions
Again, from the full test suite under the UBSAN container, as reported by @eddelbuettel:

```
test_sugar.R..................  452 tests OK /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/min.h:36:53: runtime error: inf is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/min.h:36:53 
test_sugar.R..................  454 tests OK /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/max.h:36:53: runtime error: -inf is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/local/lib/R/site-library/Rcpp/include/Rcpp/sugar/functions/max.h:36:53
```

Same issue: we are returning `R_PosInf`/`R_NegInf` if the object has no elements:

https://github.com/RcppCore/Rcpp/blob/7ff6ba25225c883f139f6cbf2669d164e7053827/inst/include/Rcpp/sugar/functions/max.h#L32-L47

So one possible approach would be similar to this one:

https://github.com/RcppCore/Rcpp/blob/7ff6ba25225c883f139f6cbf2669d164e7053827/inst/include/Rcpp/sugar/operators/Comparator_With_One_Value.h#L36-L43

I.e., check the object size on object creation and assign the proper operator: one that returns a `double` with infinity if the object has no elements, or one that returns `STORAGE` with the computed min/max.

Maybe there's a more sophisticated solution using `std::enable_if`.

Enchufa2:
Closes #1455.

`min(integer(0))` shows a warning and results in a numeric value. We cannot do that (because we keep the same type in storage instead of returning a SEXP), so I think that it's best to just throw an error in such cases.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2026-02-27  Iñaki Ucar <iucar@fedoraproject.org>
+
+	* inst/include/Rcpp/sugar/functions/max.h: Fix UB for empty integer
+	* inst/include/Rcpp/sugar/functions/min.h: Idem
+	* inst/tinytest/test_sugar.R: Adapt tests
+
 2026-02-17  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date


Enchufa2:
Thanks, merging this now, will rebase the other one after revdep checks.

eddelbuettel:
Well these are 'behavior changes' (that are obviously merited) so we need rev.deps (in case we need to talk affected maintainers).  "Moar cpus" would be good; back on planet earth we just wait.

Enchufa2:
Ops, you are right, I should have waited. In my defense, if someone was using min/max on an empty integer vector, this would have been flagged on CRAN by UBSAN checks, so it is extremely unlikely that there are any issues in revdep checks after this change.

eddelbuettel:
Agreed. And if there is something we will catch it in the next wave of tests.

PR #1462, opened 2026-03-26
Return dataptr() again to avoid getting an invalid address for empty vectors

### LINKED ISSUE #1461: UBSAN issue can arise when std::copy(ing) an Rcpp::NumericVector of length 0
Testing a package I am contributing to ([gregorkastner/factorstochvol](https://github.com/gregorkastner/factorstochvol)) triggered UBSAN using the [clang-ubsan r-hub container](https://r-hub.github.io/containers/containers.html#clang-ubsan). 

The problem occurs when I `std::copy()` an `Rcpp::NumericVector` of length 0. Here is a minimal reproducible example which triggers the error:
```
$ podman run --rm -ti --pull=newer ghcr.io/r-hub/containers/clang-ubsan
root@bf478e112875:/# R
> pak::pak("Rcpp")

> sourceCpp(code='
  #include <Rcpp.h>
  using namespace Rcpp;
  // [[Rcpp::export]]
  void fun_rcpp(R_xlen_t p) {
  NumericVector vec1(p);
  NumericVector vec2(p);
  Rcout << vec2.begin() << "\\n";
  std::copy(vec1.begin(), vec1.end(), vec2.begin());
}')

> fun_rcpp(0)
0x1
/usr/lib/llvm-22/bin/../include/c++/v1/__string/constexpr_c_functions.h:227:25: runtime error: store to misaligned address 0x000000000001 for type 'double *', which requires 8 byte alignment
0x000000000001: note: pointer points here
<memory cannot be printed>
    #0 0x7398a95b6e5a in double* std::__1::__constexpr_memmove[abi:nqe220103]<double, double>(double*, double*, std::__1::__element_count) /usr/lib/llvm-22/bin/../include/c++/v1/__string/constexpr_c_functions.h:227:5
    #1 0x7398a95b6e5a in std::__1::pair<double*, double*> std::__1::__copy_trivial_impl[abi:nqe220103]<double, double>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy_move_common.h:64:3
    #2 0x7398a95b6e5a in std::__1::pair<double*, double*> std::__1::__copy_impl::operator()[abi:nqe220103]<double, double, 0>(double*, double*, double*) const /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy.h:115:12
    #3 0x7398a95b6e5a in std::__1::pair<double*, double*> std::__1::__copy_move_unwrap_iters[abi:nqe220103]<std::__1::__copy_impl, double*, double*, double*, 0>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy_move_common.h:94:19
    #4 0x7398a95b5cc0 in std::__1::pair<double*, double*> std::__1::__copy[abi:nqe220103]<double*, double*, double*>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy.h:122:10
    #5 0x7398a95b5cc0 in double* std::__1::copy[abi:nqe220103]<double*, double*>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy.h:128:10
    #6 0x7398a95b5cc0 in fun_rcpp(long) /tmp/RtmphbrcTn/sourceCpp-x86_64-pc-linux-gnu-1.1.1/sourcecpp_81dde26d/file81296eccf.cpp:9:3
    #7 0x7398a95b5e19 in sourceCpp_3_fun_rcpp /tmp/RtmphbrcTn/sourceCpp-x86_64-pc-linux-gnu-1.1.1/sourcecpp_81dde26d/file81296eccf.cpp:25:5
    #8 0x7798b1fc3480 in R_doDotCall /tmp/R-devel/src/main/dotcode.c:754:11
    #9 0x7798b1fc506d in do_dotcall /tmp/R-devel/src/main/dotcode.c:1437:11
    #10 0x7798b1ffa906 in Rf_eval /tmp/R-devel/src/main/eval.c:1260:9
    #11 0x7798b1ffb127 in Rf_evalList /tmp/R-devel/src/main/eval.c:3679:12
    #12 0x7798b1ffa800 in Rf_eval /tmp/R-devel/src/main/eval.c:1251:20
    #13 0x7798b1ffd308 in R_execClosure /tmp/R-devel/src/main/eval.c:2389:39
    #14 0x7798b1ffbf3b in applyClosure_core /tmp/R-devel/src/main/eval.c:2302:16
    #15 0x7798b1ffa985 in Rf_applyClosure /tmp/R-devel/src/main/eval.c:2324:16
    #16 0x7798b1ffa985 in Rf_eval /tmp/R-devel/src/main/eval.c:1280:12
    #17 0x7798b20490d7 in Rf_ReplIteration /tmp/R-devel/src/main/main.c:264:23
    #18 0x7798b204a72e in R_ReplConsole /tmp/R-devel/src/main/main.c:317:11
    #19 0x7798b204a72e in run_Rmainloop /tmp/R-devel/src/main/main.c:1235:5
    #20 0x7798b204a79a in Rf_mainloop /tmp/R-devel/src/main/main.c:1242:5
    #21 0x5a9917129c27 in main /tmp/R-devel/src/main/Rmain.c:29:5
    #22 0x7798b1a5ad8f  (/usr/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 095c7ba148aeca81668091f718047078d57efddb)
    #23 0x7798b1a5ae3f in __libc_start_main (/usr/lib/x86_64-linux-gnu/libc.so.6+0x29e3f) (BuildId: 095c7ba148aeca81668091f718047078d57efddb)
    #24 0x5a9917040384 in _start (/opt/R/devel-asan/lib/R/bin/exec/R+0x2d384) (BuildId: 6b2bb379dec6fd1104eb4971c90488e20a001ded)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/lib/llvm-22/bin/../include/c++/v1/__string/constexpr_c_functions.h:227:25
/usr/lib/llvm-22/bin/../include/c++/v1/__string/constexpr_c_functions.h:227:33: runtime error: load of misaligned address 0x000000000001 for type 'double *', which requires 8 byte alignment
0x000000000001: note: pointer points here
<memory cannot be printed>
    #0 0x7398a95b6e74 in double* std::__1::__constexpr_memmove[abi:nqe220103]<double, double>(double*, double*, std::__1::__element_count) /usr/lib/llvm-22/bin/../include/c++/v1/__string/constexpr_c_functions.h:227:5
    #1 0x7398a95b6e74 in std::__1::pair<double*, double*> std::__1::__copy_trivial_impl[abi:nqe220103]<double, double>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy_move_common.h:64:3
    #2 0x7398a95b6e74 in std::__1::pair<double*, double*> std::__1::__copy_impl::operator()[abi:nqe220103]<double, double, 0>(double*, double*, double*) const /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy.h:115:12
    #3 0x7398a95b6e74 in std::__1::pair<double*, double*> std::__1::__copy_move_unwrap_iters[abi:nqe220103]<std::__1::__copy_impl, double*, double*, double*, 0>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy_move_common.h:94:19
    #4 0x7398a95b5cc0 in std::__1::pair<double*, double*> std::__1::__copy[abi:nqe220103]<double*, double*, double*>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy.h:122:10
    #5 0x7398a95b5cc0 in double* std::__1::copy[abi:nqe220103]<double*, double*>(double*, double*, double*) /usr/lib/llvm-22/bin/../include/c++/v1/__algorithm/copy.h:128:10
    #6 0x7398a95b5cc0 in fun_rcpp(long) /tmp/RtmphbrcTn/sourceCpp-x86_64-pc-linux-gnu-1.1.1/sourcecpp_81dde26d/file81296eccf.cpp:9:3
    #7 0x7398a95b5e19 in sourceCpp_3_fun_rcpp /tmp/RtmphbrcTn/sourceCpp-x86_64-pc-linux-gnu-1.1.1/sourcecpp_81dde26d/file81296eccf.cpp:25:5
    #8 0x7798b1fc3480 in R_doDotCall /tmp/R-devel/src/main/dotcode.c:754:11
    #9 0x7798b1fc506d in do_dotcall /tmp/R-devel/src/main/dotcode.c:1437:11
    #10 0x7798b1ffa906 in Rf_eval /tmp/R-devel/src/main/eval.c:1260:9
    #11 0x7798b1ffb127 in Rf_evalList /tmp/R-devel/src/main/eval.c:3679:12
    #12 0x7798b1ffa800 in Rf_eval /tmp/R-devel/src/main/eval.c:1251:20
    #13 0x7798b1ffd308 in R_execClosure /tmp/R-devel/src/main/eval.c:2389:39
    #14 0x7798b1ffbf3b in applyClosure_core /tmp/R-devel/src/main/eval.c:2302:16
    #15 0x7798b1ffa985 in Rf_applyClosure /tmp/R-devel/src/main/eval.c:2324:16
    #16 0x7798b1ffa985 in Rf_eval /tmp/R-devel/src/main/eval.c:1280:12
    #17 0x7798b20490d7 in Rf_ReplIteration /tmp/R-devel/src/main/main.c:264:23
    #18 0x7798b204a72e in R_ReplConsole /tmp/R-devel/src/main/main.c:317:11
    #19 0x7798b204a72e in run_Rmainloop /tmp/R-devel/src/main/main.c:1235:5
    #20 0x7798b204a79a in Rf_mainloop /tmp/R-devel/src/main/main.c:1242:5
    #21 0x5a9917129c27 in main /tmp/R-devel/src/main/Rmain.c:29:5
    #22 0x7798b1a5ad8f  (/usr/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 095c7ba148aeca81668091f718047078d57efddb)
    #23 0x7798b1a5ae3f in __libc_start_main (/usr/lib/x86_64-linux-gnu/libc.so.6+0x29e3f) (BuildId: 095c7ba148aeca81668091f718047078d57efddb)
    #24 0x5a9917040384 in _start (/opt/R/devel-asan/lib/R/bin/exec/R+0x2d384) (BuildId: 6b2bb379dec6fd1104eb4971c90488e20a001ded)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/lib/llvm-22/bin/../include/c++/v1/__string/constexpr_c_functions.h:227:33
```

With `std::vector<double>` instead of `Rcpp::NumericVector` the error will not occur: 
```
> sourceCpp(code='
  #include <Rcpp.h>
  using namespace Rcpp;
  // [[Rcpp::export]]
  void fun_std(size_t p) {
  std::vector<double> vec1(p);
  std::vector<double> vec2(p);
  auto x = vec2.begin();
  Rcout << reinterpret_cast<const void*>(&x) << "\\n";
  std::copy(vec1.begin(), vec1.end(), vec2.begin());
}')
> fun_std(0)
0x7ffc96ba38d0
```

<details>
  <summary>sessionInfo</summary>

  ```
> sessionInfo()
R Under development (unstable) (2026-03-25 r89703)
Platform: x86_64-pc-linux-gnu
Running under: Ubuntu 22.04.5 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so;  LAPACK version 3.10.0

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C
 [3] LC_TIME=en_GB.UTF-8        LC_COLLATE=C
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8
 [7] LC_PAPER=en_GB.UTF-8       LC_NAME=C
[11] LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C

time zone: Etc/UTC
tzcode source: system (glibc)

attached base packages:

other attached packages:
[1] Rcpp_1.1.1

[1] compiler_4.6.0 tools_4.6.0    pak_0.9.2.9000
```

</details>

Interestingly, using [docker.io/rocker/r-devel-ubsan-clang](docker.io/rocker/r-devel-ubsan-clang) instead of the r-hub container does not trigger the error.

Enchufa2:
Closes #1461. This happens because [this](https://github.com/wch/r-source/blob/a619877deecfa4ed3e05e371f8172c11b4a116ad/src/include/Defn.h#L421) returns `0x1` for empty vectors. The question is whether we should do the same check in [here](https://github.com/RcppCore/Rcpp/blob/b2f6ad8d0be085acc43b9f73304655db5962e2df/src/barrier.cpp#L78-L81). I think we should.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 2026-03-26  Iñaki Ucar <iucar@fedoraproject.org>
 
-	* inst/include/Rcpp/internal/r_vector.h: Return null start for empty vectors
-	instead of an invalid pointer, which causes UB in e.g. std::copy
+	* inst/include/Rcpp/internal/r_vector.h: Use dataptr() again to avoid an
+	invalid pointer for empty vectors, which causes UB in e.g. std::copy
 	* inst/tinytest/test_vector.R: Add tests for std::copy
 	* inst/tinytest/cpp/Vector.cpp: Idem
 
--- a/inst/include/Rcpp/internal/r_vector.h
+++ b/inst/include/Rcpp/internal/r_vector.h
@@ -33,23 +33,6 @@ typename Rcpp::traits::storage_type<RTYPE>::type* r_vector_start(SEXP x) {
     return reinterpret_cast<pointer>(dataptr(x));
 }
 
-// add specializations to avoid use of dataptr
-#define RCPP_VECTOR_START_IMPL(__RTYPE__, __ACCESSOR__)                                              \
-    template <>                                                                                      \
-    inline typename Rcpp::traits::storage_type<__RTYPE__>::type* r_vector_start<__RTYPE__>(SEXP x) { \
-        if (Rf_xlength(x) == 0)                                                                      \
-            return NULL;                                                                             \
-        return __ACCESSOR__(x);                                                                      \
-    }
-
-RCPP_VECTOR_START_IMPL(LGLSXP,  LOGICAL);
-RCPP_VECTOR_START_IMPL(INTSXP,  INTEGER);
-RCPP_VECTOR_START_IMPL(RAWSXP,  RAW);
-RCPP_VECTOR_START_IMPL(CPLXSXP, COMPLEX);
-RCPP_VECTOR_START_IMPL(REALSXP, REAL);
-
-#undef RCPP_VECTOR_START_IMPL
-
 /**
  * The value 0 statically casted to the appropriate type for
  * the given SEXP type


eddelbuettel:
That looks great. I could turn on the big bad rev.dep machine but I am not even sure it's needed.

kevinushey:
Are we sure `std::copy()` is going to accept a null pointer here? If we don't already have it, a unit test would be nice -- presumably ASAN / UBSAN builds of Rcpp with those tests would catch this issue.

I'm also fairly certain the actual issue here comes from this bit of cruel code.

https://github.com/wch/r-source/blob/6b5f72eebd9211766bd159e017a167385c93772e/src/include/Rinlinedfuns.h#L135-L142

kevinushey:
Perhaps we can return `dataptr(x)` instead of `NULL`? With the appropriate casting.

Or, alternatively -- instead of using `__ACCESSOR__` here, just use `DATAPTR_RO`, and cast to the appropriate type. But we're sort of cheating since `DATAPTR_RO` is supposed to only give read access to memory, and these might be writable...

I just hate the idea of this zero-length check becoming necessary. :|

Enchufa2:
dataptr has the same problem, and I think we should add the same workaround, because all these accessors end up calling that macro I linked above, which doesn't check the length and therefore return an invalid pointer when the vector is empty. 

std::copy works fine with a null pointer, I tested it with the reprex provided by the update, but of course I can add a test. 

Note also that the vector cache sets start to zero upon initialization, but this is messed up when the update method is called. So this follows the initial idea of having a null pointer as the start of there is no start.

kevinushey:
> dataptr has the same problem, and I think we should add the same workaround, because all these accessors end up calling that macro I linked above, which doesn't check the length and therefore return an invalid pointer when the vector is empty.

Are you sure? `dataptr` goes through R's `DATAPTR_RO`, which shouldn't have the same 0x1 path for empty vectors. Please correct me if I'm wrong.

https://github.com/RcppCore/Rcpp/blob/7ff6ba25225c883f139f6cbf2669d164e7053827/src/barrier.cpp#L77-L81

https://github.com/wch/r-source/blob/6b5f72eebd9211766bd159e017a167385c93772e/src/include/Rinlinedfuns.h#L148-L154

https://github.com/wch/r-source/blob/6b5f72eebd9211766bd159e017a167385c93772e/src/include/Defn.h#L421

Enchufa2:
The last one is `STDVEC_DATAPTR`, which is what I linked above, and has this `+1` unconditionally. Or am I missing something?

Enchufa2:
Oh, I get what you mean, the other macro will return a valid address +1. Is that a valid memory address anyway? In the sense, it may be unallocated.

Enchufa2:
In fact, `DATAPTR` was avoided in #1310 in response to #1308. :-\

kevinushey:
> Oh, I get what you mean, the other macro will return a valid address +1. Is that a valid memory address anyway? In the sense, it may be unallocated.

Good question. I think the C++ standard guarantees that pointers to the "end" of an allocation are still valid, as long as you don't de-reference those pointers. I don't know if that extends to memory allocated from C, though...

Honestly though, the right solution might just be for us to request R Core back out this change upstream.

Enchufa2:
Ok, let's go back to `dataptr()` then without workarounds. The original test case #1461 works without UB. And I've run the entire test suite and there are no issues. (In fact, I've found another UB, but unrelated; another fight for another day).

kevinushey:
Thanks @Enchufa2; sorry for being so nit-picky here!

Enchufa2:
That's fine, that's the purpose of a review!

eddelbuettel:
Reverse-depends results are [done](https://github.com/RcppCore/rcpp-logs/commit/247aae5a6eec02e8a14255be7b8950b8f53aaaa9). The machine shows its age so I needed to double check a few things but 'no change to worse' found.

Enchufa2:
Thanks, merging then!

PR #1469, opened 2026-04-03
Switch to R_getRegisteredNamespace 

eddelbuettel:
Closes #1468 

R-devel (very recently) added `R_getRegisteredNamespace()` allowing a lookup with in namespace without requiring the `R_NamespaceRegistry` symbol which is retired.

This minimal PR adjusts accordingly in one spot, and protects the condition by also selecting on the SVN commit number as the (automated) R-devel build we use in CI has not yet caught up.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2026-04-02  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/Function.h: Further refinement for 4.6.0 to not
+	require R_NamespaceRegistry, using R_getRegisteredNamespace() instead
+
 2026-04-01  Mattias Ellert  <mattias.ellert@physics.uu.se>
 
 	* inst/discovery/cxx0x.R: Set execute permissions for script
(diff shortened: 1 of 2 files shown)


Enchufa2:
Moving the conditional from 4.5.0 to 4.6.0 for `Rf_findVarInFrame`, won't it generate NOTEs for 4.5.0?

eddelbuettel:
Not for checking Rcpp itself.  That may have been a 'during development of 4.6.0' change too.

eddelbuettel:
Rebasing now after your merge. And indeed while that now works for 4.5.3 and R-devel it bites with the older stuff.  Will have to add another `#if` branch.

Enchufa2 (approved):

eddelbuettel:
Behaves locally under R 4.4.3, R 4.5.3 and R 4.7.0 so into reverse dependency checking it goes, first updating the machine to current CRAN and then we'll roll for the 2 1/2 or whatever it takes days.

kevinushey (approved):
LGTM!

eddelbuettel:
This worked out fine, result of reverse depends [committed here](https://github.com/RcppCore/rcpp-logs/commit/643f14c81a8bdaef57ab8ad35ce5496bd6419122)

So ... merging and on to re-running the updated previous one.

PR #1471, opened 2026-04-08
Enhance Rcpp::Nullable::as with explicit cast (Closes #1470)

eddelbuettel:
As discussed in #1470 and building on this morning's discussion at the bottom of #1451, we can make `Nullable<T>` a little nicer.  And that is all this PR does:  for cases where the _implicit_ `as<>()` needs a nudge, we can hide the nudge inside that (member function) `as()` function body accessing the matching `as<>` for the given T.   A little bit of ellbow grease.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2026-04-12  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/Nullable.h (Nullable): 'operator T()' is now
+	behind an opt-in #define
+
 2026-04-09  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,10 @@
 2026-04-09  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
+2026-04-08  Dirk Eddelbuettel  <edd@debian.org>
+
+	* inst/include/Rcpp/Nullable.h (as): Enhance by adding an explicit
+	Rcpp::as<T>() aiding cases where implicit as<> does not suffice
 
 2026-04-07  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,8 @@
 2026-04-09  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
+	* inst/include/Rcpp/Nullable.h (T): Also add 'operator T()'
+
 2026-04-08  Dirk Eddelbuettel  <edd@debian.org>
 
 	* inst/include/Rcpp/Nullable.h (as): Enhance by adding an explicit
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
 	* inst/include/Rcpp/Nullable.h (T): Also add 'operator T()'
+	(conditional on R (>= 4.3.0) as older versions croak)
 
 2026-04-08  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -82,6 +82,16 @@ namespace Rcpp {
         }
 
         /**
+         * operator T() to return nullable object
+         *
+         * @throw 'not initialized' if object has not been set
+         */
+        inline operator T() const {
+            checkIfSet();
+            return Rcpp::as<T>(m_sexp);
+        }
+
+        /**
          * get() accessor for object
          *
          * @throw 'not initialized' if object has not been set
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -81,7 +81,7 @@ namespace Rcpp {
             return m_sexp;
         }
 
-#if R_VERSION > R_Version(4,3,0)
+#if R_VERSION > R_Version(4,3,0) && defined(RCPP_ENABLE_NULLABLE_T)
         /**
          * operator T() to return nullable object
          *
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -126,7 +126,7 @@ namespace Rcpp {
         /**
          * Returns m_sexp as a T
          */
-        inline T as() { return get(); }
+        inline T as() { return Rcpp::as<T>(get()); }
 
         /**
          * Return a clone of m_sexp as a T
--- a/inst/include/Rcpp/Nullable.h
+++ b/inst/include/Rcpp/Nullable.h
@@ -81,6 +81,7 @@ namespace Rcpp {
             return m_sexp;
         }
 
+#if R_VERSION > R_Version(4,3,0)
         /**
          * operator T() to return nullable object
          *
@@ -90,6 +91,7 @@ namespace Rcpp {
             checkIfSet();
             return Rcpp::as<T>(m_sexp);
         }
+#endif
 
         /**
          * get() accessor for object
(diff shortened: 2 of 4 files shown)


Enchufa2 (approved):
Fine, although `Nullable` could be further improved. Having an `as()` method in the first place seems odd, because it deviates from the rest of the API. I.e., objects typically have `T()` and `SEXP()` operators. This one has the latter, and it would be nice to have the former for consistency instead of, or at least on top of, `as()`. Also typically `set()` and `get()` are private. See e.g. [`AttributeProxy`](https://github.com/RcppCore/Rcpp/blob/777a34c5057b64d762d8defb5ec6b77b9a671ecf/inst/include/Rcpp/proxy/AttributeProxy.h#L27-L54), just to name something recent.

eddelbuettel:
TBH I had forgotten these had `as()` and `get()` and I think they are barely used.  The main use, and I just used it some more from another package, is just straight up instantiation as the `T`.  I.e. in RQuantLib I let 'time to expiry' be either the `double` it always was (applied as 'time' relative to valuation date) or as a date and a number of functions now accomodate both types and then call this helper to determine the date:

```
QuantLib::Date getFutureDate(const QuantLib::Date today,
                             Rcpp::Nullable<double> maturity,
                             Rcpp::Nullable<QuantLib::Date> exDate) {
    if (exDate.isUsable()) {
        return Rcpp::as<QuantLib::Date>(exDate);
    } else if (maturity.isUsable()) {
        double mat = Rcpp::as<double>(maturity);
        // depending on the compile-time option, this is either intra-day or not
        // in minutes
        auto length = boost::posix_time::minutes(boost::uint64_t(mat * 360 * 24 * 60));
        return QuantLib::Date{today.dateTime() + length}; // high-res time ctos
    } else {
        Rcpp::stop("Excactly one of 'maturity' or 'exDate' needs to be supplied and be non-null.");
        return QuantLib::Date(); // not reached
    }
}
```

and I just realize I still need to rename `today` to `evalDate` or alike.

eddelbuettel:
@enchufa You convinced me that `operator T()` is a good idea so I added that.  But given that `as()` has been exposed for a decade I do not want to remove it overnight (also known as 'R Core-style' around here).  I would be up for sticking deprecation warnings into `as()` if we think `operator T()` suits.   As for `get()`, we could deprecate / hide at the same time.

PS Oh gnarl, may have to back that out or revisit as CI is unhappy.

Enchufa2:
That's weird. Is it due to a different C++ standard with those versions?

eddelbuettel:
I honestly cannot work it out!   But it is related to our `Rcpp::String` so it is a fair chance it is us.  I am not sure I care.  

You were right we want `operator T`.  Back in the day we couldn't, apparently.  Now we can, so now we do. Works for me.

eddelbuettel:
Ok rebased following merge and about to turn on the reverse-depends machine for the last currently-open PR.

eddelbuettel:
Ahh, sad news @Enchufa2 :  `operator T()` is not going to make it.  A regression in `gdalraster`:

```
gdalvector.cpp: In constructor ‘GDALVector::GDALVector(const Rcpp::CharacterVector&, const std::string&, bool, const Rcpp::Nullable<Rcpp::Vector<16> >&, const std::string&, const std::string&)’:
gdalvector.cpp:74:51: error: operands to ‘?:’ have different types ‘const Rcpp::Nullable<Rcpp::Vector<16> >’ and ‘Rcpp::Vector<16>’
   74 |           m_open_options(open_options.isNotNull() ?
      |                          ~~~~~~~~~~~~~~~~~~~~~~~~~^
   75 |                          open_options : Rcpp::CharacterVector::create()),
      |                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
gdalvector.cpp:74:51: note:   and each type can be converted to the other
make: *** [/usr/lib/R/etc/Makeconf:211: gdalvector.o] Error 1
ERROR: compilation failed for package ‘gdalraster’
```

If I remove the definition of the member function, it passes and builds.

I could leave it behind an opt-in `#define` with a comment.   What do you think?

eddelbuettel:
It also passes when I change `gdalraster` by adding a cast

```diff
root@25dd6436899c:/tmp/gdalraster/gdalraster# diff -u src/gdalvector.cpp.orig src/gdalvector.cpp
--- src/gdalvector.cpp.orig     2026-04-10 18:38:13.351663913 +0000
+++ src/gdalvector.cpp  2026-04-10 18:35:23.322868566 +0000
@@ -72,7 +72,7 @@
                        const std::string &dialect = "")
         : m_layer_name(layer), m_dialect(dialect),
           m_open_options(open_options.isNotNull() ?
-                         open_options : Rcpp::CharacterVector::create()),
+                         (Rcpp::CharacterVector) open_options : Rcpp::CharacterVector::create()),
           m_spatial_filter(spatial_filter),
           m_ignored_fields(Rcpp::CharacterVector::create()),
           m_hDataset(nullptr), m_eAccess(GA_ReadOnly), m_hLayer(nullptr) {
root@25dd6436899c:/tmp/gdalraster/gdalraster# 
```

This is closer to the default pattern of `Rcpp::Nullable<T>`: when not null, for the given `T` we usually instantiate as `T(obj)`.

Paging @ctoney : Hi Chirs,  I see you just released a `gdalraster` two weeks ago.  When do you plan a new release, and would you consider tossing a cast like the one I have in this comment into `src/gdalvector.cpp` ?  (Context is a not overly-important small enhancement to `Rcpp::Nullable` which we are currently testing, and with about 1/3 of packages done only `gdalraster` croaked.  No rush, we can wait at least until the reverse depends check is done.

Enchufa2:
How did that work before anyway? Shouldn't it get an error because there is no conversion from `const Rcpp::Nullable<Rcpp::Vector<16>>` to `<Rcpp::Vector<16>`?

eddelbuettel:
Don't ask me. It's on CRAN.  Maybe implicitly via `operator SEXP()` ?  C++ conversion rules are weird!

ctoney:
I was actually planning another release pretty close to that one. There's a very small performance regression if progress bar is turned on which it is by default. A few feature adds with that make a release worthwhile. I was going to wait until about the one month mark just for CRAN etiquette but can do it whenever. I'll test that change on my end shortly but seems fine. Thanks for investigating it.

eddelbuettel:
Awesome!  As for 

> I was going to wait until about the one month mark just for CRAN etiquette 

I usually wait a week to not trigger the nag and try to keep it under 'six in six months' which with RcppArmadillo and upstream's frequency had its challenges at times :)    But I don't think you have to have a month-long gap. With CRANberries I see all sorts of patterns. Next day, same day, you name it. 

As for context:  I think by adding `operator T()` we are expanding the surface a little, and something that used to work for you no longer does, but as hinted above, it may not have been the cleanest anyway.  But chapeau to using `(test ? yes : no)` for `Nullable` use -- that hadn't occured to until this week.  But it also bit us here so maybe the old school `if (obs.IsNotNull()) foo() else bar()` is cleaner :) 

In any event, we are probably not going to upload until July (unless CRAN forces us, you never know) so no rush.  But I'd rather sort this out now as otherwise I will have forgotten about it by then ....

eddelbuettel:
About half-way done and two more packages showing up with issues.  Adding `operator T()` was a nice idea to complete and standardize an existing API but it may not work.  Will review once we have all 3k reverse dependencies looked at.

eddelbuettel:
So six packages fail if add `operator T()` directly to `Nullable`.  The `gdalraster` by @ctoney was first alphabetically :)

I committed the results for the [full run](https://github.com/RcppCore/rcpp-logs/commit/d76df6f80075569d8901c896b74a5ca2684d4f14) as well as [incremental run](https://github.com/RcppCore/rcpp-logs/commit/b0f5b349ffc55fbaca96a44b947b6714043dc77c) once `Nullable::operatorT()` is hidden behind an opt-in `#define`.  We could merge with this, and I could reach out with (likely all very simple) patches to the six packages.  Once they add the explicit cast the need we can remove the opt-in `#define` / change it to an opt-out if think it adds value.

Thoughts?

PS And now that we confirmed that it addresses the issue initially observed, committed and pushed as https://github.com/RcppCore/Rcpp/pull/1471/commits/8dcc96bba3503fb463109935746a32e1f315eadf

ctoney:
re: gdalraster

>How did that work before anyway?

I'm also not sure how that worked before. It looks like it potentially would not in that case. The constructors for `GDALVector` definitely needed some cleanup. That is done in https://github.com/firelab/gdalraster/pull/948 and should address the issue revealed here. Thanks for catching it.

Enchufa2 (approved):
Ok!

PR #1484, opened 2026-06-26
fix ListOf compile by not relying on get__() in attribute accessors

### LINKED ISSUE #1483: New issue on `get__()` versus `get()`
This came up in the reverse depends check for #1482, but reproduces locally for me.  Package `RestRserve` has a `.hasAttribute()` check, and that fails on 

```
ccache g++ -I"/usr/share/R/include" -DNDEBUG  -I'/usr/local/lib/R/site-library/Rcpp/include'    -DRCPP_NO_MODULES -fpic  -g0 -O3 -Wall -pipe -Wno-unused  -fext-numeric-literals -DBOOST_NO_AUTO_PTR    -c format_\
headers.cpp -o format_headers.o
In file included from /usr/local/lib/R/site-library/Rcpp/include/Rcpp/proxy/proxy.h:24,
                 from /usr/local/lib/R/site-library/Rcpp/include/RcppCommon.h:132,
                 from /usr/local/lib/R/site-library/Rcpp/include/Rcpp.h:27,
                 from format_headers.cpp:3:
/usr/local/lib/R/site-library/Rcpp/include/Rcpp/proxy/AttributeProxy.h: In instantiation of ‘bool Rcpp::AttributeProxyPolicy<CLASS>::hasAttribute(const std::string&) const [with CLASS = Rcpp::ListOf<Rcpp::Vecto\
r<16> >; std::string = std::__cxx11::basic_string<char>]’:
format_headers.cpp:12:22:   required from here
   12 |   if (!x.hasAttribute("names")) {
      |        ~~~~~~~~~~~~~~^~~~~~~~~
/usr/local/lib/R/site-library/Rcpp/include/Rcpp/proxy/AttributeProxy.h:100:61: error: ‘const class Rcpp::ListOf<Rcpp::Vector<16> >’ has no member named ‘get__’; did you mean ‘get’?
  100 |         return R_hasAttrib(static_cast<const CLASS&>(*this).get__(), Rf_install(attr.c_str()));
      |                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
      |                            get
make: *** [/usr/lib/R/etc/Makeconf:199: format_headers.o] Error 1
ERROR: compilation failed for package ‘RestRserve’
```

It does indeed go away if we replace `get__()` with `get()` here:

https://github.com/RcppCore/Rcpp/blob/8d2f8f880830135c93bf41006e2e3f4f4c7e33a0/inst/include/Rcpp/proxy/AttributeProxy.h#L100

but as even that file has another `get__()` I am wondering if we need a more general chase for it?

kevinushey:
Fixes #1483.

### Problem

`AttributeProxyPolicy<CLASS>` is a policy base class. Its `attributeNames()` and `hasAttribute()` members fetched the underlying `SEXP` via `static_cast<const CLASS&>(*this).get__()`. But `get__()` is not part of this policy -- it is provided by the *storage* policy (`PreserveStorage` / `NoProtectStorage`). `Vector`, `XPtr` and the macro-generated classes inherit a storage policy, so they have `get__()`; `ListOf<T>` inherits `AttributeProxyPolicy` *without* a storage policy and has only `get()` / `operator SEXP()`.

As a result, e.g. `RestRserve`'s `x.hasAttribute("names")` on a `ListOf<...>` failed to compile under R >= 4.6.0:

```
'const class Rcpp::ListOf<Rcpp::Vector<16> >' has no member named 'get__'; did you mean 'get'?
```

### Fix

Drop `.get__()` and rely on the implicit `operator SEXP()` conversion that all of these classes provide. This matches the adjacent `R_getAttribNames(static_cast<const CLASS&>(*this))` call (which already worked for `ListOf`) and is equivalent to the pre-4.6.0 `.attr()` path. Both the R >= 4.6.0 `hasAttribute()` and the pre-4.6.0 `attributeNames()` spot are fixed (the latter had the same latent bug for `ListOf`).

### Tests

Added `listof_has_attr` / `listof_attribute_names` exports and corresponding cases in `test_listof.R`. Full `test_listof.R` passes (20 results).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2026-06-26  Kevin Ushey <kevinushey@gmail.com>
+
+	* inst/include/Rcpp/proxy/AttributeProxy.h: Do not rely on get__()
+	in attribute accessors so that classes such as ListOf compile
+	* inst/tinytest/cpp/ListOf.cpp: Added tests
+	* inst/tinytest/test_listof.R: Idem
+
 2026-06-18  Dirk Eddelbuettel  <edd@debian.org>
 
 	* vignettes/rmd/Rcpp.bib: Updated references
--- a/inst/NEWS.Rd
+++ b/inst/NEWS.Rd
@@ -36,6 +36,9 @@
       has been fixed (Kevin in \ghpr{1475} fixing \ghit{1474})
       \item The \code{Nullable::operatorT()} has been added as a 'opt-out'
       (Dirk in \ghpr{1477} with coordination in \ghit{1472})
+      \item The attribute accessors in \code{AttributeProxyPolicy} no longer
+      rely on \code{get__()} so that classes such as \code{ListOf} compile
+      (Kevin in \ghpr{1484} fixing \ghit{1483})
     }
     \item Changes in Rcpp Documentation:
     \itemize{
--- a/inst/include/Rcpp/proxy/AttributeProxy.h
+++ b/inst/include/Rcpp/proxy/AttributeProxy.h
@@ -86,7 +86,7 @@ public:
             v.push_back(std::string(CHAR(STRING_ELT(attrs, i))));
         }
 #else
-        SEXP attrs = ATTRIB( static_cast<const CLASS&>(*this).get__());
+        SEXP attrs = ATTRIB( static_cast<const CLASS&>(*this) );
         while( attrs != R_NilValue ){
             v.push_back( std::string(CHAR(PRINTNAME(TAG(attrs)))) ) ;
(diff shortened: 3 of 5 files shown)


eddelbuettel (approved):
Nice!

Usual micro nit:  ChangeLog entry in two lines, if we can?

Enchufa2:
In #1464 I already removed a couple of them to rely on SEXP for ListOf, so it only makes sense to do the same here.

eddelbuettel:
No (new) issues in reverse dependency checks.

PR #1490, opened 2026-08-03
protect new cons cell in named StretchyList::push_back()

### LINKED ISSUE #1489: StretchyList::push_back() can corrupt memory: cons cell unprotected across Rf_install()
In the named-object variant of `StretchyList::push_back()`, the freshly allocated cons cell is neither protected nor reachable when the `Symbol` constructor runs `Rf_install()`:

https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/api/meat/StretchyList.h#L36-L45

```cpp
StretchyList_Impl<StoragePolicy>& StretchyList_Impl<StoragePolicy>::push_back__impl( const T& obj, traits::true_type ){
    Shield<SEXP> s( wrap(obj.object) ) ;
    SEXP tmp  = Rf_cons( s, R_NilValue );
    Symbol tag  = obj.name ;   // Rf_install() -- can allocate and trigger GC
    SET_TAG(tmp, tag) ;
    SEXP self = Storage::get__() ;
    SETCDR( CAR(self), tmp) ;
    SETCAR( self, tmp ) ;
    return *this ;
}
```

If the tag name is not yet interned, `Rf_install()` allocates (`mkChar()` plus the SYMSXP), which can trigger a garbage collection. At that point `tmp` is only referenced from the C stack, so it is collected; `SET_TAG()` then writes into a reclaimed node, and `SETCDR()` / `SETCAR()` link that reclaimed node into the (preserved) list.

Note that the named variant of `push_front()` in the same file gets the ordering right, constructing the `Symbol` *before* calling `Rf_cons()`, so only `push_back()` is affected.

## Reproducible example

```r
writeLines('
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
SEXP stretchy_push_back(std::string name) {
    StretchyList out;
    out.push_back(Named(name, 42));
    return out;
}
', "repro.cpp")

Rcpp::sourceCpp("repro.cpp")

# a symbol name that has never been interned in this session, so that
# Rf_install() inside push_back() has to allocate
name <- paste(sample(c(letters, LETTERS), 32, TRUE), collapse = "")

gctorture(TRUE)
x <- stretchy_push_back(name)
gctorture(FALSE)

cat("typeof:", typeof(x), "\n")
str(x)
```

Output:

```
typeof: symbol 
 symbol oZoNOOQgnEmozfyTBxxQsXnamjjNqEYY
```

Under `gctorture()`, the GC triggered inside `Rf_install()` collects the cons cell, and its node is immediately reused for the newly created SYMSXP -- so the "pairlist" we get back is actually the symbol itself. Expected output would be a one-element pairlist, e.g.:

```
typeof: pairlist 
Dotted pair list of 1
 $ <name>: num 42
```

Depending on allocation timing this can also surface as corruption of unrelated objects (the reclaimed node can be reused for anything), so the symptom in real code would typically be a segfault or silently corrupted data rather than this tidy misclassification.

Observed with Rcpp master (3a9aeaaa) on R 4.6.1.

A PR with a fix (reordering `push_back__impl()` to construct the `Symbol` before `Rf_cons()`, matching `push_front__impl()`) will follow shortly.

kevinushey:
Fixes #1489.

The named variant of `StretchyList::push_back()` allocated the new cons cell before constructing the `Symbol` for its tag. If the tag name was not yet interned, `Rf_install()` could allocate and trigger a garbage collection that reclaimed the still-unreachable cell; `SET_TAG()` then wrote into a reclaimed node, which was subsequently linked into the list. See #1489 for a reproducible example via `Rcpp::sourceCpp()` and `gctorture()`.

The fix constructs the `Symbol` before calling `Rf_cons()`, matching the ordering already used by the named variant of `push_front()`.

Also adds a regression test exercising named `push_back()` with a freshly generated (never-interned) tag name under `gctorture()`; the test fails without the header change and passes with it.

#### Checklist

- [x] Code compiles correctly
- [x] `R CMD check` still passes all tests (checked locally with `RunAllRcppTests=yes` on R 4.6.1; `--no-manual --no-vignettes`, so the only flagged items were two vignette-packaging warnings from the local `--no-build-vignettes` build)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,12 @@
 	* inst/include/Rcpp/proxy/NamesProxy.h: Idem
 	* inst/include/Rcpp/proxy/SlotProxy.h: Idem
 	* inst/tinytest/cpp/misc.cpp: Add regression tests
+
+2026-08-02  Kevin Ushey <kevinushey@gmail.com>
+
+	* inst/include/Rcpp/api/meat/StretchyList.h: Protect the new cell
+	from collection during Rf_install() in named push_back() (#1489)
+	* inst/tinytest/cpp/misc.cpp: Add regression test
 	* inst/tinytest/test_misc.R: Idem
 
 2026-07-24  Dirk Eddelbuettel  <edd@debian.org>
--- a/inst/include/Rcpp/api/meat/StretchyList.h
+++ b/inst/include/Rcpp/api/meat/StretchyList.h
@@ -35,8 +35,8 @@ namespace Rcpp{
     template< typename T>
     StretchyList_Impl<StoragePolicy>& StretchyList_Impl<StoragePolicy>::push_back__impl( const T& obj, traits::true_type ){
         Shield<SEXP> s( wrap(obj.object) ) ;
-        SEXP tmp  = Rf_cons( s, R_NilValue );
         Symbol tag  = obj.name ;
+        SEXP tmp  = Rf_cons( s, R_NilValue );
         SET_TAG(tmp, tag) ;
         SEXP self = Storage::get__() ;
         SETCDR( CAR(self), tmp) ;
--- a/inst/tinytest/cpp/misc.cpp
+++ b/inst/tinytest/cpp/misc.cpp
@@ -174,6 +174,13 @@ StretchyList named_stretchy_list() {
 }
 
 // [[Rcpp::export]]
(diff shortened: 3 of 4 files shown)


eddelbuettel (approved):
LGTM

PR #1492, opened 2026-08-03
protect fresh proxy get() results in conversion and assignment operators

### LINKED ISSUE #1491: proxy conversion and assignment operators use unprotected fresh SEXPs
The proxy classes share a pattern that passes freshly returned, unprotected SEXPs across allocating calls. One member of the family is reproducible today through the public API on a protect-checking build of R-devel; the rest have the same shape and were genuinely exposed on R versions Rcpp still supports (details below).

## Reproducible example

Requires an R built with structure checking, e.g. `--enable-strict-barrier` (the configuration used for CRAN's ASan checks; tested against an ASan/UBSan build of R-devel r90339). On standard builds the failure is silent and timing-dependent.

```r
writeLines('
#include <Rcpp.h>
using namespace Rcpp;

// [[Rcpp::export]]
void copy_field(Reference a, Reference b) {
    a.field("x") = b.field("y");
}
', "repro.cpp")
Rcpp::sourceCpp("repro.cpp")

## an active field, so b$y computes a fresh value on every read
Foo <- setRefClass("Foo", fields = list(
    x = "ANY",
    y = function(value) {
        if (missing(value)) new.env(parent = emptyenv()) else stop("read-only")
    }
))

a <- Foo$new(x = NULL)
b <- Foo$new(x = NULL)

gctorture(TRUE)
for (i in 1:200)
    copy_field(a, b)
gctorture(FALSE)
```

Output:

```
Error in copy_field(a, b) :
  unprotected object (0x6250035f0238) encountered (was ENVSXP)
```

R's protect checking is reporting that the value read from `b$y` was garbage-collected and then used.

## Mechanism (the reproduced case)

`FieldProxy::operator=(const FieldProxy&)` (inst/include/Rcpp/api/meat/proxy.h:179) does `set(rhs.get())`. `rhs.get()` returns the result of an `Rcpp_fast_eval` of `b$y` -- referenced from nowhere the GC can see -- and `set()` (inst/include/Rcpp/proxy/FieldProxy.h:48-53) then runs `Rf_mkString(field_name)` before the value is captured by `Rf_lang4`:

```cpp
void set(SEXP x ) {
    SEXP dollarGetsSym = Rf_install( "$<-");
    Shield<SEXP> str(Rf_mkString(field_name.c_str()));   // <- allocates; x unprotected
    Shield<SEXP> call(Rf_lang4(dollarGetsSym, parent, str, x));
    parent.set__( Rcpp_fast_eval( call, R_GlobalEnv ) );
}
```

A GC triggered by that allocation collects `x`; the subsequent `$<-` call assigns a freed object into the field.

## The wider family

The same shape -- fresh `get()` result handed to something that allocates before securing it -- appears in:

- All 12 `operator T()` conversions in inst/include/Rcpp/api/meat/proxy.h (`AttributeProxy`, `NamesProxy`, `SlotProxy`, `TagProxy`, `Binding`, `DottedPairProxy`, `FieldProxy` and their const variants), plus `const_SlotProxy::operator T()` (proxy/SlotProxy.h:68) and the attribute proxy in vector/proxy.h:173. These do `as<T>(get())`; for coercing targets, `r_cast` reaches `Rf_coerceVector`, which allocates the destination. Fresh-`get()` producers include active bindings through `Environment::get()`, `Rf_getAttrib()` on compact row names, and field access through `Rcpp_fast_eval`.
- The proxy-to-proxy assignments `Binding::operator=(const Binding&)` (proxy/Binding.h:49, i.e. `envA["x"] = envB["y"]`), `SlotProxy::operator=(const SlotProxy&)` (proxy/SlotProxy.h:36) and `NamesProxy::operator=(const NamesProxy&)` (proxy/NamesProxy.h:34), whose `set()` implementations reach allocating calls (`Rf_defineVar`, `R_do_slot_assign`) with the fresh value unprotected.

## Why standard builds rarely show this

Three unrelated internals currently mask the windows on recent R:

1. `coerceVector()` protects its argument internally -- but only since **R 4.1.0** (`PROTECT(v)` before the dispatch switch; absent in 3.5.0 through 4.0.5, verified against the R sources for 3.5.0, 3.6.0, 4.0.0, 4.0.5, 4.1.0). Rcpp supports R >= 3.5.0, so on 3.5.0-4.0.5 the `operator T()` coercion chains (e.g. `double x = env["count"]` where the binding is active, or `std::vector<double> rn = df.attr("row.names")`) are genuinely exposed.
2. `Rf_cons`/`Rf_lang*` protect their arguments, which covers several construction sequences.
3. `R_UnwindProtect()` stores the evaluation result into the unwind continuation token (`SETCAR(cont, result)`, src/main/context.c). Rcpp's token is protected for the duration of the eval, so it is old-generation by the time it is dropped, and it accidentally keeps every `Rcpp_fast_eval()` result reachable until the next *full* collection. This is undocumented behavior of R internals, not a guarantee, and it is what makes the windows above hard to hit on standard builds: they require a full GC landing inside a one-or-two-allocation window.

None of these is a documented contract; the documented convention is that the caller protects. The fix is mechanical: `Shield<SEXP>` the `get()` result in the conversion operators and in the proxy-to-proxy assignments. A PR will follow.

kevinushey:
Fixes #1491.

The proxy `operator T()` conversions and the proxy-to-proxy assignment operators passed the raw result of `get()` into allocating calls: `as<T>()` coercions on the conversion side, and `set()` implementations that allocate before capturing the value on the assignment side. When `get()` returns a freshly computed SEXP (an active binding or active RefClass field, a field access evaluated via `$`, an attribute expanded by `Rf_getAttrib()`), that object is referenced from nowhere the GC can see, and a collection inside the window frees it while still in use. See #1491 for the analysis, including why standard builds of recent R rarely surface this (internal protection in `coerceVector()` since R 4.1.0, and `R_UnwindProtect()` accidentally keeping eval results reachable via the continuation token).

Changes:

- `api/meat/proxy.h`: `Shield` the `get()` result in all thirteen `operator T()` conversions and in `FieldProxy::operator=(const FieldProxy&)`.
- `proxy/SlotProxy.h`: same for `const_SlotProxy::operator T()` (defined inline) and `SlotProxy::operator=(const SlotProxy&)`.
- `proxy/Binding.h`, `proxy/NamesProxy.h`: same for the proxy-to-proxy assignments.
- `Binding::operator=(const Binding&)` additionally had a latent compile error: its self-assignment guard `*this != rhs` is ambiguous (the template conversion operators make it match unrelated `!=` overloads), so the operator failed to compile whenever it was actually instantiated. The guard now compares the environment and name directly; the new regression test is this operator's first instantiation.
- Reviewed but intentionally not changed: `AttributeProxy::operator=(const AttributeProxy&)` (its `set()` shields immediately upon entry) and `generic_name_proxy::operator T()` in `vector/proxy.h` (its `get()` returns a parent-reachable element).

Verification:

- On an ASan/UBSan build of R-devel (r90339) with structure checking, the repro from #1491 fails with `unprotected object (0x...) encountered (was ENVSXP)` against current master and runs clean with this patch.
- New regression tests exercise `FieldProxy` and `Binding` copies from actively-computed values under `gctorture()`; they pass on standard builds and guard the fix on protect-checking builds such as CRAN's ASan setup.

#### Checklist

- [x] Code compiles correctly
- [x] `R CMD check` still passes all tests (checked locally with `RunAllRcppTests=yes` on R 4.6.1; `--no-manual --no-vignettes`, so the only flagged items were two vignette-packaging warnings from the local `--no-build-vignettes` build)

### DIFF
--- a/inst/include/Rcpp/proxy/Binding.h
+++ b/inst/include/Rcpp/proxy/Binding.h
@@ -45,8 +45,12 @@ public:
             env.unlockBinding(name) ;
         }
         Binding& operator=(const Binding& rhs){
-            if( *this != rhs )
-                set( rhs.get() ) ;
+            // NB: '*this != rhs' previously used here would not compile
+            // when this operator was instantiated
+            if( env.get__() != rhs.env.get__() || name != rhs.name ) {
+                Shield<SEXP> x( rhs.get() ) ;
+                set(x) ;
+            }
             return *this ;
         }
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2026-08-03  Kevin Ushey <kevinushey@gmail.com>
+
+	* inst/include/Rcpp/api/meat/proxy.h: Protect fresh SEXPs returned
+	from proxy get() in conversion and assignment operators (#1491)
+	* inst/include/Rcpp/proxy/Binding.h: Idem
+	* inst/include/Rcpp/proxy/NamesProxy.h: Idem
+	* inst/include/Rcpp/proxy/SlotProxy.h: Idem
+	* inst/tinytest/cpp/misc.cpp: Add regression tests
+	* inst/tinytest/test_misc.R: Idem
+
 2026-07-24  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
--- a/inst/include/Rcpp/api/meat/proxy.h
+++ b/inst/include/Rcpp/api/meat/proxy.h
@@ -37,7 +37,8 @@ AttributeProxyPolicy<CLASS>::AttributeProxy::operator=(const T& rhs) {
 template <typename CLASS>
 template <typename T>
 AttributeProxyPolicy<CLASS>::AttributeProxy::operator T() const {
-    return as<T>(get());
+    Shield<SEXP> x(get());
+    return as<T>(x);
 }
 
 template <typename CLASS>
@@ -48,7 +49,8 @@ AttributeProxyPolicy<CLASS>::AttributeProxy::operator SEXP() const {
 template <typename CLASS>
 template <typename T>
 AttributeProxyPolicy<CLASS>::const_AttributeProxy::operator T() const {
-    return as<T>(get());
+    Shield<SEXP> x(get());
+    return as<T>(x);
 }
 
 template <typename CLASS>
@@ -68,13 +70,15 @@ NamesProxyPolicy<CLASS>::NamesProxy::operator=(const T& rhs) {
 template <typename CLASS>
 template <typename T>
 NamesProxyPolicy<CLASS>::NamesProxy::operator T() const {
-    return as<T>( get() );
+    Shield<SEXP> x(get());
+    return as<T>(x);
 }
 
 template <typename CLASS>
 template <typename T>
 NamesProxyPolicy<CLASS>::const_NamesProxy::operator T() const {
-    return as<T>( get() );
+    Shield<SEXP> x(get());
+    return as<T>(x);
 }
 
 // SlotProxy
@@ -89,7 +93,8 @@ SlotProxyPolicy<CLASS>::SlotProxy::operator=(const T& rhs) {
 template <typename CLASS>
 template <typename T>
 SlotProxyPolicy<CLASS>::SlotProxy::operator T() const {
-    return as<T>(get());
+    Shield<SEXP> x(get());
+    return as<T>(x);
 }
 
 // TagProxy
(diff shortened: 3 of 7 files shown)


inst/include/Rcpp/proxy/Binding.h
```diff
@@ -45,8 +45,12 @@ class BindingPolicy {
             env.unlockBinding(name) ;
         }
         Binding& operator=(const Binding& rhs){
-            if( *this != rhs )
-                set( rhs.get() ) ;
+            // NB: '*this != rhs' previously used here would not compile
+            // when this operator was instantiated
```
eddelbuettel, line 49: Odd.  Thanks for persisting here too.

eddelbuettel (approved):
Thank you!  The actual change is, as you said, somewhat mechanical.  

It's good to get to these.

PR #1494, opened 2026-08-03
protect module method and function results while wrapping in invoke

### LINKED ISSUE #1493: module method results can be garbage collected during result wrapping in class_::invoke()
In Rcpp modules, `class_<T>::invoke()` wraps a freshly computed method result like this (`inst/include/Rcpp/module/class.h`):

```cpp
return Rcpp::List::create( false, m->operator()( XP(object), args ) ) ;
```

`m->operator()()` returns a raw `SEXP` -- the wrapped result of the user's C++ method -- and nothing the garbage collector can see references it. `List::create()` then performs at least two allocations (the `VECSXP` for the result list, and `Rf_ScalarLogical()` for the voidness flag) before the result is stored in the (protected) list. A garbage collection inside that window frees the method result while it is still in use.

`Module::invoke()` has the same shape for module *functions* (`inst/include/Rcpp/module/Module.h`):

```cpp
return List::create(
    _["result"] = fun->operator()( args ),
    _["void"]   = fun->is_void()
) ;
```

Here the window is wider still: the result list, the voidness flag, and the `names` attribute (a `STRSXP` plus two `CHARSXP`s) are all allocated while `fun->operator()( args )`'s result is unprotected.

### Reachability

`class_::invoke()` is reached from ordinary R code whenever a module class method name has overloads with *mixed voidness* (some overloads `void`, some not): `method_wrapper()` in `R/Module.R` only routes through `CppMethod__invoke` -- and therefore `class_::invoke()` -- in that case, so that voidness can be resolved at call time. Method names whose overloads are all non-void dispatch through `CppMethod__invoke_notvoid`, which returns the result directly and is not affected.

`Module__invoke` (the entry point for `Module::invoke()`) is registered in `rcpp_init.cpp`, though Rcpp's own R code routes module functions through `InternalFunction_invoke`, which is not affected.

### Why this usually goes unnoticed

When a method returns an Rcpp object (say a `NumericVector`), the result is kept on Rcpp's precious list until the method wrapper's temporaries are destroyed. Garbage collections that run while it is still protected mark the object and promote it out of generation 0, after which the (mostly level-0) collections inside the `List::create()` window usually skip it -- the same generational masking described in #1491. The window only bites when a deeper collection lands inside it, so on standard R builds the use-after-free is both rare and silent (the freed node's contents typically survive until the node is reused). A protect-checking build of R-devel detects it reliably. Methods returning a raw, freshly allocated `SEXP` are more exposed still: nothing protects the result at any point, so it is generation 0 and completely unreferenced when `List::create()` allocates.

### Reproducible example

`repro.cpp`:

```cpp
#include <Rcpp.h>

using namespace Rcpp;

class Gadget {
public:
    Gadget() {}

    // void overload: its presence forces dispatch through class_::invoke,
    // which wraps results as List::create(false, <method result>)
    void value(int x) {
        (void) x;
    }

    // non-void overload: returns a freshly allocated vector
    NumericVector value() {
        return NumericVector::create(1.0, 2.0, 3.0);
    }
};

RCPP_MODULE(gadget_module) {
    class_<Gadget>("Gadget")
        .constructor()
        .method("value", static_cast<void (Gadget::*)(int)>(&Gadget::value))
        .method("value", static_cast<NumericVector (Gadget::*)()>(&Gadget::value));
}
```

`repro.R`:

```r
Rcpp::sourceCpp("repro.cpp")

g <- methods::new(Gadget)

stopifnot(identical(g$value(), c(1, 2, 3)))

gctorture(TRUE)
hit <- 0L
bad <- NULL
for (i in 1:100) {
    x <- g$value()
    if (!identical(x, c(1, 2, 3))) {
        hit <- i
        bad <- x
        break
    }
}
gctorture(FALSE)

if (hit > 0L) {
    cat("iteration", hit, "- corrupted method result!\n")
    cat("typeof:", typeof(bad), "(expected: double)\n")
    print(bad)
} else {
    cat("no corruption in 100 iterations\n")
}
```

On a protect-checking ASan/UBSan build of R-devel (r90339), this fails within the loop -- the freed `REALSXP` is the method result, caught when `dealWith()` in `R/Module.R` touches the result list:

```
Error in isTRUE(x[[1]]) : 
  unprotected object (0x625003d726a8) encountered (was REALSXP)
Calls: <Anonymous> -> <Anonymous> -> isTRUE -> .External
Execution halted
```

On a standard build of R 4.6.1 the loop runs "clean" (the freed node's contents happen to survive), which is exactly why this class of bug hides so well in practice.

I'll follow up with a PR that protects the freshly computed result in both `class_::invoke()` and `Module::invoke()`.

kevinushey:
Fixes #1493.

In Rcpp modules, `class_<T>::invoke()` passed the freshly computed method result straight into `Rcpp::List::create()`, which allocates the result list and the voidness flag before the result is stored anywhere the garbage collector can see. A collection inside that window frees the result while it is still in use. `Module::invoke()` had the same shape for module functions, with a wider window (the result list's names are allocated too). See #1493 for the analysis, including why results returned as Rcpp objects are usually (but not reliably) masked by generational promotion.

Changes:

- `module/class.h`: `Shield` the method result in `class_::invoke()` before wrapping it in the result list.
- `module/Module.h`: same for the function result in `Module::invoke()`.
- `inst/tinytest/cpp/Module.cpp`, `inst/tinytest/test_module.R`: regression test exercising a method name with mixed-voidness overloads (the dispatch path that reaches `class_::invoke()`, previously uncovered) under `gctorture()`.

Verification:

- On an ASan/UBSan build of R-devel (r90339) with structure checking, the repro from #1493 fails with `unprotected object (0x...) encountered (was REALSXP)` against current master, and runs clean with this patch.
- The new regression test passes on standard builds and guards the fix on protect-checking builds such as CRAN's ASan setup.

#### Checklist

- [x] Code compiles correctly
- [x] `R CMD check` still passes all tests (checked locally with `RunAllRcppTests=yes` on R 4.6.1; `--no-manual --no-vignettes`, so the only flagged items were the two vignette-packaging warnings from the local `--no-build-vignettes` build)
- [x] Preferably, new tests were added which fail without the change (fails on protect-checking builds of R; on standard builds the use-after-free is silent)

### DIFF
--- a/inst/include/Rcpp/module/class.h
+++ b/inst/include/Rcpp/module/class.h
@@ -193,7 +193,8 @@
                 m->operator()( XP(object), args );
                 return Rcpp::List::create( true ) ;
             } else {
-                return Rcpp::List::create( false, m->operator()( XP(object), args ) ) ;
+                Shield<SEXP> res( m->operator()( XP(object), args ) ) ;
+                return Rcpp::List::create( false, static_cast<SEXP>(res) ) ;
             }
             END_RCPP
                 }
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2026-08-03  Kevin Ushey  <kevinushey@gmail.com>
+
+	* inst/include/Rcpp/module/class.h (invoke): Protect freshly
+	computed method results while wrapping them in the result list
+	(#1493)
+	* inst/include/Rcpp/module/Module.h (invoke): Idem for module
+	function results
+	* inst/tinytest/cpp/Module.cpp: Add regression test
+	* inst/tinytest/test_module.R: Idem
+
 2026-07-24  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
--- a/inst/include/Rcpp/module/Module.h
+++ b/inst/include/Rcpp/module/Module.h
@@ -60,8 +60,9 @@ namespace Rcpp {
                 throw std::range_error( "incorrect number of arguments" ) ;
             }
 
+            Shield<SEXP> res( fun->operator()( args ) ) ;
             return List::create(
-                _["result"] = fun->operator()( args ),
+                _["result"] = static_cast<SEXP>(res),
                 _["void"]   = fun->is_void()
             ) ;
         }
--- a/inst/tinytest/cpp/Module.cpp
+++ b/inst/tinytest/cpp/Module.cpp
@@ -136,6 +136,27 @@ double Test_get_x_pointer(ModuleTest* x) {
     return x->value;
 }
 
+class ModuleGadget {
+public:
+    ModuleGadget() {}
+
+    // void overload: forces dispatch through class_::invoke(), which
+    // wraps method results as list(voidness, result)
+    void value(int x) {
+        (void) x;
+    }
+
+    // non-void overload: nothing protects the raw SEXP result while
+    // class_::invoke() allocates the result list
+    SEXP value() {
+        SEXP x = Rf_allocVector(REALSXP, 3);
+        REAL(x)[0] = 1;
+        REAL(x)[1] = 2;
+        REAL(x)[2] = 3;
+        return x;
+    }
+};
+
 RCPP_MODULE(demoModule) {
     function("hello", &hello);
     function("bar"  , &bar  );
@@ -193,6 +214,12 @@ RCPP_MODULE(demoModule) {
         .constructor<double,double>()
 
         .method("get" , &ModuleRandomizer::get)
+        ;
+
+    class_<ModuleGadget>("ModuleGadget")
+        .constructor()
+        .method("value", static_cast<void (ModuleGadget::*)(int)>(&ModuleGadget::value))
+        .method("value", static_cast<SEXP (ModuleGadget::*)()>(&ModuleGadget::value))
         ;
 }
 
--- a/inst/tinytest/test_module.R
+++ b/inst/tinytest/test_module.R
@@ -109,15 +109,11 @@ expect_equal( test_const( seq(0,10) ), 11L )
 
 ## mixed-voidness method overloads dispatch through class_::invoke(),
 ## which must protect the freshly allocated method result while it
-## wraps it in the result list (#1493)
+## wraps it in the result list (#1493); under gctorture every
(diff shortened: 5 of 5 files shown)


inst/include/Rcpp/module/class.h
```diff
@@ -193,7 +193,8 @@
                 m->operator()( XP(object), args );
                 return Rcpp::List::create( true ) ;
             } else {
-                return Rcpp::List::create( false, m->operator()( XP(object), args ) ) ;
+                Shield<SEXP> res( m->operator()( XP(object), args ) ) ;
+                return Rcpp::List::create( false, static_cast<SEXP>(res) ) ;
```
eddelbuettel, line 197: Do we cast the result of `Shield<SEXP>` back to `SEXP` in other spots?

eddelbuettel (approved):
Looks good too, left one question inline.

inst/include/Rcpp/module/class.h
kevinushey: Only one other spot: `Environment::assign()` has a dedicated `Shield<SEXP>` overload casting back with `(SEXP)` (`Environment.h:193`, added in 8670bb69) -- required there because the templated `assign(const WRAPPABLE&)` out-matches the `SEXP` overload for a `Shield` argument (deduction ignores `operator SEXP()`) and its implementation would otherwise recurse. Everywhere else `Shield` feeds plain `SEXP`-taking C functions, where the implicit conversion applies and no cast is needed.

Strictly, the cast here isn't required: `List::create()` takes `const T&` and `wrap()` has a convertible-to-SEXP fallback, so the uncast form compiles and behaves correctly (verified, including under gctorture) -- `Vector::create()` itself relies on that path when assigning its `Shield`ed names via `res.attr("names")`. But with the cast, `_["result"]` instantiates the `named_object<SEXP>` specialization rather than a generic `named_object<Shield<SEXP>>`, keeping this on the standard exact-`SEXP` paths, so I'd prefer to keep it.

I also checked whether other module-code sites need the same protect-then-cast treatment: `invoke_notvoid()` and `getProperty()` return the raw result directly with no allocation afterward, and `Module::get_function()`'s `create()` arguments are all rooted (`XPtr` via the preserve list, `formals` is a stored `Rcpp::List`), so these two sites appear to be the only ones with the unprotected window.

PR #1496, opened 2026-08-03
convert c++ exceptions to r errors in module .External entry points

### LINKED ISSUE #1495: calling a method on an uninitialized module object aborts the R session
For a module class without a default constructor, calling `new(Class)` with no arguments deliberately creates an "uninitialized" object: the generated `initialize` method falls back to `cpp_object_dummy()` (`R/Module.R`), which stores a dummy pointer and notes

```r
## like initializer but a dummy for the case of no default
## constructor.  Will throw an error if the object is used.
```

The C++ side implements that contract in `src/module.cpp`:

```c
#define CHECK_DUMMY_OBJ(p) if (p == rcpp_dummy_pointer) throw Rcpp::not_initialized()
```

But the three method-invocation entry points that use it -- `CppMethod__invoke`, `CppMethod__invoke_void`, and `CppMethod__invoke_notvoid` -- are plain `.External` functions with no `BEGIN_RCPP`/`END_RCPP`. The `Rcpp::not_initialized` exception escapes the entry point into R's C evaluator, where there is no handler, and the process calls `std::terminate()`: the entire R session aborts. `tryCatch()` on the R side never gets a chance. The intended behavior is clearly a regular R error ("C++ object not initialized. (Missing default constructor?)").

The `XPtr` constructor validation throws (`not_compatible`, null-pointer checks) in the same functions and in `class__newInstance` can escape the same way, though those require calling the internal entry points with malformed arguments; the dummy-object path is reachable from ordinary, documented usage.

### Reproducible example

`repro.cpp`:

```cpp
#include <Rcpp.h>

using namespace Rcpp;

class Widget {
public:
    Widget(double v_) : v(v_) {}

    double value() {
        return v;
    }

private:
    double v;
};

RCPP_MODULE(widget_module) {
    class_<Widget>("Widget")
        .constructor<double>()
        .method("value", &Widget::value);
}
```

`repro.R`:

```r
Rcpp::sourceCpp("repro.cpp")

# Widget has no default constructor, so new(Widget) with no arguments
# takes the documented cpp_object_dummy() path
w <- methods::new(Widget)
cat("uninitialized object created\n")

# intended: an R error, "C++ object not initialized. (Missing default
# constructor?)"; actual: the C++ exception escapes .External uncaught
res <- tryCatch(w$value(), error = function(e) conditionMessage(e))
cat("caught R error:", res, "\n")
```

Output with R 4.6.1 (macOS, Rcpp 1.1.2 -- `src/module.cpp` is unchanged on current master):

```
uninitialized object created
libc++abi: terminating due to uncaught exception of type Rcpp::not_initialized: C++ object not initialized. (Missing default constructor?)
Abort trap: 6
```

`Rscript` exits with code 134 (`SIGABRT`); the final `cat()` is never reached.

I'll follow up with a PR that wraps these entry points in `BEGIN_RCPP`/`END_RCPP`, so the exception is converted to the intended R error. (This also lets the `XPtr` locals in those frames destruct properly on the error path instead of being skipped by the exception's escape.)

kevinushey:
Fixes #1495.

`CppMethod__invoke`, `CppMethod__invoke_void`, `CppMethod__invoke_notvoid`, and `class__newInstance` in `src/module.cpp` are `.External` entry points that can throw C++ exceptions -- most notably `CHECK_DUMMY_OBJ`'s `Rcpp::not_initialized`, the designed error for using a module object created via the dummy-object path (`new(Class)` with no arguments, for a class without a default constructor). With no `BEGIN_RCPP`/`END_RCPP` in these functions, the exception escapes into R's C evaluator uncaught and `std::terminate()` aborts the whole R session, instead of raising the intended R error. See #1495 for details.

Changes:

- `src/module.cpp`: wrap the four entry points in `BEGIN_RCPP`/`END_RCPP`, so exceptions are converted to R errors as elsewhere in Rcpp. On the `not_initialized` path this also unwinds the local `XPtr` objects properly before the R error is raised.
- `inst/tinytest/test_module.R`: regression test -- calling a method on an uninitialized `ModuleRandomizer` must raise the "C++ object not initialized" R error. Without the fix this test aborts the tinytest process.

One pre-existing quirk surfaced while testing (left unchanged, noting it here): when a dummy object is garbage collected, the reference class finalizer calls `class_<T>::run_finalizer()`, whose `XP(object)` rejects the dummy pointer (an environment), printing `Error: Expecting an external pointer: [type=environment].` at shutdown. That happens on current master for any dummy object whose finalizer runs and is independent of this change; it could be addressed separately by having `run_finalizer()` (or `CppObject__finalize`) skip the dummy instance.

Verification:

- The repro from #1495 aborts (`SIGABRT`, exit code 134) against current master and prints `caught R error: C++ object not initialized. (Missing default constructor?)` with exit code 0 against a build of this branch.
- `R CMD check` with `RunAllRcppTests=yes` passes on R 4.6.1.

#### Checklist

- [x] Code compiles correctly
- [x] `R CMD check` still passes all tests (checked locally with `RunAllRcppTests=yes` on R 4.6.1; `--no-manual --no-vignettes`, so the only flagged items were the two vignette-packaging warnings from the local `--no-build-vignettes` build)
- [x] Preferably, new tests were added which fail without the change (without the fix, the new test terminates the test process)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2026-08-03  Kevin Ushey  <kevinushey@gmail.com>
+
+	* src/module.cpp (class__newInstance, CppMethod__invoke,
+	CppMethod__invoke_void, CppMethod__invoke_notvoid): Convert C++
+	exceptions to R errors at these .External entry points, so that
+	e.g. calling a method on an uninitialized module object raises
+	the intended R error rather than terminating the R session
+	(#1495)
+	* inst/tinytest/test_module.R: Add regression test
+
 2026-07-24  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll micro version and date
--- a/inst/tinytest/test_module.R
+++ b/inst/tinytest/test_module.R
@@ -102,6 +102,12 @@ x10 <- runif(10, 10.0, 20.0)
 set.seed(123)
 expect_equal(r$get(10), x10)
 
+## calling a method on an uninitialized module object (created via the
+## dummy-object path for classes without a default constructor) should
+## raise an R error rather than terminating the R session (#1495)
+r <- new( ModuleRandomizer )
+expect_error( r$get(10L), "not initialized" )
+
 #    test.Module.flexible.semantics <- function( ){
 expect_equal( test_reference( seq(0,10) ), 11L )
 expect_equal( test_const_reference( seq(0,10) ), 11L )
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -135,12 +135,14 @@ END_RCPP
 }														// #nocov end
 
 SEXP class__newInstance(SEXP args) {
+BEGIN_RCPP
     SEXP p = CDR(args);
 
     XP_Module module(CAR(p)); p = CDR(p);
     XP_Class clazz(CAR(p)); p = CDR(p);
     UNPACK_EXTERNAL_ARGS(cargs,p)
     return clazz->newInstance(cargs, nargs);
+END_RCPP
 }
 
 // relies on being set in .onLoad()
@@ -164,6 +166,7 @@ SEXP class__dummyInstance(SEXP args) {
 }
 
 SEXP CppMethod__invoke(SEXP args) {						// #nocov start
+BEGIN_RCPP
     SEXP p = CDR(args);
 
     // the external pointer to the class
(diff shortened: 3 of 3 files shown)


eddelbuettel (approved):
LGTM.  

Interesting that 'merely' wrapping the standard entry / exit of generated functions does the trick.

Issue #1497, opened 2026-09-01
Aiming at a more strict set of compilation flags


Enchufa2:
As reported in #1485, some environments may activate `-Wextra`, which uncovers a number of warnings, mainly about unused variables. Apart from that, CRAN is constantly evolving and strengthening the checks, as e.g. we recently learned with `-Wconversion` in #1488. So we could benefit from a CI setup that enables us to expand the set of flags to detect, and then fix, those warnings.

Package installation only uses a limited part of the codebase, but the tests have a good coverage and should be representative of what client packages experience. Therefore the proposal would be to turn on `-Werror`, which will make the tests fail if any warning is found, and then start pushing and fixing `-Wextra` and other variants.

I would do this in just one of the CI jobs. The question is whether a new one, or one of the existing ones.

PR #1501, opened 2026-09-01
Add -Werror to CXXFLAGS for dev CI job

Enchufa2:
As a first step towards #1497, this simply converts warnings into errors for the checks with R devel, which I thought it was the most appropriate one. All checks still pass because I didn't activate any extra warnings yet.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,7 +52,6 @@ jobs:
           - name: dev
             cntr: rcpp/ci-dev
             r: RD
-            cxxflags: -Werror
 
     steps:
       - uses: actions/checkout@v6
@@ -62,9 +61,6 @@ jobs:
 
       - name: Build
         run: ${{matrix.r }} CMD build --no-build-vignettes --no-manual .
-
-      - name: Set flags
-        run: mkdir -p ~/.R && echo "CXXFLAGS += ${{ matrix.cxxflags }}" >> ~/.R/Makevars
 
       - name: Check
         run: CI=true ${{ matrix.r }} CMD check --no-vignettes --no-manual Rcpp_*.tar.gz
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,6 +52,7 @@ jobs:
           - name: dev
             cntr: rcpp/ci-dev
             r: RD
+            cxxflags: -Werror
 
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


eddelbuettel (approved):
A good start

Enchufa2:
The only problem with this setup is that, when tests fail due to warnings turned into compilation errors, these issues are not shown in the runner logs, but in R's own install or check logs. So we could add an additional step after the checks that, when these fail, either

- cats the relevant output; or
- uploads the `Rcpp.Rcheck` folder as an artifact for further inspection.

I personally like the second option.

eddelbuettel:
Had we discussed this before you altered an existing file :laughing: I might have convinced you to start with a new one :laughing:.  If we start from the plain `r-ci` template [r-ci.yaml](https://github.com/eddelbuettel/r-ci/blob/master/docs/r-ci.yaml) you note that it already has a part to dump logs in case of `failure()`:

https://github.com/eddelbuettel/r-ci/blob/4316c2c3cd96eb457ab2ca4e04abe52146c9e61c/docs/r-ci.yaml#L36-L38

Adding an artifact is good, and easy to add in a one-off file.

But as always there are more than one way to skin a cat and you are in the driver's seat, so other ideas welcome too.

Enchufa2:
I thought it would save some CPU cycles and therefore some water if we reused the ci-dev run already in place, but happy to add a new yaml if you think is cleaner. :)

Same applies to cat vs artifact. I personally find the artifact useful, but I'm fine with dumping the logs if you prefer.

eddelbuettel:
I am open to either approach.  What I want to avoid, mostly, is for this to inflict :x: on routine PRs and updates. If we can avoid that, good. If we cannot, I would think an opt-in is cleaner. Let's see how it goes.

Enchufa2:
Ok, let's go for the cleaner approach then. :) Give me a sec.

Enchufa2:
All good then with this approach? Should I merge?

eddelbuettel:
Yes I think this is a good base layer.  I would have called it, say `werror.yaml` not `warning.yaml` but that can be changed subsequently too.

So next we start PRs adding additional triggering flags and tried to wind down the number of warnings / errors we get that way to zero?

Enchufa2:
> Yes I think this is a good base layer. I would have called it, say `werror.yaml`not`warning.yaml` but that can be changed subsequently too.

Said and done.

> So next we start PRs adding additional triggering flags and tried to wind down the number of warnings / errors we get that way to zero?

Yes, that's the idea.

PR #1502, opened 2026-09-03
Avoid -Wcast-function-type warnings

### LINKED ISSUE #1500: Consider -Wcast-function-type warnings
As reported in #1485, `-Wextra` reports this kind of warning at the interface with R. We need to evaluate whether this is something we can fix maybe with an explicit cast, or we should just whitelist it with `-Wno-cast-function-type` once `-Wextra` is added.

Enchufa2:
Closes #1500. Addressing this one first because this flag is added by `-Wextra` and it gets in the way of other warnings.

As the [gcc manual suggests](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wcast-function-type), and [others apply](https://github.com/postgres/postgres/blob/534db08f972852b77ea5e96e77d5fe45f3d8df93/src/include/c.h#L602-L607), I've defined a new `typedef void (*RCPP_FUNC)(void)` that is used as a generic intermediate cast. This affects not only our function casts, but also to the generated interfaces, so client packages will see this warning go away too only when they regenerate their interface with an updated version of Rcpp.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/werror.yml
+++ b/.github/workflows/werror.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  RCPP_CXXFLAGS: "-Werror"
+  RCPP_CXXFLAGS: "-Werror -Wcast-function-type"
 
 jobs:
   ci:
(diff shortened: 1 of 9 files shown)


eddelbuettel (approved):
Very nice!

PR #1503, opened 2026-09-03
Avoid -Wdeprecated-copy warnings

### LINKED ISSUE #1498: Consider -Wdeprecated-copy-with-user-provided-copy warnings
As reported in #1485, `-Wextra` reports this kind of warning. As discuss there, rather than adding an explicit default copy constructor, it would probably better to just remove the custom assignment operator. If I'm correct, if we remove it, the compiler generates a default one that delegates to the base class, which already handles the assignment nicely (i.e. shares the underlying SEXP, which has the tzone attribute already).

Enchufa2:
Closes #1498. As discussed there, this was triggered by the _[rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming))_ in the `Date(time)Vector` classes. This removes the redundant copy constructor, delegating in the parent class (note that the `tzone` assignment was a no-op, because the value is already there).

Like the previous PR, this should be harmless. But anyway I checked `anytime`, which uses these classes, and all is good.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/werror.yml
+++ b/.github/workflows/werror.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  RCPP_CXXFLAGS: "-Werror -Wcast-function-type"
+  RCPP_CXXFLAGS: "-Werror -Wcast-function-type -Wdeprecated-copy"
 
 jobs:
   ci:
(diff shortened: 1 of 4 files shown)


eddelbuettel:
RQuantlLib uses them too, and I just checked again the branch, also fine.

eddelbuettel (approved):
Nice and clean!  

These date(time) vector are still a mess, it all goes back to work I did when know way less about all this ... so it's still a bit of a scar.

PR #1505, opened 2026-09-03
Avoid -Wimplicit-fallthrough warnings

### LINKED ISSUE #1504: Consider -Wimplicit-fallthrough warnings
Another step towards `-Wextra`.

Enchufa2:
Closes #1504. We had some fallthrough comments in some places but not in others. This completes all cases.

Question: the `RCPP_LOOP_UNROLL_PTR` macro that I patch here too didn't raise any warning because it's not used at all. It's dead. I didn't find any references of client packages using it either. Should we just remove it?

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/werror.yml
+++ b/.github/workflows/werror.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   _R_CHECK_FORCE_SUGGESTS_: "false"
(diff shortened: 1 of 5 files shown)


eddelbuettel:
"Likely" but before, or "when", narrowing the API surface we probably want to run rev.deps.

eddelbuettel (approved):
:+1:

Issue #1488, opened 2026-07-29
Consider -Wconversion warnings


eddelbuettel:
Enabling `-Wconversion` (particularly with `clang++` as on the M1mac machine at CRAN, see [its README](https://www.stats.ox.ac.uk/pub/bdr/M1mac/00README.txt)) exposes a number of warnings.  The Oxford machine uses `clang++-21`, on Ubuntu 26.04 we also have `clang++-22` which Vienna uses (without the same options). Per [the README](https://www.stats.ox.ac.uk/pub/bdr/M1mac/00README.txt) this uses 

```
CFLAGS="-falign-functions=8 -g -O2 -Wall -pedantic -Wconversion -Wno-sign-conversion -Wstrict-prototypes"
C17FLAGS="-falign-functions=8 -g -O2 -Wall -pedantic -Wconversion -Wno-sign-conversion -Wno-strict-prototypes"
# ...
CXXFLAGS="-g -O2 -Wall -pedantic -Wconversion -Wno-sign-conversion"
```

Note that this turns 'sign-conversion' warnings off.  Those account for a ton.  I found first set of changes we could make by _not doing this_ i.e. by running with only `-Wconversion`.  That may be too radical, but it could still hide some overflows.

Also, our own compilation is 'clean' under `-Wconversion -Wno-sign-conversion` however other package may see

```
In file included from /Users/ripley/R/Library/Rcpp/include/Rcpp/sugar/functions/functions.h:63:
/Users/ripley/R/Library/Rcpp/include/Rcpp/sugar/functions/mean.h:38:14: warning: implicit conversion from 'R_xlen_t' (aka 'long') to 'long double' may lose precision [-Wimplicit-int-float-conversion]
   38 |         s /= n;
      |           ~~ ^
/Users/ripley/R/Library/Rcpp/include/Rcpp/sugar/functions/mean.h:44:20: warning: implicit conversion from 'R_xlen_t' (aka 'long') to 'long double' may lose precision [-Wimplicit-int-float-conversion]
   44 |             s += t/n;
      |                   ~^
```

Enchufa2:
This is with gcc 16.1.1 [install.log](https://github.com/user-attachments/files/30539536/install.log)

- 60 (unique) conversion warnings
- 108 (unique) sign conversion warnings

Enchufa2:
If we add the warnings from the tests and restrict the search to "may change" (value or the sign of the result), I get

- 2 unique conversion warnings
[...]

PR #1506, opened 2026-09-03
Avoid -Wunused-parameter warnings

### LINKED ISSUE #1497: Aiming at a more strict set of compilation flags

### LINKED ISSUE #1499: Consider -Wunused-parameter warnings
As reported in #1485, `-Wextra` reports this kind of warning. The proposal is to comment out unused parameters in function definitions.

Enchufa2:
Closes #1499, and closes #1497 by setting just `-Wall -Wextra` instead of all individual warnings we've been fixing.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/werror.yml
+++ b/.github/workflows/werror.yml
@@ -13,7 +13,7 @@ env:
 jobs:
   ci:
     runs-on: ubuntu-latest
-    container: rocker/drd
+    container: eddelbuettel/drd:26.04
 
     steps:
       - name: Checkout
--- a/.github/workflows/werror.yml
+++ b/.github/workflows/werror.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  RCPP_CXXFLAGS: "-Werror -Wcast-function-type -Wdeprecated-copy -Wimplicit-fallthrough"
+  RCPP_CXXFLAGS: "-Werror -Wall -Wextra"
 
 jobs:
   ci:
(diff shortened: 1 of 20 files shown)


kevinushey (approved):
LGTM!

.github/workflows/werror.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 env:
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  RCPP_CXXFLAGS: "-Werror -Wcast-function-type -Wdeprecated-copy -Wimplicit-fallthrough"
+  RCPP_CXXFLAGS: "-Werror -Wall -Wextra"
```
eddelbuettel, line 11: Woot!

eddelbuettel (approved):
You are on a very impressive roll!

.github/workflows/werror.yml
Enchufa2: For reference: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-W

Enchufa2:
Thanks, all set here. We fixed #1485 and paved the way for #1488. But that's for another day. :)

PR #1508, opened 2026-09-04
Fix -Wconversion warnings

### LINKED ISSUE #1488: Consider -Wconversion warnings

Enchufa2:
Closes #1488. There are 3 categories of changes here (divided in blocks in the ChangeLog):

1. **Ambiguous overload resolution** in Rcpp types. This happened to work, but depends on the compiler to do the right thing. The main problem was the `GenericProxy` constructor in `macros/interface.h`, which is inherited by pretty much everything, but each type has its own constructor that actually checks the SEXP, etc. So by [marking the `GenericProxy` constructor as `explicit`](https://github.com/RcppCore/Rcpp/pull/1508/changes/e8f723ed6a5b3ed6b78daf55038f9b67a1276924), we tell the compiler that implicit conversions should go through the specific constructors. Same in `DataFrame` and `ListOf` with their own generic constructors. Actually the `ListOf` case [was also special](https://github.com/RcppCore/Rcpp/pull/1508/changes/05ebafaaf057b9871250e85cc0e98268f55a511b) in its own way, just like everything there. ;-)
2. **Homogenization of `R_xlen_t` vs `int`**. This is a bit of a mess. There were methods that weren't converted to `R_xlen_t` like similar others and now are, and others were the use of `R_xlen_t` was unjustified, unnecessary or plainly wrong. Hopefully this brings some more order.
3. A bunch of explicit casts (not related to the previous point) that don't have any impact apart from being explicit about what we are doing, which is nice.

This requires full revdeps.

#### Checklist

- [x] Code compiles correctly

### DIFF
--- a/.github/workflows/werror.yml
+++ b/.github/workflows/werror.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  RCPP_CXXFLAGS: "-Werror -Wall -Wextra"
+  RCPP_CXXFLAGS: "-Werror -Wall -Wextra -Wconversion"
 
 jobs:
   ci:
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,51 @@
 	* inst/include/Rcpp/vector/Vector.h: Idem
 	* inst/tinytest/cpp/InternalFunction.cpp: Idem
 
+	* .github/workflows/werror.yaml: Add -Wconversion
+
+	* inst/include/Rcpp/macros/interface.h: Mark GenericProxy constructor as
+	explicit to resolve ambiguities with validated SEXP constructors
+	* inst/include/Rcpp/vector/Vector.h: Idem
+	* inst/include/Rcpp/DataFrame.h: Idem with ad hoc constructor
+	* inst/include/Rcpp/vector/ListOf.h: In this case, the implicit conversion
+	of ListOf<T> e.g. from List is intentional, so the previously generic
+	constructor now gets restricted to cases that are convertible to SEXP
+
+	* inst/include/Rcpp/stats/dpq/dpq.h: Expand and homogenize R_xlen_t use
+	* inst/include/Rcpp/sugar/functions/head.h: Idem
+	* inst/include/Rcpp/vector/MatrixColumn.h: Idem
+	* inst/include/Rcpp/vector/MatrixRow.h: Idem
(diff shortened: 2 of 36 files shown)


eddelbuettel (approved):
Conditional thumbs up, another rev.dep run will tell us more.

PS That said we will wait til we're good with the reverse depends runs.

eddelbuettel (approved):
Thanks for this, and thanks especially for excellent follow-up commit.

This should now be good to be merged, I will run another full set but we do not have to wait on it.

